PR #18, opened 2019-03-14
Allow running "serveFunctions" without package.json

### LINKED ISSUE #15: Serve fails if there's no package.json
If there's no package.json in either the function folder or the root, we fail to determine
the module path and break with a stacktrace.

Benaiah:
**- Summary**

Uses as the function "moduleDir" the first .git directory found by ascending the folder structure if "findModuleDirectory" returns false. If that still fails, uses "process.cwd()". Uses a shared memoized implementation for ascending directory finders.

This PR completes the work necessary in this repo to close #15.

**- Test plan**

Manually tested.

Allow running "serveFunctions" without a package.json

### DIFF
--- a/src/finders.js
+++ b/src/finders.js
@@ -157,17 +157,27 @@ Please ensure "${moduleName}" is installed in the project.`);
   return Array.from(filePaths);
 }
 
-function findModuleDir(dir) {
-  let basedir = dir;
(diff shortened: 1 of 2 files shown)


sw-yx:
we no longer serve functions in this repo - i'm still not sure if we need the functionality in this PR (i wasnt involved in this) but suggest moving it to netlify-dev-plugin.

PR #20, opened 2019-03-22
[STILL DISCUSSING, DO NOT MERGE] add webpack to ZISI: run webpack.config.js if detected in a folder

sw-yx:
**- Summary**

add webpack to ZISI: run webpack.config.js if detected in a folder

**- Test plan**

(see slack)

add webpack to ZISI: run webpack.config.js if detected in a folder, meaning we can support typescript functions (and all other capabilities of webpack)

image


sw-yx:
demo video https://netlify.slack.com/archives/CGW2Y6XPH/p1553258921129800

bcomnes:
Cool idea, and I think the workflow is ace, we should support it as easily as this! 

My only worry is that this means we are absorbing user's build dependencies into the dependencies of our zipping tool.  This definitely feels like a crossing of boundaries here.  Let me explain.

I do think there is more we can do to improve the install/build step of functions, but I would rather have a narrow kneck API rather than "here are the specific build tools we support" and take on dependency management for users.  I would bet money it's going to be a nightmare to support and cause version conflicts down the road.

I think we could achieve the same UX though, with a simpler contract (and we are customizing build bot stages as we speak so we could add in these hooks easily):

- Top level: we install dependencies, and run our users build command.  Their build command should run any build steps on their functions (in addition to their site) which depend on hoisted/top-level dependencies with whichever dependencies they need to do it with defined in their top-level package.json.  We zip and ship with hoisted node_modules the build function.  

- Nested package.jsons 

We descend into the functions folder.  If we see a `functions/.package.json` we run install on it, which can include any `prepare` or package.json install hooks.  The functions near this package.json should be build within those hooks.  Build tools and steps they need for that process should be defined in that package.json.  We decent further one more level: `functions/foo/package.json`.  Any folders with a package.json we run an install on, and in the install hooks, the function should be built.  After this process its complete, we zip and ship the whole shebang. 

So what does this accomplish? 

- A) Less top level scripting for users.  We already support a top level installation and build hook, which can be used to build functions in place, and then shipped in place zipItAndShipIt.
- b) We adopt automated and additional package.json install steps:
  - One for `functions/package.json` if it exists
  - One for every `functions/foo/package.json` if it exists.
  - Each install step provides ample hooks for running a build step and populating the closest node_modules.  Webpacking and TSing would be defined here.
  - ZipItAndShipIt would continue to prune and zip as normal.

bcomnes:
Conversely, we could have a tool that recursively runs the rebuild step based on these rules as well, if we are looking for a dev cycle tool.

bcomnes:
Per discussion with @sw-yx: Im going to get ZISI into the buildbot without this, and then experiment with a version that includes this, and maybe some other arrangements time permitting.  

I do think there is a good argument to include this, or something like it but I have a lot of questions:

- perhaps this is too limited in the long run (just webpack transforms) and is it something we want to maintain for eternity?
- What if you want to use some other transform (babel-cli or ts directly)? 
- What if you want installs to run on each function (does that buy a solution that can also work with go)? 
- What if a user wants to transform with a different webpack version? 
- What if a user wants to transform in place using their own version of webpack a webpack config file?  What does an opt out look like then?
- What if it becomes hard to upgrade this version of webpack, and everyone who wants the new version is stuck with having to go back to the src/dist pattern etc. 
- What happens when webpack falls out of favor for a hot new bundler?  
- is tying our zipping algorithm to on type of transform the right place to do this? 
- Would a different layer serve this capability better? (e.g. auto-run netlify-lambda in place if the webpack configs are found)... then run ZISI.  (Separation of concerns)
- Or perhaps this is the best bang for our buck right now and we should do it until we have something better?  


Open questions for me.   Curious on others thoughts.

sw-yx:
bret and i had a good convo about this today, centered around weighing the cost of adding more responsibility (a build step) into ZISI, vs the benefit (users not needing to setup their build step). The angles we want to think about are:

- how easy is it to opt out (e.g. not including a `webpack.config.js` file)
- debuggability (if something goes wrong with the build inside ZISI, will users feel like its our fault/know how to debug, vs owning their own build step)
- maintenance (if a webpack 5 comes along, do we upgrade? break compat?)
- how it works inside buildbot (there is an existing flow already, how much do we disrupt it)
- amount of "magic" vs desired DX (how much are users having to adapt to netlify specific conventions, and how much they get vs how much they give up)

one separate/parallel angle to this is that if we do end up installing deps in functions folders as part of the deploy step, we can also squeeze in a build step there as part of npm postinstall. i like this idea, however i also think it lengthens the feedback loop and doesnt work as well for local dev. idk, we can keep iterating!

bcomnes:
@sw-yx I saw a mention of builders in dev.  Does that take the place of this? Having a separate subsystem that we can run in front, separate from this step makes sense to me.

sw-yx:
agreed, closing

PR #22, opened 2019-04-02
use assignLoudly to make clear defaultPort is not a mistake

sw-yx:
**- Summary**

using default port of 30001 makes it look like a string concat error. (users will see this: https://github.com/netlify/netlify-dev-plugin/issues/58) we use 34567 instead.

also when using default port, we should tell the user about it.

image


biilmann:
LGTM

sw-yx:
just realized my prettier settings are messing with the diffs :( we should have a .prettierrc in this project.

PR #23, opened 2019-04-03
add prettier

sw-yx:
**- Summary**

adding standard project level prettier config so as to have clean diffs

i went with what we have in netlify-dev-plugin but up to you what options you want: https://prettier.io/docs/en/options.html

```
{
  "semi": false,
  "singleQuote": true,
  "printWidth": 120
}
```

please feel free to edit as you see fit!

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,5 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "printWidth": 120
+}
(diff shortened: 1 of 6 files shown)


sw-yx:
applied blank prettier: https://github.com/netlify/netlify-react-ui/blob/master/.prettierrc.json same as neltify react UI as expressed by calavera https://netlify.slack.com/archives/CBC2U9MMG/p1554304727059400

bcomnes:
Rebased this.

PR #34, opened 2019-04-26
Install function deps before zip + ship

DavidWells:
Automatically install sub function folder deps


src/install-deps.js
```diff
@@ -0,0 +1,37 @@
...
+const cp = require("child_process");
+const parallelLimit = require("run-parallel-limit");
+const series = require("run-series");
+
+module.exports = function installFunctionDeps(src) {
+  const fns = functionGlobber(src)
+  return new Promise((resolve, reject) => {
+    if (fns.length === 0) return resolve()
+
+    const jobs = fns.map(fnPath => {
```
sw-yx: worth filtering something here in case some function folders do not have a package.json?

bcomnes: The globber should only select folders with package.json if I understand correctly:

https://github.com/netlify/zip-it-and-ship-it/pull/34/files#diff-f1580f02b18be88d05da772c352b9bbaR27

src/install-deps.js
```diff
@@ -0,0 +1,37 @@
...
+      return resolve(data)
+    })
+  })
+}
+
+function functionGlobber(baseDir) {
+  const globStr = path.join(baseDir, "*/package.json")
+  const functions = glob.sync(globStr)
+  return functions.map(fnFolder => {
+    return fnFolder.substring(0, fnFolder.indexOf("package.json"))
```
bcomnes: Is this redundant with the glob?

bin.js
```diff
@@ -50,9 +51,16 @@ if (argv.help || !sourceArg || !destArg) {
...
 const dest = path.resolve(process.cwd(), destArg)
 
-zipIt.zipFunctions(source, dest, { skipGo: !argv['zip-go'] })
-.then(console.log)
-.catch(err => {
+/* Recursively loop through functions + function folders and install deps */
+installFunctionDeps(source).then(() => {
+  /* Then zip em up */
+  zipIt.zipFunctions(source, dest, { skipGo: !argv['zip-go'] })
+  .then(console.log)
```
bcomnes: Duplicate promise chain path? Maybe just return the promise from the arrow function and handle the chain in a non-nested manner?

bcomnes: Unless there is a reason to do it this way and I'm not seeing it.

src/install-deps.js
```diff
@@ -0,0 +1,37 @@
+const path = require("path");
+const glob = require("glob");
+const cp = require("child_process");
+const parallelLimit = require("run-parallel-limit");
```
bcomnes: If you prefer to use a promise version of this too, thats fine with me.  Just throwing that out there.

src/install-deps.js
DavidWells: yeah it ignores folders without `package.json`

src/install-deps.js
DavidWells: IDK you wrote this 😁

DavidWells: It works. I know that much

bin.js
DavidWells: Flattened it!

src/install-deps.js
DavidWells: Not familiar with `p-` libs. If you want to update go for it.

I promisified the main API

src/install-deps.js
bcomnes: OH! I see, its just the dirname.  Yorp.

bcomnes:
To capture the issue here: running this in buildbot isn't currently working because its run outside of build context.

Here are possible solutions:

- Refactor the build environment to set up and maintain a build environment across the different build steps in https://github.com/netlify/buildbot/blob/master/bot/stage.go#L13-L24 . This is the best solution because it gives clear separate between steps and would help reduce complexity in https://github.com/netlify/build-image
- Move the function prep step into the build step inside of run_build_functions.  This would be fairly simple, but we would need to inject a tmp folder for ZISI to work in.
- Something else... Ideas?

While I strongly advocate for the first option, the second option is a decent compromise without adding additional complexity (more of a reshuffle of where it is now.)

sw-yx:
lets just npm all the way for now? stop the immediate pain

bcomnes:
It shouldn't be difficult to accommodate yarn if we move ZISI into the build environment directly. That is the change that is necessary.

sw-yx:
you do have the most context here. really up to you.

bcomnes:
@biilmann would love your input on this as well. 

- The alternative would be a bash version of the nested npm/yarn install logic inside of https://github.com/netlify/build-image
- Dependency caching is limited/complex and unimplemented with this path.

sw-yx:
sweet! shall we ship it?

bcomnes:
I want to get some more sign-off on it today, but yeah, I think its the best approach we've come up with so far.  Also, it's it a non-breaking change from the module API perspective. You can opt in to the install flow by swapping the method you consume.

sw-yx:
fixed a merge conflict

bcomnes:
WE should release this a major or minor: (0.4.0 or 1.0.0).  The bin shouldn't be consumed anywhere else as a library, so minor is safe, or maybe we just call this 1.0.

You can play with this right now on `xenial` by setting `ZISI_VERSION` to `0.4.0-2`.  

Screen Shot 2019-05-08 at 4 19 26 PM

src/install-deps.js
```diff
@@ -0,0 +1,87 @@
...
+    return fnFolder.substring(0, fnFolder.indexOf("package.json"))
+  })
+}
+
+function verifyInstaller(type) {
+  if (type !== 'npm' && type !== 'yarn') {
+    throw new Error('Must check npm or yarn')
+  }
+  return new Promise((resolve, reject) => {
+    cp.exec(`which ${type}`, (error, stdout, stderr) => {
```
bcomnes: Need to double check if this works on windows.

ehmicky: This should not work on Windows. There's [`node-which`](https://github.com/npm/node-which) but this is a bad hack that is not maintained.

ehmicky:
This PR is currently very outdated compared to the latest `master`.

Netlify Build [automatically installs dependencies](https://github.com/netlify/build/blob/master/packages/build/src/install/functions.js) in Netlify Functions before calling `zip-it-and-ship-it`. Our users will get this feature as they opt-it in the Netlify Build beta.

Therefore I think we probably should close this PR? 

@erquhart

erquhart:
Go for it 👍

PR #35, opened 2019-05-07
Don't fail when encoding is missing

bcomnes:
Node-fetch is a very popular module, yet it uses undeclared peer dependencies.  This breaks some of our assumptions about whether or not we can safely skip peer and optional dependencies automatically.  If a module does something weird, it breaks that assumption (e.g. node-fetch not declaring peer deps).

Since this has been the number one complaint since we launched bundled deps, wee can add a single exception for this.


sw-yx:
awesome stuff bret

Issue #38, opened 2019-08-20
Zipped subdirectories paths seem incorrect


peterp:
I'm transpiling a lambda function using babel which is built into `dist/functions/test.js`. 

<img width="277" alt="image" src="https://user-images.githubusercontent.com/44849/63329884-0ff56680-c333-11e9-8bac-7cce88de0f72.png">

`test.js` requires code from another directory: `var _auth = require("../lib/auth0");`

I'm [packaging this function with ZISI](https://github.com/hammerframework/billable/blob/9b561f715c7a757db4e834e2207db40392b5b744/api/scripts/package-functions.js#L1-L32), and the resulting zip file has this content:
<img width="279" alt="image" src="https://user-images.githubusercontent.com/44849/63329497-5bf3db80-c332-11e9-919b-006c58600249.png">

Based on the screenshot above the problem seems that the paths are incorrect when trying to [execute the function](https://5d5ba8e475cf20b485c487ec--sleepy-chandrasekhar-b9d767.netlify.com/.netlify/functions/test):

```json
{"errorMessage":"Cannot find module '../lib/auth0'","errorType":"Error","stackTrace":["Function.Module._load (module.js:474:25)","Module.require (module.js:596:17)","require (internal/module.js:11:18)","Object.<anonymous> (/var/task/test.js:8:13)","Module._compile (module.js:652:30)","Object.Module._extensions..js (module.js:663:10)","Module.load (module.js:565:32)","tryModuleLoad (module.js:505:12)","Function.Module._load (module.js:497:3)"]}
```

Am I thinking about ZISI incorrectly?

ehmicky:
Fixed by #48.

ehmicky:
This fix should also be available in our [new Build beta](https://www.netlify.com/build/plugins-beta/) if you want to try it out :)

danoc:
> This fix should also be available as part of our [new Build beta](https://www.netlify.com/build/plugins-beta/) if you want to try it out :)

Cool! How does this relate to Netlify Build Plugins though? Is that infrastructure using a pre-release version of `zip-it-and-ship-it`?

ehmicky:
Yes exactly, the new Netlify Build includes several features, including Netlify Plugins. It uses the pre-release version of `zip-it-and-ship-it`.

Issue #37, opened 2019-07-02
usage with yarn workspaces


tibotiber:
**- Do you want to request a *feature* or report a *bug*?**
bug / better docs

**- What is the current behavior?**
From what I understand, using zisi with [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) does not work due to the hoisting of modules causing a module to potentially be higher in the tree than the `package.json` it was declared in. The problem and potential solution are described [here](https://yarnpkg.com/blog/2018/02/15/nohoist/).

To be more precise, I have tried to add to the root `package.json`:
```
  "workspaces": {
    "packages": [
      "A",
      "B",
      "functions"
    ],
    "nohoist": [
      "functions/**"
    ]
  },
```
and to the `package.json` of my functions:
```
  "private": true,
  "workspaces": {
    "nohoist": [
      "**"
    ]
  },
```

I can confirm that no module in the functions directory is hoisted after that. However my deploy still fails with the following error:
```
12:34:14 AM: Function Dir: /opt/build/repo/functions/dist
12:34:14 AM: TempDir: /tmp/zisi-857404437
12:34:15 AM: Prepping functions with zip-it-and-ship-it 0.3.1
12:34:34 AM: [ { path: '/tmp/zisi-857404437/health.zip', runtime: 'js' },
12:34:34 AM:   { path: '/tmp/zisi-857404437/process3dModel.zip',
12:34:34 AM:     runtime: 'js' } ]
12:34:34 AM: Prepping functions complete
[...]
12:34:34 AM: Build script success
12:34:34 AM: Starting to deploy site from 'studio/build'
12:34:34 AM: Creating deploy tree 
12:34:34 AM: 0 new files to upload
12:34:34 AM: 2 new functions to upload
12:36:56 AM: Failed to upload file: &{Name:process3dModel Sum:15bec3b5d5780a2927fdb3d879651765dd97f9adcaab59915caf184e81eadc27 Runtime:js Size:<nil> Path: Buffer:0xc000934000}
12:37:04 AM: Failed to upload file: &{Name:health Sum:c3f2203387ca2e24ec3d20d8c27359ee1b8e49e2787bba658975267c978ae874 Runtime:js Size:<nil> Path: Buffer:0xc0001f6330}
12:37:04 AM: failed during stage 'deploying site': Failed to execute deploy: Upload cancelled: health
12:37:04 AM: Failing build: Failed to deploy site
[...]

PR #39, opened 2019-09-11
include sibling and parent paths in function zips

DavidWells:
Fix for https://github.com/netlify/zip-it-and-ship-it/issues/38 (and maybe #37).

Need to verify against Go function zips before taking this live.

Repro repo https://github.com/netlify-labs/repro-zisi-file-resolution

### DIFF
--- a/src/zip.js
+++ b/src/zip.js
@@ -5,6 +5,7 @@ const archiver = require("archiver");
 const elfTools = require("elf-tools");
 const { getDependencies, findModuleDir, findHandler } = require("./finders");
 const pAll = require("p-all");
+const getRelativePath = require('relative');
 
 class Zip {
   constructor(path) {
@@ -14,7 +15,11 @@ class Zip {
   }
 
   addLocalFile(path, data) {
-    this.archive.file(path, data);
+    if (Buffer.isBuffer(path)) {
+      this.archive.append(path, data);
+    } else {
+      this.archive.file(path, data);
+    }
   }
 
   finalize() {
@@ -84,9 +89,11 @@ function isGoExe(file) {
 }
 
 function zipEntryPath(file, basedir, moduledir) {
+  const parentDir = path.dirname(moduledir)
   return file
     .replace(basedir, "")
     .replace(moduledir, "")
+    .replace(parentDir, "")
     .replace(/^\//, "");
 }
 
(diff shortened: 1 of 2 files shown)


DavidWells:
To verify:

Clone https://github.com/netlify-labs/repro-zisi-file-resolution

`npm install`

then 

`npm run localbuild`

This should throw with `Error: Cannot find module '../../double-parent' from '/Users/123/Netlify/scratch/zisi-parent-dir/functions'`

In a parent dir from repo make `double-parent.js` or remove import https://github.com/netlify-labs/repro-zisi-file-resolution/blob/master/functions/two.js#L2

DavidWells:
Todo:

- [ ] Clean up code
- [ ] Verify go functions
- [ ] Look into binaries being excluded (separate issue)

src/zip.js
```diff
@@ -110,19 +117,180 @@ function zipJs(functionPath, zipPath) {
...
+          const modulePath = getOriginalPath(requirePath, file)
+          const fileInFolder = isPathInside(modulePath, originalDirectory)
+          // const isParentPath = requirePath.match(parentPathRegex)
+
+          /* If file in function folder or is a node module */
+          if (fileInFolder || inNodeModules(modulePath)) {
+            processedPaths.push(requirePath)
+          } else {
+          /* If require paths are in parent/sibling directories we will need to fix code paths */
+            console.log(`Including ${requirePath} from parent paths for ${file}`)
```
ehmicky: It seems like the rest of the code is using `debug()` statements instead?

src/zip.js
```diff
@@ -110,19 +117,180 @@ function zipJs(functionPath, zipPath) {
...
+        stats: stat
+      });
+    } else {
+      // Find and Fix relative paths parent paths
+      const rawContent = fs.readFileSync(file, 'utf-8')
+      const requireStatements = rawContent.match(requireRegex)
+
+      // If mapping found, use modifed entryPath as file name
+      if (mappings[file]) {
+        // console.log('Map match on', file)
```
ehmicky: Those comments can probably be removed. There are several others below as well.

src/zip.js
```diff
@@ -110,19 +117,180 @@ function zipJs(functionPath, zipPath) {
   }
   const moduledir = findModuleDir(basedir);
 
+  const requireRegex = /require\(['|"](.*)['|"]\)/gm
```
ehmicky: This probably should be `/require\(['|"][^'"]+['|"]\)/gm` otherwise this might fail e.g. if two `require` statements are on the same line, e.g. `require("a"); require("b");`

src/zip.js
```diff
@@ -110,19 +117,180 @@ function zipJs(functionPath, zipPath) {
...
+
+function removeQuotes(str) {
+  return str.replace(/['|"]/g, '')
+}
+
+function getOriginalPath(requirePath, file) {
+  const cleanp = removeQuotes(requirePath)
+  // Path ./whatever
+  if (cleanp.match(/^\./)) {
+    return require.resolve(path.join(path.dirname(file), cleanp))
```
ehmicky: Using the `resolve` node module (that we use in the `finders.js`) might be safer?

src/zip.js
```diff
@@ -110,19 +117,180 @@ function zipJs(functionPath, zipPath) {
...
+      zip.addLocalFile(file, {
+        name: entryPath,
+        mode: stat.mode,
+        date: new Date(0), // Ensure sha256 stability regardless of mtime
+        stats: stat
+      });
+    } else {
+      // Find and Fix relative paths parent paths
+      const rawContent = fs.readFileSync(file, 'utf-8')
+      const requireStatements = rawContent.match(requireRegex)
```
ehmicky: Would using `precinct` be better?

src/zip.js
```diff
@@ -110,19 +117,180 @@ function zipJs(functionPath, zipPath) {
...
+    // Remove file extension
+    .replace(/\.js$/, '')
+  return `${removeQuotes(entry)}.js`
+}
+
+function inNodeModules(filePath) {
+  return filePath.match(/\/node_modules\//)
+}
+
+function removeFileExtension(str) {
```
ehmicky: This could be `.json` too

ehmicky:
I might be missing some information on this, but it seems to me that the following points might create issues down the road?
  - modifying user code (`require()` statements)
  - several directories trees gets merged in the zipped file at different directory levels. This could lead to different files with the same filename to overwrite each other

Instead would there be a solution that keeps the initial directory structure? For example, if the zipped file was a mirror of the package root, but excluding the non-required files. 

This would mean the entry point would not top-level inside the zipped file: would this be an issue? If this was an issue, this could be solved by creating a dummy top-level `index.js` pointing to the real entry point. This seems to me this would be less intrusive that modifying `require()` statements. What do you think?

ehmicky:
I've added a PR at #49 that does just that. I tried it against the demo repository and it works :tada: 

Please let me know what you think!

ehmicky:
This PR has been moved to #49 which itself has been merged to #48.

PR #45, opened 2019-09-13
Improve packaging

ehmicky:
**- Summary**

This improves the `pkg` packaging:
  - add packages for Windows
  - add packages for Node 8 and 12, not only Node 10
  - make it work on Windows. For example the `cut` and `tar` global dependencies might not available on Windows.
  - add unit tests (100% test coverage).  
  - improve speed by running each target in parallel and using streaming
  - does not rely on `git` global dependency
  
Some other changes:
  - replace `rimraf` dependency by [`del`](https://github.com/sindresorhus/del). It's built on top of `rimraf` but adds a promise interface.
  - remove `mkdirp` dependency

Note that the names of the GitHub assets have been changed. This is because they previously did not include the Node.js version, but we now published assets for different Node.js versions, which needs to be shown in the GitHub asset name. Please let me know if this might be a problem.

Also the logic can be abstracted to its own repository if we have other repositories that need to do similar packaging. Please let me know if we should do that.
 
**- Test plan**

Unit tests are included. They try to package files, then unpackage them and execute them, for each supported target.

Improve packaging.


sw-yx:
monster PR!

ehmicky:
Closed by #94

PR #48, opened 2019-09-17
Improve core logic

### LINKED ISSUE #15: Serve fails if there's no package.json

### LINKED ISSUE #24: [easy] fail more gracefully when no functions folder
**- Do you want to request a *feature* or report a *bug*?**

feature

when a `functions` field is specified but the folder doesnt exist, we should create it, or warn and exit. instead, now we show this error:

```
Error: ENOENT: no such file or directory, scandir 'lambda'
    at createHandler (~/Netlify/netlify-dev-plugin/node_modules/@netlify/zip-it-and-ship-it/src/serve.js:61:6)
    at serveFunctions (~/Netlify/netlify-dev-plugin/node_modules/@netlify/zip-it-and-ship-it/src/serve.js:159:16)
```

### LINKED ISSUE #65: HTML files are removed from node_modules dependencies
Hi!

I am not exactly sure this is the place to report this.

I started using Netlify Functions where I have a function that runs Google Lighthouse.
I started getting this error: 
`Error: ENOENT: no such file or directory, open '/var/task/node_modules/lighthouse/lighthouse-core/report/html/report-template.html'`

Using Netlify CLI I ran `netlify functions:build` and inspected the zip file it returned, to see that all the html files from node_modules/lighthouse folder has been removed by the build.

I looked at the code for the `functions:build` command and it led me to here.

I am not getting this error when running `netlify dev` so I assume that it doesn't zip the functions.

I might be doing something wrong here and would love an answer.

Thanks a lot!

### LINKED ISSUE #67: Some dependencies are not zipped in the bundle if they are declared via a local path
**- Do you want to request a *feature* or report a *bug*?**
Bug

The function is zipped and deployed correctly. However the function does not work properly because of a missing dependency.

**- If the current behavior is a bug, please provide the steps to reproduce.**
Go to https://nx-crusher.netlify.com/.netlify/functions/main/graphql

You will see the following stack trace : 
```
{
"errorMessage": "Cannot find module 'apollo-server-core'",
"errorType": "Error",
"stackTrace": [
"Function.Module._load (module.js:474:25)",
"Module.require (module.js:596:17)",
"require (internal/module.js:11:18)",
"Object.<anonymous> (/var/task/src/node_modules/@nestjs/graphql/dist/graphql-definitions.factory.js:13:30)",
"Module._compile (module.js:652:30)",
"Object.Module._extensions..js (module.js:663:10)",
"Module.load (module.js:565:32)",
"tryModuleLoad (module.js:505:12)",
"Function.Module._load (module.js:497:3)"
]
}
```

I tried to troubleshoot the issue:

It looks like my function depends on `@nestjs/graphql` package, which uses `apollo-server-express` : https://github.com/nestjs/graphql/blob/master/package.json#L27

`apollo-server-express` depends on `apollo-server-core` : https://github.com/apollographql/apollo-server/blob/master/packages/apollo-server-express/package.json#L35

But as you can see, `apollo-server-core` dependency is declared via a [local path](https://docs.npmjs.com/files/package.json#local-paths).

Not sure if it's related to the current bug, but my intuition tells me it is.

Dependencies declared with local path should be bundled in the zip archive.

ehmicky:
**- Summary**

This PR improves the core logic:
  - use async I/O everywhere instead of sync I/O
  - when the destination folder does not exist, it is created instead of throwing an error
  - allow `functionFolder/index.js`
  - allow requiring files outside of the function directory (fixes #49)
  - allow Node.js `>= 8.3.0` (instead of `>= 8.5.0`)
  
In addition the following bugs are fixed:
  - missing optional dependencies should not throw. However, at the moment, this only works for optional dependencies of dependencies, not top-level optional dependencies.
  - when the function folder has a `node_modules` subfolder, it is included in the zipped file even if it's not required.
  - it does not work sometimes when there are no `package.json` (fixes #15)
  - better error messages when the functions folder does not exist (fixes #24)
  - some files are excluded from node modules where they should not (e.g. Google `lighthouse` did not work) (fixes #65).
  - `@scoped` packages dependencies are sometimes missing (fixes #67) 

Also this reduces the number of exported functions. It now only exports `zipFunctions()` and `zipFunction()`. I checked on all the libraries using `zip-it-and-ship-it` and all are either using the binary or `zipFunctions()`. Removing those extra exports will make it easier to change the logic without introducing breaking changes.

Finally the following refactoring has been done to improve code quality:
  - remove `npm-packlist`. It is replaced by performing some file globbing, which does exactly the same, but does not tie us to the inner workings of `npm-packlist`.
  - remove some dead code
  - simplify the code
  - split big functions into smaller functions
  - split big files into smaller files

**- Test plan**

100% test coverage!

Since this is a major rewrite, I would really like this to be fully tested against live data. For example (if that's possible) deploying every current Functions in a stage environment (but with the production data) with the new logic and make sure everything looks fine. Deploying gradually to few users at a time would be great too (if that's possible).

Improve the core logic.


DavidWells:
Need the exports back for build plugins to use getDependencies, filesForFunctionZip, zipFunction, zipFunctions

Also for ‘functionFolder/index.js’ we need to get API to support or automatically rename index.js to the parent folder name

DavidWells:
We can probably clean up filesforfunctionzip

ehmicky:
For `index.js`, this should work without API support providing #49 is merged. This would result in the following archive:

```
name.js (points to src/index.js)
src/
  index.js
```

#49 decouples the entry point (`name.js`, used by the API) from the main file (`name.js`, `name/name.js` or `name/index.js`, defined by the user), which allows us to change the main file convention (e.g. supporting `index.js`) without breaking changes in the future.

For the exports, I've added back `zipFunction()`, so now `zipFunctions()` and `zipFunction()` are exported. For `filesForFunctionsZip()` and `getDependencies()`, I would like to confirm: are we currently using those? I could not find any libraries using those exports. The reason I am asking is:
  - their signature has changed a little, so I want to make sure they are not used
  - if they were exported, some logic would need to be repeated for each function file (such as retrieving the `package.json`) instead of once for all functions
  - they are Node.js-specific and tied to the internal logic so I was wondering if we should hide them from the exported interface

src/dependencies.js
```diff
@@ -0,0 +1,180 @@
...
+// `require(path)`
+const getImportDependencies = function(dependency, basedir, packageJson, state) {
+  if (LOCAL_IMPORT_REGEXP.test(dependency)) {
+    return getLocalImportDependencies(dependency, basedir, packageJson, state)
+  }
+
+  return getModuleDependencies(dependency, basedir, state, packageJson)
+}
+
+const LOCAL_IMPORT_REGEXP = /^\.|\//
```
peterp: I think this might be incorrect, I'm guessing we're trying to match `/path/to/file` or `./path/to/file` for local files? This could also the slash in `@apollo/lambda-server`.

```
// startswith '.' or '/'
const LOCAL_IMPORT_REGEXP = /^(\.|\/)/
```

peterp:
I'm not 100% sure if this is including all the `files: []` from a `package.json`, but I think it should based on [this comment](https://github.com/netlify/zip-it-and-ship-it/issues/65#issuecomment-546266683):

> When it comes to node modules (such as lighthouse), it includes all the files if the package.json files property is defined. Otherwise we ignore any file whose extension is log, html, md, map, ts, png, jpeg, jpg, gif, css or patch.

This is what's reported by `--dry-run`:
```
$ npm publish --dry-run
npm notice 📦  @generated/photon@0.0.0-generated
npm notice === Tarball Contents ===
npm notice 21.3MB  runtime/query-engine-linux-glibc-libssl1.0.2
npm notice 24.9kB  index.js
npm notice 405.6kB runtime/index.js
npm notice 225B    package.json
npm notice 205B    README.md
npm notice === Tarball Details ===
```

but `runtime/query-engine-linux-glibc-libssl1.0.2` is not included, only when it's explicitly required by this package.

I did some digging, and I think the way that the local files and modules are detected may be incorrect, I've left a comment on the line that I think may be causing the problem.

ehmicky:
@peterp thanks a lot for reviewing this! :tada: 

I've fixed the issue above, could you check whether it works now?

src/dependencies.js
ehmicky: Great catch! Fixed in 858a0ae

peterp:
Works 100%, thank you!

PR #49, opened 2019-09-19
Allow requiring files outside of the function directory

### LINKED ISSUE #38: Zipped subdirectories paths seem incorrect

ehmicky:
**- Summary**

Fixes #38 and (maybe) #37.
There is an alternative solution at #39.

At the moment, `zip-it-and-ship-it` fails when requiring any files outside the function directory. 

This solves it by keeping the same directory structure inside the zip archive. The required files and node modules will be present in the archive, but not the non-required files and node modules. The lowest common ancestor directory is used, which makes it work whether there's a `package.json` or not.

Since the BuildBot relies on the entry point being the same filename as the archive (can you please confirm this?), the archive includes an entry point that simply forwards to the function main file using `module.exports = require("...")`. The archive is namespaced under a `src/` folder to ensure the entry point does not overwrite any files.

For example, when using the following source files:

```
one/
  one.js
  two.js (required by one/one.js)
  three.js (not required by one/one.js)
four.js
test/
  one.js (required by one/one.js)
  four.js (required by four.js)
```  

Results in `one.zip` containing:

```
src/
  one/
    one.js
    two.js
  test/
    one.js
one.js (this requires "./src/one/one.js")
```

and `two.zip` containing:

```
src/
  four.js
  test/
    four.js
four.js (this requires "./src/four.js")
```

**- Test plan**

There's some unit tests for it at #47.

I've also tried it against the [demo repository](https://github.com/netlify-labs/repro-zisi-file-resolution) and it fixes the issue.
 
Allow requiring files outside of the function directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "archiver": "^3.0.0",
+    "common-path-prefix": "^2.0.0",
     "debug": "^4.1.1",
     "elf-tools": "^1.1.1",
     "glob": "^7.1.3",
--- a/src/zip.js
+++ b/src/zip.js
@@ -5,6 +5,7 @@ const glob = require('glob')
 const archiver = require('archiver')
 const elfTools = require('elf-tools')
 const pAll = require('p-all')
+const commonPathPrefix = require('common-path-prefix')
 
 const { getDependencies, findModuleDir, findHandler } = require('./finders')
 
@@ -76,13 +77,6 @@ function isGoExe(file) {
   }
 }
 
-function zipEntryPath(file, basedir, moduledir) {
-  return file
-    .replace(basedir, '')
-    .replace(moduledir, '')
-    .replace(/^\//, '')
-}
-
 function zipGoExe(file, zipPath) {
   const zip = new Zip(zipPath)
   const stat = fs.lstatSync(file)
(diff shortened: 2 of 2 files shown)


DavidWells:
If this solves https://github.com/netlify/zip-it-and-ship-it/pull/39 looks good to me.

We need to verify this works as expected in prod and functions fire correctly (e.g no missing modules/file path errors returned from lambda functions)

Can’t tell from the paths used here if it will work.

Functions always have to get “flattened” to:

```
name.js
Or
name/name.js
name/name-sib-file.js
name/node_modules
```

peterp:
Thanks for this 🙇 , I'll test it out on my current repo

ehmicky:
Thanks @peterp, please let us know if this works!

peterp:
It seems to be working correctly!

I've got some personal issues to resolve in my project, but it looks good.

ehmicky:
@peterp, great! This PR might need a little time to get merged because this impacts every user of Netlify Functions.

@DavidWells do you have an idea on how to test this against every Netlify Function in production? If we don't have a process for this, I think it would be a good idea to set one up.

peterp:
@ehmicky No problem, I'll deploy locally in the meantime with `netlify deploy`. Thanks for this. Please let me know if you need anything else.

peterp:
@ehmicky Unfortunately I just realised that this does not include dynamically imported files. We use `requireDir` to import a whole slew of things, and I guess that there's no way that it could include those since that's only determined during runtime.

It would be cool if there was a way to pass paths that could be explicitly bundled in addition to those that are "determined" to be bundled.

ehmicky:
At the moment we don't support `require(variableName)`. Note that this is not specific to this PR (we have never supported it).

That's because we rely on static analysis (we use the [`precinct`](https://github.com/dependents/node-precinct) library) to find all `require()` statements in order to only zip the files that are used by your Functions. This ensures the zip archive is as small as possible, which gives you a faster cold start on AWS Lambda.

To be able to support `require(variableName)` we would need to do dynamic analysis instead (i.e. run your code) which is quite tricky. If you have any ideas on how to approach this problem, please let us know! However please do so in a separate issue/PR as this is unrelated to this PR I believe.

peterp:
@ehmicky Thanks, I totally understand, and sorry for branching off topic.

I think there's a way that we can deal with this during our build time by converting the dynamic imports into static ones. 

I'll give it a go and write something down. It might be useful for future serverless people.

ehmicky:
This PR has been merged to #48.

PR #64, opened 2019-10-21
Improve CI OS coverage

ehmicky:
This adds Ubuntu to the list of OS tested by our GitHub actions.


ehmicky:
There seems to be some issues with GitHub actions right now. Holding off on this.

PR #75, opened 2019-11-27
 Add support for optional peer dependencies

### LINKED ISSUE #74: Add support for optional peer dependencies
While [trying to deploy](https://community.netlify.com/t/cannot-upload-build-function-to-aws-lambda/5261) a GraphQL server utilizing Prisma Framework, I observed that the generated bundle is overly large (~70 MB), due to the fact that `prisma2` gets bundled as being the peer dependency of `@prisma/photon`. In order to avoid this behavior, I'm [making `prisma2` an optional peer dependency](https://github.com/prisma/photonjs/pull/313) of `@prisma/photon` by using the `peerDependenciesMeta` property in `package.json` [introduced by Yarn](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md).

This project should avoid bundling optional dependencies by default, with exclusions specifiable through additional configuration.

kripod:
**Summary**

Fixes #74.

Optional peer dependencies may not be necessary in the final bundle, so they can safely be ignored.

**Description for the changelog**

 Add support for optional peer dependencies

### DIFF
--- a/src/finders.js
+++ b/src/finders.js
@@ -127,7 +127,17 @@ function getDependencies(filename, basedir) {
 
       if (dependencies) {
         Object.keys(dependencies).forEach(dependency => {
-          handle(dependency, currentModulePath, new Set(Object.keys(packageJson.optionalDependencies)))
+          handle(
+            dependency,
+            currentModulePath,
+            new Set([
+              ...Object.keys(packageJson.optionalDependencies),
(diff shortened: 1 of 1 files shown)


ehmicky:
See #74 for the ongoing discussion about this.

PR #83, opened 2020-01-27
Check peerDependenciesMeta for Optional Dependencies

devotox:
- Summary

This updates the optional dependencies check to also check peer dependencies meta

The Peer Dependencies Meta field was added to Yarn here:
[Yarn#6671](https://github.com/yarnpkg/yarn/pull/6671)

This is used by a number of packages to allow them to be `yarn -pnp` compatible, most notably [Knex](https://github.com/knex/knex). 
This feature was added in version [0.20.1](https://github.com/knex/knex/blob/master/CHANGELOG.md#0201---29-october-2019)

- Test plan

Unit tests are included but cannot be run as it cannot get to that branch of the code in a similar way as the `Ignore missing optional dependencies` test.

- Description for the changelog

Check Peer Dependencies Meta for Optional Dependencies

### DIFF
--- a/src/finders.js
+++ b/src/finders.js
@@ -117,6 +117,16 @@ function getDependencies(filename, basedir) {
     const currentModulePath = path.join(currentModule.path, '..')
     const packageJson = currentModule.packageJson
 
+    if (packageJson.peerDependenciesMeta) {
+      packageJson.optionalDependencies = packageJson.optionalDependencies || {}
+
+      for (const depKey in packageJson.peerDependenciesMeta) {
+        if (packageJson.peerDependenciesMeta[depKey].optional) {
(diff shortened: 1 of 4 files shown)


ehmicky:
Thanks a lot @devotox for this PR!

devotox:
@ehmicky when do you think this will be released and be used with Netlify CLI

ehmicky:
@devotox This should be live if you use the [Netlify Build beta](https://www.netlify.com/build/plugins-beta/).

devotox:
@ehmicky I have not been given access to Netlify Build Beta and was wondering how I could use this with [Netlify Actions](https://github.com/netlify/actions/tree/master/cli).

Would I need to fork it all?

ehmicky:
@DavidWells Would it be possible to give @devotox access to the Netlify Build beta?

PR #89, opened 2020-04-01
Fix Go Functions

ehmicky:
Go Functions can be bundled in two different modes:
  1. Copied as is
  2. Bundled in a `.zip` file

The `skipGo` option (called `zip-go` in the CLI with the inverse value) controls this.

We accidentally changed the default value of `skipGo` with the latest beta version of `zip-it-and-ship-it`. Go Functions are now bundled as `.zip` file (`2.`) instead of being copied as is (`1.`). Due to this, Go Functions are currently failing in production when Netlify Build beta is enabled (https://github.com/netlify/build/issues/1089). 

Part of the confusion was due to the fact that we are using:
  - the `skipGo` boolean option programmatically
  - the `--zip-go` boolean CLI flag, with an inverse value

This PR also introduces a `--skip-go` boolean CLI flag with the same value as the `skipGo` programmatic option. It deprecates the `--zip-go` CLI flag which should be removed in the next major release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ $ zip-it-and-ship-it --help
 @netlify/zip-it-and-ship-it: Zip lambda functions and their dependencies for deployment
 
 Usage: zip-it-and-ship-it [source] [destination] {options}
-    --zip-go, -g          zip go binaries (default: false)
+    --skip-go             do not zip go binaries (default: true)
     --help, -h            show help
     --version, -v         print the version of the program
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -101,7 +101,7 @@ $ zip-it-and-ship-it --help
 @netlify/zip-it-and-ship-it: Zip lambda functions and their dependencies for deployment
 
 Usage: zip-it-and-ship-it [source] [destination] {options}
-    --zip-go, -g          zip go binaries (default: false)
+    --skip-go             do not zip go binaries (default: true)
```
mraerino, line 104: a boolean flag with a default of `true` does not make sense to me

ehmicky: You're right!
I just fixed it in #91, could you check it out? Thanks!

PR #100, opened 2020-04-27
fix(dependencies): ignore `@prisma/cli`

timsuchanek:
We renamed `prisma2` to `@prisma/cli`.

### DIFF
--- a/src/dependencies.js
+++ b/src/dependencies.js
@@ -169,7 +169,7 @@ const shouldIncludePeerDependency = function(name) {
   return !EXCLUDED_PEER_DEPENDENCIES.includes(name)
 }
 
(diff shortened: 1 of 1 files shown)


ehmicky:
Thanks for the update!
Maybe the old name should be kept in that array for users of the older `prisma2`?

timsuchanek:
Yes, that's true! Just added it.

ehmicky:
This is released in `0.4.0-15`.

Pending PR to include it in the Netlify CLI: https://github.com/netlify/cli/pull/867

I am also currently working on including it our production builds for Netlify Functions. This would be live only for users with the Build plugins beta enabled because the non-beta users are currently using an older version of `zip-it-and-ship-it`.

PR #138, opened 2020-07-15
Upgrade several devDependencies

ehmicky:
This upgrades several `devDependencies`.


ehmicky:
I am not sure which dependency is making the tests fail (although it seems to be `adm-zip`). I am going to break this down into several PRs.

PR #226, opened 2020-10-21
feat(binary): improve binary file detection

calavera:
Use a more fully featured library to detect binary files that Netlify Functions support.

**- Test plan**

I've added new tests to verify that this continues working as expected.


src/binary.js
```diff
@@ -0,0 +1,35 @@
+const { readFile } = require('fs')
+
+const { detect, Runtime } = require('elf-cam')
```
ehmicky, line 3: Is there an option to target Node `8.3.0` with `wasm-pack`?

`elf-cam` transpiled code uses `TextDecoder`. This was added in `8.3.0`, but as an export of `require('util')`. In Node `11.0.0`, it was added as a global variable. Part of the transpiled code uses it as a global variable, which throws a deprecation warning on Node `<11.0.0`:

```
ExperimentalWarning: The WHATWG Encoding Standard implementation is an experimental API. It should not yet be used in production applications.
```

In production, `zip-it-and-ship-it` is run with Node `12`. However, in the CLI (`netlify build` command), we are using the user's Node.js version, which might be Node `<11`. We plan on deprecating Node `8`, but not Node `10`.

src/binary.js
```diff
@@ -0,0 +1,35 @@
...
+        return 'go'
+      case Runtime.Rust:
+        return 'rs'
+      default:
+        return undefined
+    }
+  } catch (error) {
+    return undefined
+  }
+}
```
ehmicky, line 26: Would it make sense to use a plain object instead?

```js
const binaryRuntime = async function(path) {
  try {
    const buffer = await pReadFile(path)
    return RUNTIMES[detect(buffer)]
  } catch (error) {}
}

const RUNTIMES = {
  [Runtime.Go]: 'go',
  [Runtime.Rust]: 'rs',
}
```

src/main.test.js
```diff
@@ -333,3 +333,25 @@ test('Can list all function files with listFunctionsFiles()', async t => {
...
+  // The library we use for unzipping does not keep executable permissions.
+  // https://github.com/cthackers/adm-zip/issues/86
+  // However `chmod()` is not cross-platform
+  if (platform === 'linux') {
+    await pChmod(unzippedFile, 0o755)
+
+    const { stdout } = await execa(unzippedFile)
+    t.is(stdout, 'Hello, world!')
+  }
+})
```
ehmicky, line 357: Should we move this test below the `Zips Go function files` test?

src/main.js
```diff
@@ -52,14 +52,25 @@ const zipFunction = async function(srcPath, destFolder, { skipGo = true, zipGo =
...
+      await zipBinary(srcPath, destPath, filename, stat)
       return { path: destPath, runtime }
     } else {
       const destPath = join(destFolder, filename)
       await cpFile(srcPath, destPath)
       return { path: destPath, runtime }
     }
   }
+
+  if (runtime === 'rs') {
```
ehmicky, line 64: Would it be possible to add `rs` to the `README.md` documentation of the `runtime` property?

ehmicky (changes requested):
This is awesome! :tada:

src/main.test.js
calavera: I don't see a clear reason why when both tests are independent of each other.

calavera:
@ehmicky I've updated this PR with your suggestions.

src/binary.js
calavera: I don't think that warning is related to global exports. The transpiled code always use the export from the util library. I put the code in a gist so you can see it:

https://gist.github.com/calavera/0c77feb68fd57ce6ebcd20c4b3592366

Node 8.3.0 always prints that warning using the util library:

image

ehmicky: Oh you're correct. :+1: 
This warning is not printed in Node `10.0.0`. Considering [we're now dropping support for Node 8](https://github.com/netlify/cli/issues/1568) in the CLI, we're probably good there.

ehmicky (approved):
Sweet! :tada:

PR #231, opened 2020-10-27
Return function names in `listFunctions()` and `listFunctionsFiles()`

ehmicky:
Required by https://github.com/netlify/cli/issues/1477

Function names are important because they are the ones used in the function URL. 
They are different from the function filename since they do not include the file extension.

This PR returns the function `name` in both `listFunctions()` and `listFunctionsFiles()`.

### DIFF
--- a/src/dependencies.js
+++ b/src/dependencies.js
@@ -14,7 +14,7 @@ const pGlob = promisify(glob)
 // Retrieve the paths to the Node.js files to zip.
 // We only include the files actually needed by the function because AWS Lambda
 // has a size limit for the zipped file. It also makes cold starts faster.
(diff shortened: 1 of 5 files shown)


src/dependencies.js
```diff
@@ -14,7 +14,7 @@ const pGlob = promisify(glob)
 // Retrieve the paths to the Node.js files to zip.
 // We only include the files actually needed by the function because AWS Lambda
 // has a size limit for the zipped file. It also makes cold starts faster.
-const listNodeFiles = async function(srcPath, filename, mainFile, srcDir, stat) {
```
ehmicky, line 17: This variable was not used.

erezrokah (approved):

PR #242, opened 2020-11-19
feat(js): add list of tree shakable external dependencies

erezrokah:
**- Summary**

Still in draft as it needs testing (automated and manual) and some code cleanup.
For some reason I'm getting 500 errors when trying to deploy functions locally so I'm having a hard time trying to test this.

**- Test plan**

Going to add tests once I'm finished with verifying this locally.

Add a list of tree shakable external modules to help reduce bundle size.
I'm only considering adding an escape hatch for this via a env variable (or any other configuration) so this behavior can be disabled/changed.

🐱


src/dependencies.js
```diff
@@ -92,15 +92,22 @@ const getFileDependencies = async function(path, packageJson, state) {
 // `require()` statements can be either `require('moduleName')` or
 // `require(path)`
 const getImportDependencies = function(dependency, basedir, packageJson, state) {
-  if (LOCAL_IMPORT_REGEXP.test(dependency)) {
+  const moduleName = getModuleName(dependency)
+  if (LOCAL_IMPORT_REGEXP.test(dependency) || isTreeShakable(moduleName)) {
     return getLocalImportDependencies(dependency, basedir, packageJson, state)
```
erezrokah, line 97: We should rename `getLocalImportDependencies` to `getTreeShakedDependencies` and `getModuleDependencies` to `getAllDependencies` (or something similar)

src/dependencies.js
```diff
@@ -92,15 +92,22 @@ const getFileDependencies = async function(path, packageJson, state) {
...
+  return getModuleDependencies(moduleName, basedir, state, packageJson)
 }
 
 const LOCAL_IMPORT_REGEXP = /^(\.|\/)/
 
+const isTreeShakable = function(moduleName) {
+  return TREE_SHAKABLE_MODULES.includes(moduleName)
+}
+
+const TREE_SHAKABLE_MODULES = ['next']
```
ehmicky, line 109: Did we double-check that no part of `next` source code is either doing a dynamic require, or reading a file published to npm?

erezrokah:
> For some reason I'm getting 500 errors when trying to deploy functions locally so I'm having a hard time trying to test this.

Looks like I'm hitting the issue described in [this thread](https://community.netlify.com/t/500-internal-server-error-on-cdn-diffing-files/20715/23)

src/dependencies.js
erezrokah: I want to verify this works for the `Next.js` plugin first, then we need to think about other use cases.
Using specific paths would keep this scoped to `next-on-netlify` - good point

erezrokah: See the list here:
https://github.com/netlify/zip-it-and-ship-it/pull/242/commits/940ca53cdb176c14b520abbcb66d175b7aa13a67#diff-d960073cd7b388518947c4e516460d086b7c49fdb95b3de6b16c3edfd002a8ccR108
based on https://github.com/erezrokah/erez-test-next-bundling (Next.js 10).

In `serverless` mode Next.js inlines all dependencies except `@ampproject/toolbox-optimizer`:
https://github.com/vercel/next.js/blob/4bca939ff72c83fbedae2cba91317e91b804c8d3/packages/next/build/webpack-config.ts#L723

In `experimental-serverless-trace` it uses the following logic:
https://github.com/vercel/next.js/blob/4bca939ff72c83fbedae2cba91317e91b804c8d3/packages/next/build/webpack-config.ts#L562 (this might imply we can do the same for the entire `next` package)

Without this fix functions sizes for https://github.com/erezrokah/erez-test-next-bundling are:
```
25M Dec  9 16:47 next_api_hello.zip
25M Dec  9 16:47 next_test.zip
```

With this fix functions sizes for https://github.com/erezrokah/erez-test-next-bundling are:
```
345K Dec  9 16:42 next_api_hello.zip
5.0M Dec  9 16:42 next_test.zip
```

src/dependencies.js
```diff
@@ -47,7 +47,7 @@ const getDependencies = async function(mainFile, srcDir) {
   const packageRoot = await pkgDir(srcDir)
   const packageJson = getPackageJson(packageRoot)
 
-  const state = { localFiles: [], modulePaths: [] }
+  const state = { localFiles: new Set(), modulePaths: new Set() }
```
erezrokah, line 50: Using `Set` instead of an array makes sense to me

src/dependencies.js
```diff
@@ -92,25 +92,64 @@ const getFileDependencies = async function(path, packageJson, state) {
...
 const LOCAL_IMPORT_REGEXP = /^(\.|\/)/
 
+const isTreeShakable = function(dependency) {
+  if (dependency.includes('next') && !TREE_SHAKABLE_DEPENDENCIES.has(dependency)) {
+    console.log(dependency)
+  }
+  return TREE_SHAKABLE_DEPENDENCIES.has(dependency)
+}
+
+const TREE_SHAKABLE_DEPENDENCIES = new Set([
```
erezrokah, line 111: Based on bundling `Next.js` v10 and v9:
https://github.com/erezrokah/erez-test-next-bundling
https://github.com/erezrokah/erez-test-next9-bundling

erezrokah: This seems very fragile as any new `next` require not on this list will result in bundling the entire framework.
However, for Next.js 9 this will work well since it is not updated frequently (I'm assuming only for security patches)

src/dependencies.js
```diff
@@ -106,7 +106,6 @@ const isTreeShakable = function(dependency) {
 }
 
 const TREE_SHAKABLE_DEPENDENCIES = new Set([
-  '@next/env',
```
erezrokah: Adding this to the list breaks the functions

src/fixtures/node-module-next/function.js
```diff
@@ -0,0 +1,7 @@
+try {
+  // eslint-disable-next-line node/no-extraneous-require
+  require('next/dist/next-server/lib/constants')
+  // eslint-disable-next-line no-empty
+} catch (error) {}
```
ehmicky, line 5: Do we need with `try`/`catch` block. I am assuming this file is read but never executed?

src/dependencies.js
ehmicky: Would the following idea help with the brittleness? 

Only keep in that list the files that directly required, not indirectly.
When a Function is directly requiring one of those files, it would be tree-shaked. However, any of the subsequent indirect `require()` within the same Node module would also be tree-shaked. 

This would require keeping some state indicating "currently tree-shaking Node module X" so the logic knows to keep tree-shaking any files from that Node module.

I am not sure whether this makes sense in our case. Would this help reduce the size of this list?
What do you think?

erezrokah: That's a good suggestion. I did another test and everything on that list that's under `next/dist/next-server` is required directly (`next/dist/next-server/lib/i18n/normalize-locale-path` only when using `Next.js` 10).

Let me try you suggestion and see how it goes

src/fixtures/node-module-next/function.js
erezrokah: No we don't, see https://github.com/netlify/zip-it-and-ship-it/pull/242/commits/61fc87169ec4c773840551b7ff670667f9dd102f

src/dependencies.js
erezrokah: @ehmicky let me know if https://github.com/netlify/zip-it-and-ship-it/pull/242/commits/9716a604b1472e6e03b60cab05123fe425e91d2d is what you had in mind.
This works well for `Next.js` v9 and v10

src/dependencies.js
```diff
@@ -84,33 +84,66 @@ const getFileDependencies = async function(path, packageJson, state) {
...
+
+const getImportDependencies = function({ dependency, basedir, packageJson, state, treeShakingModule }) {
+  if (LOCAL_IMPORT_REGEXP.test(dependency) || isTreeShakingModule({ dependency, treeShakingModule })) {
+    return getTreeShakedDependencies({ dependency, basedir, packageJson, state, treeShakingModule })
+  }
+
+  if (isTreeShakable(dependency)) {
+    const module = getModuleName(dependency)
+    return getTreeShakedDependencies({ dependency, basedir, packageJson, state, treeShakingModule: module })
   }
```
ehmicky, line 106: [sand] Would the following be simpler?

```js
const isTreeShakingModule = function(moduleName, treeShakingModule) {
  return treeShakingModule === moduleName
}

const getImportDependencies = function({ dependency, basedir, packageJson, state, treeShakingModule }) {
  const moduleName = getModuleName(dependency)
  if (LOCAL_IMPORT_REGEXP.test(dependency) || isTreeShakingModule(moduleName, treeShakingModule) || isTreeShakable(dependency)) {
    return getTreeShakedDependencies({ dependency, basedir, packageJson, state, treeShakingModule: moduleName })
  }

  return getAllDependencies(dependency, basedir, state, packageJson)
}
```

src/dependencies.js
```diff
@@ -84,33 +84,66 @@ const getFileDependencies = async function(path, packageJson, state) {
...
+  return treeShakingModule === getModuleName(dependency)
+}
+
+const getImportDependencies = function({ dependency, basedir, packageJson, state, treeShakingModule }) {
+  if (LOCAL_IMPORT_REGEXP.test(dependency) || isTreeShakingModule({ dependency, treeShakingModule })) {
+    return getTreeShakedDependencies({ dependency, basedir, packageJson, state, treeShakingModule })
+  }
+
+  if (isTreeShakable(dependency)) {
+    const module = getModuleName(dependency)
```
ehmicky, line 104: We might want not to use `module` as a variable name, due to its usage in Node.js, what do you think?
Note that this should not be a problem in this particular case due to variable shadowing.
For example, `moduleName` could work instead?

src/dependencies.js
ehmicky: Thanks!
Are all of the following entry points used in the Next.js plugin?

src/dependencies.js
erezrokah: Sure, I'm good either way

src/dependencies.js
erezrokah: Those are all direct requires in `nextPage.js` (v9 or v10 and page or api)

src/dependencies.js
erezrokah: Update let me just check passing `moduleName` for local imports doesn't break anything

erezrokah: Update; It does break as it stops tree shaking when switching between local imports inside a `next/*` `require`

erezrokah: Is this https://github.com/netlify/zip-it-and-ship-it/pull/242/commits/d4edf4930de0f8f2f48e52fb5cdf159006a82f1a better?

ehmicky: Yes, this works! :+1:

src/dependencies.js
ehmicky: Discussed on Slack: what about this:
If we detect that a Function was bundled by `next-on-netlify`, then tree-shake _any_ `next/*` module.
We would need to have some way to detect that a Function was bundled by `next-on-netlify`.
But this would allow us not to rely on specific file paths.

src/dependencies.js
```diff
@@ -84,33 +84,64 @@ const getFileDependencies = async function(path, packageJson, state) {
...
+  'next/dist/next-server/lib/constants',
+  'next/dist/next-server/lib/constants.js',
+  'next/dist/next-server/lib/document-context.js',
+  'next/dist/next-server/lib/head.js',
+  'next/dist/next-server/lib/i18n/normalize-locale-path',
+  'next/dist/next-server/lib/router/utils/get-route-from-asset-path',
+  'next/dist/next-server/lib/router/utils/path-match',
+  'next/dist/next-server/lib/router/utils/prepare-destination',
+  'next/dist/next-server/lib/utils',
+  'next/dist/next-server/lib/utils.js',
```
ehmicky, line 125: Note: we could strip the file extension first, so we don't need to have duplicate entries with/without the file extension.

src/dependencies.js
```diff
@@ -84,25 +84,38 @@ const getFileDependencies = async function({ path, packageJson, state }) {
...
+  const shouldTreeShakeNext = treeShakeNext || isNextOnNetlify(dependency)
+  if (shouldTreeShake(dependency, shouldTreeShakeNext)) {
+    return getTreeShakedDependencies({ dependency, basedir, packageJson, state, treeShakeNext: shouldTreeShakeNext })
   }
 
   return getAllDependencies({ dependency, basedir, state, packageJson })
 }
 
+const isNextOnNetlify = function(dependency) {
+  return dependency === './renderNextPage'
```
ehmicky, line 102:
```suggestion
  return basename(dependency, '.js') === 'renderNextPage'
```

To make it more resilient to changes in `next-on-netlify` related to either the optional file extension, or the directories (but not the filename itself). What do you think?

src/dependencies.js
```diff
@@ -84,25 +84,38 @@ const getFileDependencies = async function({ path, packageJson, state }) {
...
+}
+
+const shouldTreeShake = function(dependency, treeShakeNext) {
```
ehmicky, line 105: We might want to add some code comment to explain why we are doing this?

ehmicky (approved):
:tada:

src/dependencies.js
erezrokah: Let me know if this https://github.com/netlify/zip-it-and-ship-it/pull/242/commits/5ed3906dbf853073c0ed4a393935efc7f76a3caf is clear enought

ehmicky: `since there is no guarantee that would work for every module`: maybe we should explain that this is because some Node.js modules either: 1. do dynamic requires or 2. read files, which would not work with our basic tree shaking logic.

Apart from this, looks good :+1:

erezrokah: Done in https://github.com/netlify/zip-it-and-ship-it/pull/242/commits/6f4a9e0912131ed85d22f0dabd3d528a0d13cefd

PR #255, opened 2020-12-07
Use native promisify instead of util.promisify

realityking:
It's available since Node 8.0

**- Summary**

Reduce unneeded dependencies by relying on built-in functionality. `util.promisify` is a surprisingly heavy dependency since it relies on `es-abstract`.

**- Test plan**

Test suite still passes, no change of the interface.

Reduce dependencies by using native promisify function


ehmicky (approved):
Thanks a lot @realityking!

Would it be possible to rebase this PR onto the latest `master`? Thanks!

realityking:
@ehmicky done

PR #256, opened 2020-12-10
Add metadata file to binary zip files.

calavera:
This metadata file, called "netlify-toolchain" allows adding extra information about how the file was generated and what's in it without having to check other files.

This will help fixing https://github.com/netlify/build/issues/2114

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/src/binary.js
+++ b/src/binary.js
@@ -26,7 +26,7 @@ const RUNTIMES = {
 const zipBinary = async function(srcPath, destPath, filename, stat, runtime) {
   const { archive, output } = startZip(destPath)
   addZipFile(archive, srcPath, filename, stat)
-  addZipContent(archive, `runtime=${runtime}`, "netlify-toolchain")
+  addZipContent(archive, `runtime=${runtime}`, 'netlify-toolchain')
   await endZip(archive, output)
 }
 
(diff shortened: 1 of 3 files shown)


src/binary.js
```diff
@@ -23,9 +23,10 @@ const RUNTIMES = {
 }
 
 // Zip a binary function file
-const zipBinary = async function(srcPath, destPath, filename, stat) {
+const zipBinary = async function(srcPath, destPath, filename, stat, runtime) {
   const { archive, output } = startZip(destPath)
   addZipFile(archive, srcPath, filename, stat)
+  addZipContent(archive, `runtime=${runtime}`, 'netlify-toolchain')
```
mraerino, line 29: as noted in https://github.com/netlify/open-api/pull/272 i'd rather use json here

src/main.test.js
```diff
@@ -270,6 +269,11 @@ test('Zips Go function files', async t => {
     const { stdout } = await execa(unzippedFile)
     t.is(stdout, 'test')
   }
+
+  const tcFile = `${tmpDir}/netlify-toolchain`
+  await pathExists(tcFile)
```
ehmicky, line 274:
```suggestion
  t.true(await pathExists(tcFile))
```

Otherwise, this statement is a noop.

ehmicky (commented):
Looks good to me, thanks for the finding the problem and the solution to this @calavera!

ehmicky (approved):

mraerino (approved):
:+1:

PR #265, opened 2020-12-16
chore: add commit linting

erezrokah:
This allows automatic version bumping


ehmicky (approved):
Should we add this to `eslint-config-node` too?
Since this is linting (although not code linting).

PR #268, opened 2020-12-17
Use `@netlify/eslint-config-node`

ehmicky:
This adds `@netlify/eslint-config-node`.


package.json
```diff
@@ -65,34 +73,29 @@
...
-    "eslint-plugin-import": "^2.22.0",
-    "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-prettier": "^3.1.4",
     "execa": "^3.4.0",
     "get-stream": "^5.1.0",
     "husky": "^4.2.5",
     "npm-run-all": "^4.1.5",
     "nyc": "^15.0.0",
-    "path-exists": "^4.0.0",
-    "prettier": "^1.19.1",
```
ehmicky, line 92: `common-path-prefix` and `path-exists` are already production dependencies.

package.json
```diff
@@ -65,34 +73,29 @@
     "yargs": "^15.4.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.10.5",
-    "@babel/plugin-transform-runtime": "^7.10.5",
-    "@babel/preset-env": "^7.10.4",
-    "@babel/runtime": "^7.10.5",
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
```
erezrokah, line 77: I think we can remove `@commitlint/cli` and `@commitlint/config-conventional` too.

erezrokah (approved):
🚀

package.json
ehmicky: Done :+1:

PR #269, opened 2020-12-17
Move test files

ehmicky:
There is some merit to putting source files and test files alongside each other.
However, in this specific repository, I have always found it somewhat confusing, especially with the `helpers` and `fixtures` directories.
Our ESLint configuration is mostly geared towards a structure where source files and test files are in separate directories at the moment. We might want to improve this in the future, but at the moment, splitting those helps with fixing some linting rules (`node/no-unpublished-require`).

This PR moves the test files to a `tests` directory.

Note: currently looking at failing CI tests.


package.json
```diff
@@ -7,10 +7,7 @@
     "zip-it-and-ship-it": "src/bin.js"
   },
   "files": [
-    "src/**/*.js",
-    "!src/fixtures",
-    "!src/helpers",
-    "!src/**/**.test.js"
+    "src/**/*.js"
```
ehmicky, line 10: I checked with `npm publish --dry`: the list of published files is the same.

erezrokah (approved):

ehmicky:
@emilyzhang It looks like the Fossa GitHub actions failed due to files being renamed. Do you know what the fix might be?

PR #277, opened 2020-12-17
Enable some ESLint rules against dynamic `require()`

ehmicky:
This enables some ESLint rules against dynamic `require()`.


tests/fixtures/invalid-require/function.js
```diff
@@ -1,6 +1,6 @@
 // eslint-disable-next-line no-constant-condition
 if (false) {
-  // eslint-disable-next-line import/no-unresolved, import/no-unassigned-import
+  // eslint-disable-next-line import/no-unresolved, import/no-unassigned-import, node/global-require
```
ehmicky, line 3: We need dynamic `require()` in all those tests.

erezrokah (approved):

Issue #68, opened 2019-11-05
This does not work with dynamic or conditional module imports


ehmicky:
Some Node modules do this:

```js
try {
  require('moduleName')
} catch (error) {}
```

Or this:

```js
if (condition) {
  require('moduleName')
}
```

[Or this](https://github.com/ardatan/graphql-toolkit/blob/e6acc231db1b6ac1a11aec33a1ddf55f0f45337e/packages/loaders/code-file/src/extract-document-string-from-code-file.ts#L52):

```js
eval(`require('moduleName')`)
```

Or this:

```js
require(getModuleName())
```

A common example is `node-fetch` which [conditionally require `encoding`](https://github.com/bitinn/node-fetch/blob/8c197f8982a238b3c345c64b17bfa92e16b4f7c4/src/body.js#L14). `encoding` is only used with the `textConverted` method, which [throws if it's missing](https://github.com/bitinn/node-fetch/blob/8c197f8982a238b3c345c64b17bfa92e16b4f7c4/src/body.js#L285).

Another example is `@nestjs/graphql` which [uses `apollo-server-core`](https://github.com/nestjs/graphql/blob/f807e29bce2cdb6254b06b4be5e531e5b7b7e26c/lib/graphql-definitions.factory.ts#L2) but does not declare it in its [production `dependencies`](https://github.com/nestjs/graphql/blob/f807e29bce2cdb6254b06b4be5e531e5b7b7e26c/package.json#L41).

The reason why some modules might want to do this and not use `optionalDependencies` is to allow users to opt-in to installing specific modules. 

However this creates the following two issues with `zip-it-and-ship-it`. When the conditional `require()` is performed: 
  1. Directly from a function file, `zip-it-and-ship-it` always bundles the dependency. Users should have a way to exclude such modules in the archive if they want to.
  2. From a node module, `zip-it-and-ship-it` never bundles the dependency. This is because we find nested dependencies by only checking the `package.json` `dependencies`, `peerDependencies` and `optionalDependencies` keys (as opposed to look for `require()` statements). Users should have a way to include such modules in the archive if they want to. 

The current workaround are (for the points above):
  1. User should `npm install` the dependency
  2. User should `npm install` the dependency and add a noop `require()` call in its function file code.

However this feels hacky, so we should think of a better solution.

DavidWells:
Almost all function bundling solutions have the concept of `includes` & `excludes` options

[...]

PR #295, opened 2021-01-11
fix: don't fail if critters is missing on Next.js>=10.0.4

erezrokah:
**- Summary**

Related to https://github.com/netlify/next-on-netlify/issues/103#issuecomment-755317894
Doesn't fix the original issue as the original issue is about a missing `react-native` module (we don't have a reproduction for it).

This is a result of a [recent change](https://github.com/vercel/next.js/commit/6e4632efe18ea4b65a213f385524d5cc20f25354) in Next.js

**Note:** This use case is not covered by the logic that makes https://github.com/netlify/zip-it-and-ship-it/blob/1e248c6986d43260ac53e18dfa07ab1865589e41/tests/main.js#L127 pass, since we tree shake `Next.js` dependencies instead of copying published files, side files and nested modules.

A better solution would be to support configuring external modules (see https://github.com/netlify/zip-it-and-ship-it/issues/68) so the Next.js plugin (or any other plugin) can generate a relevant bundling configuration based on the framework.
That bundling configuration can change based on the site configuration, for example if the user has `optimizeCss` configured for their Next.js site, we would like to bundle the `critter` dependency.

Regardless, this is an [issue](https://github.com/vercel/next.js/issues/20742) with Next.js as `critters` should be an optional dependency.

**- Test plan**

Added unit tests

Don't fail when `critters` module is missing on `Next.js>=10.0.4`

image


src/node_dependencies/nested.js
```diff
@@ -60,4 +65,17 @@ const isOptionalModule = function (
...
 }
 
+// 'critters' is used only in Next.js >= 10.0.4 when enabling an experimental option and has to be installed manually
+// we ignore it if it's missing
+const isExternalModule = function (moduleName, { dependencies = {}, devDependencies = {} }) {
+  if (moduleName !== 'critters') {
+    return false
+  }
+  const nextVersion = dependencies.next || devDependencies.next
+  if (!validRange(nextVersion)) {
```
ehmicky, line 75: `satisfies()` returns `false` if `nextVersion` is not a valid version, so this can probably be removed?

src/node_dependencies/nested.js
```diff
@@ -60,4 +65,17 @@ const isOptionalModule = function (
   )
 }
 
+// 'critters' is used only in Next.js >= 10.0.4 when enabling an experimental option and has to be installed manually
+// we ignore it if it's missing
+const isExternalModule = function (moduleName, { dependencies = {}, devDependencies = {} }) {
```
ehmicky, line 70: Should we abstract this by taking the module names and versions from some constant variable such as:

```js
const EXTERNAL_MODULES = {
  critters: { next: '>= 10.0.4' }
}
```

Alternatively we could also wait for another case of external module to abstract, but then maybe we should rename `isExternalModule()` to `isExternalCrittersModule()`?

ehmicky (approved):

src/node_dependencies/nested.js
erezrokah: > Alternatively we could also wait for another case of external module to abstract

Let's wait for another use case.
Not sure we want this to be a long term fix anyway.

PR #299, opened 2021-01-11
fix: match Next.js exact versions when ignoring critters

erezrokah:
Follow up on https://github.com/netlify/zip-it-and-ship-it/pull/295

Current logic doesn't work for exact versions.

Not sure if there's a better way to do this.


src/node_dependencies/nested.js
```diff
@@ -72,11 +72,15 @@ const isExternalCrittersModule = function (moduleName, { dependencies = {}, devD
     return false
   }
   const nextVersion = dependencies.next || devDependencies.next
-  if (!validRange(nextVersion)) {
-    return false
+
+  // can the declared Next.js version resolve to >=10.0.4 ?
+  // test exact versions
+  if (validVersion(nextVersion)) {
```
erezrokah, line 78: An exact version is a valid range so this needs to come before `validRange`

ehmicky (approved):

PR #300, opened 2021-01-12
chore: add PR labeler

erezrokah:
Add an action to label PRs based on title prefix


erezrokah:
Finally got this working using a matrix (had to remove the `${{ }}` in the `if` condition).

JGAntunes (approved):
omg

I must confess I'm overly excited about this 😂 this is super useful! Thanks 🙏

PR #301, opened 2021-01-12
chore: add prepublish script

erezrokah:
Enforces syncing dependencies and passing tests before publishing a package.


JGAntunes (approved):
👍

PR #302, opened 2021-01-20
feat: allow to configure Node.js resolve paths

erezrokah:
An initial attempt at fixing https://github.com/netlify/pod-the-builder/issues/116

Probably needs some cleanup and more testing, but would like to get some feedback on the approach.

~~We can also just resolve from `.netlify/plugins/node_modules` instead of having a configurable array (since that should be the only `external` additional path).
Also, we could provide a default `nodeResolvePaths` value so consumers don't have to add `.netlify/plugins/node_modules` manually.~~

Updated the PR to look for `.netlify/plugins/node_modules` up from the `srcDir` instead of it being configurable.

## TODO
- [x] More testing - specifically with the Next.js plugin version that created the `jimp` [issue](https://github.com/netlify/netlify-plugin-nextjs/issues/85)
- [x] ~~Support configuring `nodeResolvePaths` argument in the CLI~~
- [x] ~~Update docs~~

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -10,7 +10,9 @@ const { listNodeFiles } = require('./node_dependencies')
 const { zipBinary } = require('./runtime')
 const { zipNodeJs } = require('./zip_node')
 
-const getPluginsModulesPath = (srcDir) => findUp('.netlify/plugins/node_modules', { cwd: srcDir, type: 'directory' })
+const AUTO_PLUGINS_DIR = '.netlify/plugins/'
+
+const getPluginsModulesPath = (srcDir) => findUp(`${AUTO_PLUGINS_DIR}node_modules`, { cwd: srcDir, type: 'directory' })
 
 // Zip `srcFolder/*` (Node.js or Go files) to `destFolder/*.zip` so it can be
 // used by AWS Lambda
--- a/src/main.js
+++ b/src/main.js
@@ -1,6 +1,7 @@
 const { join, basename, extname } = require('path')
 
 const cpFile = require('cp-file')
+const findUp = require('find-up')
 const makeDir = require('make-dir')
 const pMap = require('p-map')
 
@@ -9,19 +10,21 @@ const { listNodeFiles } = require('./node_dependencies')
 const { zipBinary } = require('./runtime')
 const { zipNodeJs } = require('./zip_node')
 
+const getPluginsModulesPath = (srcDir) => findUp('.netlify/plugins/node_modules', { cwd: srcDir, type: 'directory' })
+
 // Zip `srcFolder/*` (Node.js or Go files) to `destFolder/*.zip` so it can be
 // used by AWS Lambda
 // TODO: remove `skipGo` option in next major release
 const zipFunctions = async function (
   srcFolder,
   destFolder,
-  { parallelLimit = DEFAULT_PARALLEL_LIMIT, skipGo, zipGo, nodeResolvePaths = [] } = {},
+  { parallelLimit = DEFAULT_PARALLEL_LIMIT, skipGo, zipGo } = {},
 ) {
-  const srcPaths = await getSrcPaths(srcFolder)
+  const [srcPaths, pluginsModulesPath] = await Promise.all([getSrcPaths(srcFolder), getPluginsModulesPath(srcFolder)])
 
   const zipped = await pMap(
     srcPaths,
-    (srcPath) => zipFunction(srcPath, destFolder, { skipGo, zipGo, nodeResolvePaths }),
+    (srcPath) => zipFunction(srcPath, destFolder, { skipGo, zipGo, pluginsModulesPath }),
     {
       concurrency: parallelLimit,
     },
@@ -34,7 +37,7 @@ const DEFAULT_PARALLEL_LIMIT = 5
 const zipFunction = async function (
   srcPath,
   destFolder,
-  { skipGo = true, zipGo = !skipGo, nodeResolvePaths = [] } = {},
+  { skipGo = true, zipGo = !skipGo, pluginsModulesPath: defaultModulesPath } = {},
 ) {
   const { runtime, filename, extension, srcDir, stat, mainFile } = await getFunctionInfo(srcPath)
 
@@ -42,7 +45,9 @@ const zipFunction = async function (
     return
   }
 
-  const srcFiles = await getSrcFiles({ runtime, stat, mainFile, extension, srcPath, srcDir, nodeResolvePaths })
+  const pluginsModulesPath =
+    defaultModulesPath === undefined ? await getPluginsModulesPath(srcPath) : defaultModulesPath
+  const srcFiles = await getSrcFiles({ runtime, stat, mainFile, extension, srcPath, srcDir, pluginsModulesPath })
 
   await makeDir(destFolder)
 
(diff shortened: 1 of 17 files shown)


src/main.js
```diff
@@ -15,26 +15,30 @@ const { zipNodeJs } = require('./zip_node')
 const zipFunctions = async function (
   srcFolder,
   destFolder,
-  { parallelLimit = DEFAULT_PARALLEL_LIMIT, skipGo, zipGo } = {},
+  { parallelLimit = DEFAULT_PARALLEL_LIMIT, skipGo, zipGo, nodeResolvePaths = [] } = {},
```
ehmicky, line 18: Note: we need to pass `{buildDir}/.netlify/plugins/node_modules` from both:
  - `@netlify/build` (core functions bundling)
  - `@netlify/functions-util`

src/main.js
```diff
@@ -101,24 +114,37 @@ const listFunctions = async function (srcFolder) {
 }
 
 // List all Netlify Functions files for a specific directory
-const listFunctionsFiles = async function (srcFolder) {
+const listFunctionsFiles = async function (srcFolder, { nodeResolvePaths = [] } = {}) {
```
ehmicky, line 117: Should `nodeResolvePaths` be added to `listFunctions()` too?

src/zip_node.js
```diff
@@ -43,19 +43,27 @@ const addStat = async function (srcFile) {
...
   const pathA = normalize(path)
-  const pathB = pathA.replace(commonPrefix, `${ZIP_ROOT_DIR}${sep}`)
-  const pathC = unixify(pathB)
-  return pathC
+
+  // additional prefixes are used to write the dependency in the correct location in the zip file
+  // e.g. if we resolved a dependency from .netlify/plugins/node_modules/<path>, the target path should be
+  // src/node_modules/<path>
+  const pathB = additionalPrefixes.reduce(
+    (acc, prefix) => acc.replace(normalize(prefix), `${ZIP_ROOT_DIR}${sep}node_modules`),
```
ehmicky, line 61: Would `acc.replace(normalize(prefix), 'node_modules'),` be enough?
Since `${ZIP_ROOT_DIR}${sep}` is added by the next statement.

src/node_dependencies/resolve.js
```diff
@@ -42,20 +43,44 @@ const REQUEST_RESOLVE_MIN_VERSION = '8.9.0'
...
+    // report the first error
+    return { error: previousValue.error || error }
+  })
+
+  const { resolvedLocation, error } = await pWaterfall(tasks, {})
+  if (resolvedLocation !== undefined) {
+    return resolvedLocation
+  }
+
+  throw error
```
ehmicky, line 79: [sand] Would the following work?

Keeping the original `resolve` and `reject` from `resolvePathPreserveSymlinksForDir()` (so it throws on error) then:

```js
const resolvePathPreserveSymlinks = async function (path, baseDirs) {
  for (const baseDir of baseDirs) {
    const resolvedLocation = await resolvePathPreserveSymlinksForDir(path, basedir)
    if (resolvedLocation !== undefined) { 
      return resolvedLocation
    }
  }
}
```

src/main.js
erezrokah: Fixed in https://github.com/netlify/zip-it-and-ship-it/pull/302/commits/682cfdd54b1b6b9d657bf98fb6343696940178e6

src/node_dependencies/resolve.js
erezrokah: This should work:
```js
// eslint-disable-next-line fp/no-let
let firstError
// eslint-disable-next-line fp/no-loops
for (const basedir of baseDirs) {
  try {
    // eslint-disable-next-line no-await-in-loop
    return await resolvePathPreserveSymlinksForDir(path, basedir)
  } catch (error) {
    // eslint-disable-next-line fp/no-mutation
    firstError = firstError || error
  }
}

throw firstError
```

I took the first approach to follow our linting rules, but I think the one you suggested is easier to read. WDYT?

src/zip_node.js
erezrokah: There's actually a bug here I noticed thanks to your comment.
The prefixes don't have to be under the `srcDir` (the now removed use case from #116), so the `commonPrefix` approach won't work in that case.

We could either think of a way to support it or define that `nodeResolvePaths` are relative to the `srcDir`. The latter would make it easier to provide a default. WDYT?

src/node_dependencies/resolve.js
ehmicky: Oh right, I forgot the first error is better reported than the later ones. Sounds good!

erezrokah: Done in https://github.com/netlify/zip-it-and-ship-it/pull/302/commits/585c3e262aca5b856c107bc376550476e358d464

src/zip_node.js
ehmicky: `commonPrefix` should actually work even when one of the required file paths is not under the repository root. If I'm not mistaken, the logic would work even when requiring a different top-level directory (in that case, `commonPrefix` would be `/`).

Would adding `prefixes` to the `commonPathPrefix()` call work?

src/main.js
erezrokah: Actually I think `listFunctions` doesn't need it as it doesn't retrieve the source files.

erezrokah: Actually I think since `listFunctions` doesn't resolve source files we don't need it.

src/main.js
erezrokah: Also to https://github.com/netlify/js-client/blob/446c04f078c6044863c258aad5dab6e0d1970077/src/deploy/hash_fns.js#L15

erezrokah:
FYI this works. See image loaded in https://60097eb6ee8655dae9885441--erez-test-next-bundling.netlify.app/test.
That site was create by forcing `@netlify/build` to use version `1.0.6` of the plugin and telling `zisi` to resolve from `.netlify/plugins/node_modules`.

I also had to fix this line
https://github.com/netlify/next-on-netlify/blob/b4c85b9544e3608dda6b9e623c970d5f8c0ed50c/lib/templates/imageFunction.js#L10
in the generated function since `process.env.URL` gives you the URL of the main site so it didn't work for deploy previews.

I'm getting the host from the event request `event.headers.host` and using that to construct the image URL. We probably want to use `DEPLOY_URL` (see https://docs.netlify.com/configure-builds/environment-variables/#deploy-urls-and-metadata), but since I'm testing locally that didn't work for me.

ehmicky (approved):

erezrokah:
Following an internal discussion updated the PR so it doesn't require changes in `@netlify/build` or `js-client`.

src/node_dependencies/resolve.js
```diff
@@ -42,20 +42,40 @@ const REQUEST_RESOLVE_MIN_VERSION = '8.9.0'
 // We need to use `new Promise()` due to a bug with `utils.promisify()` on
 // `resolve`:
 //   https://github.com/browserify/resolve/issues/151#issuecomment-368210310
-const resolvePathPreserveSymlinks = function (path, basedir) {
+const resolvePathPreserveSymlinksForDir = function (path, basedir) {
   return new Promise((resolve, reject) => {
     resolveLib(path, { basedir, preserveSymlinks: true }, (error, resolvedLocation) => {
       if (error) {
-        return reject(error)
+        reject(error)
```
ehmicky, line 49:
```suggestion
        return reject(error)
```

ehmicky (approved):
Looks good! :tada:

PR #310, opened 2021-01-28
chore: support Node 15

ehmicky:
Part of https://github.com/netlify/team-dev/issues/1

This tests Node 15 in CI.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node: [8.17.0, 14.x]
+        node: [8.17.0, 15.5.0]
         exclude:
           - os: macOS-latest
             node: 8.17.0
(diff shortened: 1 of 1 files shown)


ehmicky:
It seems like creating the zip archive fails on [Node `15.6.0`](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V15.md#15.6.0) but not on Node `15.5.0`.
What's odd is that most of the files in the archive are correct, except the entry point.
The entry point appears to be a corrupted archive. I am wondering if there is some bug in

erezrokah:
> The entry point appears to be a corrupted archive. I am wondering if there is some bug in Node.js? Upgrading a minor version should never create failures.

That's very interesting - especially since our customers should experience the same

ehmicky:
In the meantime, I think we should still be able to merge this, what do you think?

erezrokah:
> In the meantime, I think we should still be able to merge this, what do you think?

Yes I agree.

erezrokah (approved):

ehmicky:
Issue for Node `15.6.0` at https://github.com/netlify/zip-it-and-ship-it/issues/311

PR #314, opened 2021-01-29
fix: add `prisma` CLI to EXCLUDED_PEER_DEPENDENCIES

Jolg42:
**- Summary**

Prisma is going to rename the CLI package from `@prisma/cli` to `prisma`, so the excluded list needs to be updated with that new package name (and keep the older names for retro-compatibility).

**- Test plan**

add `prisma` CLI to EXCLUDED_PEER_DEPENDENCIES

juuiev6a6oyy

### DIFF
--- a/src/node_dependencies/nested.js
+++ b/src/node_dependencies/nested.js
@@ -18,7 +18,7 @@ const shouldIncludePeerDependency = function (name) {
   return !EXCLUDED_PEER_DEPENDENCIES.has(name)
 }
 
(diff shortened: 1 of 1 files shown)


ehmicky (approved):
Thanks for the PR @joelcloralt!

Jolg42:
That was fastttt! 🚀 Thanks ❤️

Jolg42:
@ehmicky Do you know how long it takes for it to be used on Netlify's servers / deployment? (Is it automated?)

ehmicky:
We try to avoid releases on Fridays, so this should be in production on Monday :+1:

PR #319, opened 2021-02-03
feat: add esbuild bundling option

eduardoboucas:
**- Summary**

This PR is an initial attempt at introducing esbuild for function bundling.

For now, it's turned off by default, so the current bundling mechanism will still be used. To turn it on, a new `use-esbuild` flag has been introduced.

A second flag (`external-modules`) has been added. This allows developers to flag modules as [external](https://esbuild.github.io/api/#external) with esbuild, excluding them from the bundle. This is required for modules with native bindings, which esbuild can't bundle.

It's important to note that this part of the implementation is not yet complete, because excluding the modules from the bundle is just part of the story. For these modules, we need to fall back to the current mechanism and include their dependencies in a `node_modules` directory, to sit alongside the bundled file.

ZISI will need to be able to understand which modules it needs to include this way, because the transitive dependencies will also need to be included — e.g. if `opencv` is flagged as an external module, `buffers` and `nan` will also need to be part of the `node_modules` directory.

This logic will be added in a subsequent PR. For now, it means that bundling modules with native dependencies will not work with esbuild.

feat: add esbuild bundling option

cheetah-003


src/zip_node.js
```diff
@@ -30,6 +45,42 @@ const zipNodeJs = async function ({ srcFiles, destPath, filename, mainFile, plug
...
+}) {
+  const bundledFilePath = join(destFolder, filename)
+
+  await esbuild.build({
+    bundle: true,
+    entryPoints: [mainFile],
+    external: externalModules,
+    outfile: bundledFilePath,
+    platform: 'node',
+    target: ['es2017'],
```
erezrokah, line 65: [sand] Should this be based on the [runtime version](https://docs.netlify.com/functions/build-with-javascript/#runtime-settings) of the user function?
We could get that value as an option.

erezrokah (approved):
This looks great 🚀 

For configuring bundling options (as opposed to enabling the feature), we can also magically read it from `.netlify/bundle.config.js` or something similar - that will allow plugins to configure it (we already do something similar with `.netlify/plugins/node_modules`), but we can leave it for another iteration.

Also we would like to configure `esbuild` to resolve from `.netlify/plugins/node_modules`, but again this can be done in another iteration (note that this is required for `next/image` to work).

src/zip_node.js
```diff
@@ -30,6 +45,42 @@ const zipNodeJs = async function ({ srcFiles, destPath, filename, mainFile, plug
...
+  })
+
+  const { archive, output } = startZip(destPath)
```
ehmicky, line 68: Should those lines (from `startZip()` to `endZip()`) re-use the similar code above? The only difference is that there is only one `srcFile`, but we could use pass an array with a single element `[srcFile]` so that both functions use the same code. What do you think?

src/zip_node.js
```diff
@@ -30,6 +45,42 @@ const zipNodeJs = async function ({ srcFiles, destPath, filename, mainFile, plug
...
+  const { srcFile, stat } = await addStat(bundledFilePath)
+
+  addEntryFile(destFolder, archive, filename, bundledFilePath)
+
+  zipJsFile({ srcFile, destFolder, pluginsModulesPath, archive, stat })
+
+  await endZip(archive, output)
+
+  try {
```
ehmicky, line 77: Should those lines be in a `try {} finally {}` block so that the `bundleFilePath` is deleted when an exception is thrown during zipping?

ehmicky (commented):
This is looking great! :tada: 

On @erezrokah's comment about configuration: should we open an issue in `netlify/pod-the-builder` about how we want users to configure Functions bundling through `esbuild`? This would allow product (@verythorough) to weigh in since this is user-facing. This would also start raising awareness of the changes to document in the future.

src/zip_node.js
eduardoboucas: Absolutely. This would be an additional parameter which `@netlify/build` would need to pass, right?

eduardoboucas: > Also we would like to configure esbuild to resolve from .netlify/plugins/node_modules, but again this can be done in another iteration (note that this is required for next/image to work).

@erezrokah can you add some more context to this one, please? 🙏🏻

src/zip_node.js
eduardoboucas: That's a good call. I added the `try`/`catch` around `pUnlink` because I don't think we should throw if we fail to delete the temporary fail. If we put these two concepts together, we end up with:

```js
  try {
    await endZip(archive, output)
  } finally {
    try {
      await pUnlink(bundledFilePath)
    } catch (_) {
      // no-op
    }
  }
```

🤔  What do you think?

src/zip_node.js
eduardoboucas: I wanted to keep the two functions (`zipNodeJs` and `zipNodeJsWithEsbuild`) as compartmentalised as possible, at least during this initial phase. Not only because I expect the `zipNodeJsWithEsbuild` to change significantly over the coming weeks as we iron out the implementation, but also because I want to reduce as much as possible the possibility of us breaking the existing bundling behaviour.

I suggest we revisit this once the implementation matures a bit more. Do you think that's acceptable?

ehmicky: Sound good :)

src/zip_node.js
ehmicky: Yes, something like this.
I am wondering whether, to be on the safe side, the other parts of the zipping logic (`addStat()`, `addEntryFile()`, `zipJsFile()`) should be wrapped in it too?

Note: the `try`/`catch` around `pUnlink()` definitely makes sense. :+1:

ehmicky (approved):

src/zip_node.js
eduardoboucas: Agreed! I'll update it. Thanks! 🙌🏻

src/zip_node.js
erezrokah: > Absolutely. This would be an additional parameter which `@netlify/build` would need to pass, right?

Yes I think so - pending the discussion on how to configure bundling.

> @erezrokah can you add some more context to this one, please? 🙏🏻

Of course! See https://github.com/netlify/zip-it-and-ship-it/pull/302 and https://github.com/netlify/pod-workflow/issues/116

PR #323, opened 2021-02-10
feat: improvements to esbuild bundling

eduardoboucas:
**- Summary**

This PR includes [a bypass](https://github.com/netlify/zip-it-and-ship-it/blob/121bb49456a35ab99139160f3b99e12db1002d0f/src/main.js#L122-L124) for the legacy bundling logic when using esbuild, removing unnecessary computations and thus improving performance.

It also adds support for TypeScript.

Finally, to comply with the `max-lines` ESLint rule, it breaks the Go, JS and Rust runtime-specific logic into different files.

I'm sorry that the PR is larger than what I'd like, but the three changes are heavily interconnected.

feat: improvements to esbuild bundling

maxresdefault

### DIFF
--- a/src/info.js
+++ b/src/info.js
@@ -37,7 +37,7 @@ const getFunctionInfo = async function (srcPath) {
     return {}
   }
 
-  if (['.js', '.ts', '.zip'].includes(extension)) {
+  if (['.js', '.zip'].includes(extension)) {
     return { runtime: 'js', name, filename, stat, mainFile, extension, srcPath, srcDir }
   }
 
--- a/src/info.js
+++ b/src/info.js
@@ -37,7 +37,7 @@ const getFunctionInfo = async function (srcPath) {
     return {}
   }
 
-  if (extension === '.zip' || extension === '.js') {
+  if (['.js', '.ts', '.zip'].includes(extension)) {
     return { runtime: 'js', name, filename, stat, mainFile, extension, srcPath, srcDir }
   }
 
--- a/src/main.js
+++ b/src/main.js
@@ -1,14 +1,12 @@
-const { join, basename, extname } = require('path')
+const { extname } = require('path')
 
-const cpFile = require('cp-file')
 const findUp = require('find-up')
 const makeDir = require('make-dir')
 const pMap = require('p-map')
 
 const { getFunctionInfos, getSrcPaths, getFunctionInfo } = require('./info')
 const { listNodeFiles } = require('./node_dependencies')
-const { zipBinary } = require('./runtime')
-const { zipNodeJs } = require('./zip_node')
+const RUNTIMES = require('./runtimes')
 
 const AUTO_PLUGINS_DIR = '.netlify/plugins/'
 
@@ -49,7 +47,17 @@ const zipFunction = async function (
 
   const pluginsModulesPath =
     defaultModulesPath === undefined ? await getPluginsModulesPath(srcPath) : defaultModulesPath
-  const srcFiles = await getSrcFiles({ runtime, stat, mainFile, extension, srcPath, srcDir, pluginsModulesPath })
+  const srcFiles = await getSrcFiles({
+    runtime,
+    stat,
+    mainFile,
+    extension,
+    srcPath,
+    srcDir,
+    pluginsModulesPath,
+    useEsbuild,
+    externalModules,
+  })
 
   await makeDir(destFolder)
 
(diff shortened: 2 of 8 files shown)


src/info.js
```diff
@@ -37,7 +37,7 @@ const getFunctionInfo = async function (srcPath) {
     return {}
   }
 
-  if (extension === '.zip' || extension === '.js') {
+  if (['.js', '.ts', '.zip'].includes(extension)) {
```
ehmicky, line 40: Should we add `.cjs` and `.mjs`?
There is also the older / less common `.es6`.

src/main.js
```diff
@@ -169,8 +117,12 @@ const getListedFunctionFiles = async function (
   return srcFiles.map((srcFile) => ({ srcFile, name, mainFile, runtime, extension: extname(srcFile) }))
 }
 
-const getSrcFiles = function ({ runtime, stat, mainFile, extension, srcPath, srcDir, pluginsModulesPath }) {
+const getSrcFiles = function ({ runtime, stat, mainFile, extension, srcPath, srcDir, pluginsModulesPath, useEsbuild }) {
   if (runtime === 'js' && extension === '.js') {
```
ehmicky, line 121: Should this be changed to `runtime === 'js' && extension !== '.zip'`? So it works with `.ts`.

src/main.js
```diff
@@ -70,66 +78,6 @@ const zipFunction = async function (
...
-// always be `bootstrap` because they include the
-// Lambda runtime, and that's the name that AWS
-// expects for those kind of functions.
-const zipRustFunction = async function ({ srcPath, destFolder, stat, filename, runtime }) {
-  const destPath = join(destFolder, `${filename}.zip`)
-  await zipBinary({ srcPath, destPath, filename: 'bootstrap', stat, runtime })
-  return destPath
-}
-
-const RUNTIMES = {
```
ehmicky, line 127: Nice splitting into smaller files! :+1:

src/info.js
```diff
@@ -37,7 +37,7 @@ const getFunctionInfo = async function (srcPath) {
...
     return { runtime: 'js', name, filename, stat, mainFile, extension, srcPath, srcDir }
   }
```
ehmicky, line 42: `getMainFile()` should also be fixed to support the new file extensions.

src/zip_node.js
```diff
@@ -69,7 +70,7 @@ const zipNodeJsWithEsbuild = async function ({
     const { archive, output } = startZip(destPath)
     const { srcFile, stat } = await addStat(bundledFilePath)
 
-    addEntryFile(destFolder, archive, filename, bundledFilePath)
+    addEntryFile(destFolder, archive, jsFilename, bundledFilePath)
```
ehmicky, line 73: Should `addEntryFile()` only use `bundleFilePath` instead of `jsFilename`? Since this is the file actually being required.
This would also remove the need to fix the file extension, providing esbuild always outputs a `.js` file.

src/info.js
```diff
@@ -37,7 +37,7 @@ const getFunctionInfo = async function (srcPath) {
     return {}
   }
 
-  if (extension === '.zip' || extension === '.js') {
```
ehmicky, line 40: Should we add some automated tests to cover the new file extension(s)?

src/zip_node.js
eduardoboucas: `bundleFilePath` will contain a full path, whereas `jsFilename` contains the relative path, which is what we want to include in the entry file. Or am I misunderstanding?

src/main.js
eduardoboucas: As per our chat, I'm temporarily reverting TypeScript support until we're comfortable with how to proceed.

src/info.js
eduardoboucas: As per our chat, I'm temporarily reverting TypeScript support until we're comfortable with how to proceed.

src/info.js
eduardoboucas: To revisit once we add support for additional extensions.

src/info.js
eduardoboucas: To revisit once we add support for new extensions.

eduardoboucas:
@ehmicky I've addressed your comments. Also, added support for the esbuild alpha environment variables in https://github.com/netlify/zip-it-and-ship-it/pull/323/commits/c819058bc09e0a6fb431410d47ed5410d5b05e6c.

src/zip_node.js
eduardoboucas: > This would also remove the need to fix the file extension, providing esbuild always outputs a .js file.

You actually need to provide an output file with a `.js` extension so that it builds for the right target. If the output file is `.ts`, it outputs TypeScript.

(There may be an option to modify this behaviour I don't know about.)

ehmicky: Makes sense! :+1:

ehmicky (approved):
:tada:

PR #325, opened 2021-02-12
feat: bundle dependencies with external modules

eduardoboucas:
**- Summary**

This PR changes the behaviour of `externalModules` so that the modules and their dependencies are included alongside the bundled function inside the ZIP file.

A new `ignoredModules` property is introduced, which marks a module as external but does not include the source files in the ZIP.

**- Test plan**

Tests for esbuild. For now, a separate main file (`main_esbuild.js`) is added, but the goal is to go back to having a single file once we finalise the implementation.

There are 7 known failures in the esbuild test file. These relate to Next.js-specific logic that isn't yet replicated in esbuild, because we're still outlining the best way to achieve this.

```
  7 known failures

  main_esbuild › Includes specific Next.js dependencies when using next-on-netlify
  main_esbuild › Includes all Next.js dependencies when not using next-on-netlify
  main_esbuild › Throws on missing critters dependency for Next.js 9
  main_esbuild › Ignore missing critters dependency for Next.js 10
  main_esbuild › Ignore missing critters dependency for Next.js exact version 10.0.5
  main_esbuild › Ignore missing critters dependency for Next.js with range ^10.0.5
  main_esbuild › Ignore missing critters dependency for Next.js with version='latest'
```

feat: bundle dependencies with external modules

e314b6ce5073434d3484b1c5ef06c7ea


erezrokah:
> This PR changes the behaviour of `externalModules` so that the modules and their dependencies are included alongside the bundled function inside the ZIP file.

For the Next.js plugin we would need to support the following use cases:
- If the site has `optimizeCss` enabled bundle `critters`, if not installed fail the bundling process (`critters` won't be configured as external)
- If the site has `optimizeCss` disabled don't bundle `critters` (it should be configured as external but also excluded from the bundle)

See https://github.com/vercel/next.js/blob/5febe218a622022fb375ef5101f1848d9c3120cc/packages/next/build/webpack-config.ts#L770

TLDR: we need to separate the use case of a module being external and should be in included in the bundle and the use case of a module being external and should be excluded from the bundle (should we call these `ignoredModules` instead of `externalModules`?)

eduardoboucas:
This is super useful, thanks @erezrokah! Yes, I think `externalModules` and `ignoredModules` is a good nomenclature.

Btw, I think this distinction is yet another reason why exposing our own configuration API on top of esbuild makes sense.

package.json
```diff
@@ -20,8 +20,8 @@
     "format:check-fix:prettier": "run-e format:check:prettier format:fix:prettier",
     "format:check:prettier": "cross-env-shell prettier --check $npm_package_config_prettier",
     "format:fix:prettier": "cross-env-shell prettier --write $npm_package_config_prettier",
-    "test:dev": "ava",
-    "test:ci": "nyc -r lcovonly -r text -r json ava"
+    "test:dev": "ava --serial",
```
eduardoboucas, line 23: This is temporary while we have two main test files (`main.js` and `main_esbuild.js`). Some of the tests do things like renaming and deleting files, which becomes problematic when the tests run concurrently.

src/bundler.js
```diff
@@ -0,0 +1,41 @@
...
+  // De-duping external and ignored modules.
+  const external = [...new Set([...externalModules, ...ignoredModules])]
+  const jsFilename = `${basename(destFilename, extname(destFilename))}.js`
+  const bundlePath = join(destFolder, jsFilename)
+  const data = await esbuild.build({
+    bundle: true,
+    entryPoints: [srcFile],
+    external,
+    outfile: bundlePath,
+    nodePaths: additionalModulePaths,
```
eduardoboucas, line 26: Allows modules to be resolved from `.netlify/plugins/node_modules`.

src/main.js
```diff
@@ -47,33 +61,32 @@ const zipFunction = async function (
...
   await makeDir(destFolder)
 
-  const destPath = await RUNTIMES[runtime]({
+  // If the file is a zip, we assume the function is bundled and ready to go.
+  // We assume its runtime to be JavaScript and simply copy it to the
+  // destination path with no further processing.
+  if (extension === '.zip') {
+    const destPath = join(destFolder, filename)
+    await cpFile(srcPath, destPath)
+    return { path: destPath, runtime: 'js' }
```
eduardoboucas, line 73: As discussed with @ehmicky, the `runtime` property here should probably be changed to `undefined`. To be done in a separate PR.

src/node_dependencies/index.js
```diff
@@ -35,20 +35,27 @@ const isNotJunk = function (file) {
 }
 
 // Retrieve all the files recursively required by a Node.js file
-const getDependencies = async function (mainFile, srcDir, pluginsModulesPath) {
+const getDependencies = async function (mainFile, srcDir, pluginsModulesPath, moduleAllowList) {
```
eduardoboucas, line 38: We want to get a list of all the files to be included for a given dependency. I've repurposed this method and added a `moduleAllowList` parameter — when defined, only the files belonging to these dependencies will be returned.

src/node_dependencies/module.js
```diff
@@ -0,0 +1,13 @@
...
+const BACKSLASH_REGEXP = /\\/g
+
+// When doing require("moduleName/file/path"), only keep `moduleName`
+const getModuleName = function (dependency) {
+  const dependencyA = dependency.replace(BACKSLASH_REGEXP, '/')
+  const moduleName = requirePackageName(dependencyA)
+  return moduleName
+}
+
+module.exports = { getModuleName }
```
eduardoboucas, line 13: Extracted from `src/node_dependencies/index.js`.

tests/fixtures/local-require-esm/function/.eslintrc
```diff
@@ -0,0 +1,3 @@
+{
```
eduardoboucas, line 1: This is so that ESLint allows us to have a file with ES Modules.

tests/fixtures/node-module-missing-conditional/function.js
```diff
@@ -1,6 +1,6 @@
-try {
-  // eslint-disable-next-line import/no-unassigned-import, import/no-unresolved, node/global-require
-  require('encoding')
-} catch (error) {}
+// eslint-disable-next-line import/no-unresolved
+const module = require('encoding')
+
+console.log(module)
```
eduardoboucas, line 4: Added a `console.log` so that the module isn't removed in the tree shaking phase.

tests/fixtures/node-module-peer-not-optional/function.js
```diff
@@ -1,6 +1,6 @@
-try {
-  // eslint-disable-next-line import/no-unassigned-import, import/no-unresolved, node/global-require
-  require('consistent-ids')
-} catch (error) {}
+// eslint-disable-next-line import/no-unresolved
+const module = require('consistent-ids')
+
+console.log(module)
```
eduardoboucas, line 4: Added a `console.log` so that the module isn't removed in the tree shaking phase.

tests/fixtures/node-module-peer-optional-none/function.js
```diff
@@ -1,6 +1,6 @@
-try {
-  // eslint-disable-next-line import/no-unassigned-import, import/no-unresolved, node/global-require
-  require('consistent-ids')
-} catch (error) {}
+// eslint-disable-next-line import/no-unresolved
+const module = require('consistent-ids')
+
+console.log(module)
```
eduardoboucas, line 4: Added a `console.log` so that the module isn't removed in the tree shaking phase.

tests/main_esbuild.js
```diff
@@ -0,0 +1,465 @@
...
+} = require('./helpers/main')
+const { computeSha1 } = require('./helpers/sha')
+
+const pReadFile = promisify(readFile)
+const pChmod = promisify(chmod)
+const pSymlink = promisify(symlink)
+const pUnlink = promisify(unlink)
+const pRename = promisify(rename)
+
+// Temporary utility method for injecting the `useEsbuild` parameter into every
```
eduardoboucas, line 32: This will become unnecessary once we have a single main test file again.

tests/main_esbuild.js
```diff
@@ -0,0 +1,465 @@
...
+
+test('Does not inline node modules and includes them in a `node_modules` directory if they are defined in `externalModules`', async (t) => {
+  const { tmpDir } = await zipNode(t, 'node-module-included-try-catch', { opts: { externalModules: ['test'] } })
+  const requires = await getRequires({ depth: Number.POSITIVE_INFINITY, filePath: resolve(tmpDir, 'function.js') })
+
+  t.true(requires.includes('test'))
+  t.true(await pathExists(`${tmpDir}/src/node_modules/test`))
+})
+
+test('Does not inline node modules and excludes them from the bundle if they are defined in `ignoredModules`', async (t) => {
```
eduardoboucas, line 65: Tests the new `ignoredModules` parameter.

package.json
ehmicky: Should we use `test.serial()` on each of those tests instead?
This would ensure running as many tests concurrently as possible, to speed them up.

eduardoboucas:
@ehmicky @erezrokah I think this is ready for review.

Apologies for the size of the PR; the bulk of the added lines are tests, since I've added an entirely new main file for esbuild. As for the code itself, I couldn't find a way to break this down into smaller chunks. 😢 

P.S. I don't know what's up with `package-lock.json`. 🤔

src/runtimes/node.js
```diff
@@ -1,38 +1,92 @@
...
     destPath,
     filename,
     mainFile,
     pluginsModulesPath,
-    useEsbuild,
-    externalModules,
+    renames,
+    srcFiles: [...srcFiles, bundlePath],
   })
+  await cleanTempFiles()
```
ehmicky, line 87: Should this be run in `try`/`finally` in case `zipNodeJs()` throws?

ehmicky (approved):
This is looking great, thanks @eduardoboucas!

Do we know why the CI tests are not running?

For `package-lock.json`: this is because Renovate has updated it, so this PR must be rebased onto `master`.

src/runtimes/node.js
```diff
@@ -1,38 +1,92 @@
...
 
-const zipJsFunction = async function ({
-  srcPath,
+const getSrcFiles = function ({ srcPath, mainFile, srcDir, stat, pluginsModulesPath, useEsbuild, externalModules }) {
+  if (!useEsbuild) {
+    return listFilesUsingLegacyBundler({ srcPath, mainFile, srcDir, stat, pluginsModulesPath })
+  }
+
+  if (externalModules.length !== 0) {
+    return getDependencies(mainFile, srcDir, pluginsModulesPath, externalModules)
```
erezrokah, line 15: How is this impacting bundle time? Are we running both the old bundling logic and the new one in this case?
Is the purpose of this code to copy external dependencies? Will a copy files operation work?

package.json
eduardoboucas: I tried that before and it didn't work. I'll give it another try.

erezrokah (changes requested):
Looking great @eduardoboucas.
One question about `externalModules`.

Also would using something similar to https://github.com/netlify/cli/blob/8b0e0ad2b3cc4d9eec20554afb99114fd5320865/tests/command.dev.test.js#L94 work for the tests?

package.json
eduardoboucas: I solved it in https://github.com/netlify/zip-it-and-ship-it/pull/325/commits/e58cb990a00b24adcddbc2f4227670bb9c985744 by using different temporary directory prefixes for the two bundles, and using different fixtures for the problematic tests so that we don't have race conditions on the same files.

src/runtimes/node.js
eduardoboucas: Absolutely. That was removed by accident. Reinstated in https://github.com/netlify/zip-it-and-ship-it/pull/325/commits/bff936d6e81ca31e8c4af6dddfd11ff34d0c6fc5.

src/runtimes/node.js
eduardoboucas: > How is this impacting bundle time?

Every external module will carry a performance penalty, because it needs to fall back to the legacy method of traversing the module's dependency tree using Node. We're now doing for external modules what we were previously doing for every module, so unless we're adding every module as external, we're still improving the current bundle times.

> Are we running both the old bundling logic and the new one in this case?

Not really. Since we're flagging these modules as external, esbuild will simply leave them alone, avoiding the cost of traversing and inlining them. So effectively we're only running the old bundling logic for these modules.

> Is the purpose of this code to copy external dependencies?

Yes.

> Will a copy files operation work?

Yes, but we need to copy _all_ the files, including transitive dependencies, for each of these modules. To do this computation, I figured it would make sense to reuse the logic in `getDependencies`.

eduardoboucas: > Also would using something similar to https://github.com/netlify/cli/blob/8b0e0ad2b3cc4d9eec20554afb99114fd5320865/tests/command.dev.test.js#L94 work for the tests?

Yes, that's a possibility! We have some tests that we only want to run for one of the bundlers, but that's still achievable with that approach. I can revisit this, but I'd probably do it after we iron out the rest of the implementation details, in a separate PR, if that's okay?

eduardoboucas:
> For package-lock.json: this is because Renovate has updated it, so this PR must be rebased onto master.

It was a discrepancy in npm versions. I was able to generate the proper lock file after upgrading to npm v7.

src/runtimes/node.js
erezrokah: Sure, doing it in a separate PR works.

ehmicky (commented):
Note: it looks like some tests are failing in Windows and Node 8.

eduardoboucas:
> Note: it looks like some tests are failing in Windows and Node 8.

Yes, I'm still trying to understand why. I'll look into it this afternoon after I push some additional changes.

src/node_dependencies/tree_shake.js
```diff
@@ -0,0 +1,18 @@
...
+// an exception is when the function was generated by `next-on-netlify` to avoid bundling the entire `Next.js` framework
+const shouldTreeShake = function (dependency, treeShakeNext) {
+  if (LOCAL_IMPORT_REGEXP.test(dependency)) {
+    return true
+  }
+
+  return treeShakeNext && getModuleName(dependency) === 'next'
+}
+
+module.exports = { shouldTreeShake }
```
eduardoboucas, line 18: Extracted from `src/node_dependencies/index.js`.

src/runtimes/node.js
```diff
@@ -32,7 +60,16 @@ const zipFunction = async function ({
   useEsbuild,
 }) {
   const destPath = join(destFolder, `${basename(filename, extension)}.zip`)
-  const srcFiles = await getSrcFiles({
+
+  // When a module is added to `externalModules`, we will traverse its main
```
eduardoboucas, line 64: @erezrokah this is the optimisation we discussed earlier.

src/runtimes/node.js
eduardoboucas: @erezrokah I've changed the bundling mechanism for external modules, as per our conversation, in https://github.com/netlify/zip-it-and-ship-it/pull/325/commits/bd9cad72500b707bd74272144c4be81c296482d7.

src/node_dependencies/traverse.js
```diff
@@ -0,0 +1,142 @@
...
+  )
+  const { moduleNames, paths } = dependencies.reduce(
+    (aggregate, dependency) => {
+      dependency.moduleNames.forEach(aggregate.moduleNames.add, aggregate.moduleNames)
+      dependency.paths.forEach(aggregate.paths.add, aggregate.paths)
+
+      return aggregate
+    },
+    { moduleNames: new Set(), paths: new Set() },
+  )
```
ehmicky, line 64: [sand] Would the following lines be simpler? (feel free to discard this comment)

```js
const moduleNames = new Set(dependencies.map(dependency => dependency.moduleNames))
const paths = new Set(dependencies.map(dependency => dependency.paths))
```

eduardoboucas: Absolutely. Changed in https://github.com/netlify/zip-it-and-ship-it/pull/325/commits/3f0459cea8a2f2bff47faf090aaa638374c6a019.

eduardoboucas:
@erezrokah I've addressed your comments, including reorganising the tests into a single file.

src/node_dependencies/index.js
```diff
@@ -35,20 +38,26 @@ const isNotJunk = function (file) {
...
   const packageJson = await getPackageJson(srcDir)
-
-  const state = { localFiles: new Set(), modulePaths: new Set() }
+  const state = getNewCache()
 
   try {
-    return await getFileDependencies({ path: mainFile, packageJson, state, pluginsModulesPath })
+    return await getFileDependencies({ path: mainFile, moduleAllowList, packageJson, state, pluginsModulesPath })
```
erezrokah, line 46: Can we remove `moduleAllowList`? I couldn't find where it is used

tests/main.js
```diff
@@ -22,405 +22,476 @@ const pSymlink = promisify(symlink)
...
+  const { tmpDir } = await zipNode(t, 'node-module-included-try-catch', { bundler: 'esbuild' })
+  const requires = await getRequires({ filePath: resolve(tmpDir, 'function.js') })
 
-  t.true(files.every(({ runtime }) => runtime === 'rs'))
+  t.false(requires.includes('test'))
+  t.false(await pathExists(`${tmpDir}/src/node_modules/test`))
+})
 
-  await unzipFiles(files)
+test('[bundler: esbuild] Does not inline node modules and includes them in a `node_modules` directory if they are defined in `externalModules`', async (t) => {
```
erezrokah, line 436: These tests are 🔥

erezrokah (approved):
This is 🚀 made a small comment to follow up on

src/bundler.js
```diff
@@ -0,0 +1,48 @@
...
+  srcFile,
+}) {
+  // De-duping external and ignored modules.
+  const external = [...new Set([...externalModules, ...ignoredModules])]
+  const jsFilename = `${basename(destFilename, extname(destFilename))}.js`
+  const bundlePath = join(destFolder, jsFilename)
+
+  // We use the synchronous entry point of the esbuild API because it's more
+  // performant, works with Node 8, and there are no obvious downsides since
+  // we don't need to be performing other work simultaneously.
```
ehmicky, line 24: Isn't each function bundled separately?
In that case, use a synchronous API would bundle each function serially instead of in parallel.

src/node_dependencies/index.js
eduardoboucas: Oh, of course! It's no longer necessary. Thanks! 🙌

src/bundler.js
eduardoboucas: You're absolutely right, this was an oversight on my part. I was focusing on the fact that it's fine to block while the bundling is taking place, because we can't proceed with any other operation until it's complete. But I was forgetting the fact that we might actually be bundling several, independent functions at the same time, in which case spinning different processes (and utilising all the available cores) is much more performant.

Sadly, the async API doesn't work with Node 8, but rather than introducing a performance penalty for everyone, I've added a check for the version being used and switch to the sync API only for Node 8.x: https://github.com/netlify/zip-it-and-ship-it/pull/325/commits/05bedb014f2fe6ca65cdbcdeba70234b68036782.

Thanks @ehmicky!

ehmicky: This looks perfect!

PR #338, opened 2021-02-18
feat: add Next.js special case to esbuild bundler

eduardoboucas:
**- Summary**

This PR adds special cases for the Next.js dependency quirks to the esbuild bundling mechanism.

It also handles the case where a module that isn't installed is included in `externalModules`. A corresponding test for this case is included.

**- Test plan**

Relevant tests were moved from being legacy-specific to the common test pool.

feat: add Next.js special case to esbuild bundler

b7c32f70a50b163bda92b6bcf1956515

### DIFF
--- a/src/node_dependencies/special_cases.js
+++ b/src/node_dependencies/special_cases.js
@@ -1,0 +1,33 @@
+const { getPackageJson } = require('./package_json')
+
+const getPackageJsonIfAvailable = async (srcDir) => {
+  try {
+    const packageJson = await getPackageJson(srcDir)
+
+    return packageJson
+  } catch (_) {
+    return {}
+  }
+}
+
+const getModulesForNextJs = ({ dependencies }) => {
+  const externalModules = dependencies.next ? ['critters', 'nanoid'] : []
+  const ignoredModules = []
+
+  return {
+    externalModules,
+    ignoredModules,
+  }
+}
+
+const getExternalAndIgnoredModulesFromSpecialCases = async ({ srcDir }) => {
+  const { dependencies = {} } = await getPackageJsonIfAvailable(srcDir)
+  const { externalModules, ignoredModules } = getModulesForNextJs({ dependencies })
+
+  return {
+    externalModules,
+    ignoredModules,
+  }
+}
+
+module.exports = { getExternalAndIgnoredModulesFromSpecialCases }
--- a/src/node_dependencies/traverse.js
+++ b/src/node_dependencies/traverse.js
@@ -2,6 +2,7 @@ const { dirname } = require('path')
 
 const { getModuleName } = require('./module')
 const { getNestedDependencies, handleModuleNotFound } = require('./nested')
+const { getPackageJson } = require('./package_json')
 const { getPublishedFiles } = require('./published')
 const { resolvePackage } = require('./resolve')
 const { getSideFiles } = require('./side_files')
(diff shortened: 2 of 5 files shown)


src/node_dependencies/traverse.js
```diff
@@ -53,8 +54,9 @@ const getDependencyNamesAndPathsForDependencies = async function ({
       }),
     ),
   )
-  const moduleNames = new Set(dependencies.flatMap((dependency) => [...dependency.moduleNames]))
-  const paths = new Set(dependencies.flatMap((dependency) => [...dependency.paths]))
+  const filteredDependencies = dependencies.filter(Boolean)
```
erezrokah, line 57: When do we get a falsy value for a `dependency`?

tests/main.js
```diff
@@ -468,30 +496,11 @@ test('[bundler: esbuild] Include most files from node modules present in `extern
...
-    pathExists(`${tmpDir}/src/node_modules/next/dist/compiled/semver.js`),
-    pathExists(`${tmpDir}/src/node_modules/next/dist/other.js`),
-    pathExists(`${tmpDir}/src/node_modules/next/index.js`),
-  ])
-  t.true(constantsExists)
-  t.true(semverExists)
-  t.false(otherExists)
-  t.false(indexExists)
-})
+test('[bundler: esbuild] Does not throw if one of the modules defined in `externalModules` does not exist', async (t) => {
```
erezrokah, line 499: I expect this to throw since:
1. External - don't bundle but  include in zip (so if missing fail).
2. Ignored - don't bundle and don't include in zip (so don't fail if missing).

Does that make sense?

erezrokah (changes requested):
@eduardoboucas Seeing this moving so fast is amazing,
I feel we need to clarify some stuff around `external` vs `ignored`

src/node_dependencies/traverse.js
eduardoboucas: Oops. I redid this part a couple of times with different approaches and this `filter` call was an accidental leftover. Thanks for flagging!

src/node_dependencies/special_cases.js
```diff
@@ -0,0 +1,33 @@
...
+    const packageJson = await getPackageJson(srcDir)
+
+    return packageJson
+  } catch (_) {
+    return {}
+  }
+}
+
+const getModulesForNextJs = ({ dependencies }) => {
+  const externalModules = dependencies.next ? ['critters', 'nanoid'] : []
```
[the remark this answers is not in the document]
eduardoboucas: > Should we check devDependencies too?

I'll add that. 👍

> Also since we want to exclude critters and nanoid from the bundle, should we use ignoredModules

My understanding is that Next.js does use `critters` in certain circumstances. If we add it to `ignoredModules`, its `require` will be left in place but the source will not be added to the bundle, so the `require` call will always throw.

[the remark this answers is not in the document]
erezrokah: Good point - I'm still under the thought this will be configured externally.

tests/main.js
erezrokah: Ok - so per an internal discussion let's not throw on `external` until we expose them as a configuration.

erezrokah (approved):

src/node_dependencies/traverse.js
eduardoboucas: I'm going to remove this in the next PR, if that's okay.

PR #339, opened 2021-02-19
feat: add JS bundler fallback

eduardoboucas:
**- Summary**

This PR prepares zip-it-and-ship-it for the public rollout of esbuild. It includes the following breaking changes from the alpha:

1. A `jsBundler` parameter was introduced. Its possible values are:
   - `zisi` (default): Bundle JS functions with the legacy mechanism
   - `esbuild`: Bundle JS functions with esbuild
   - `esbuild_zisi`: Bundle JS functions with esbuild and fallback to the legacy mechanism if it fails

2. The `useEsbuild` parameter was removed in favour of 1.

3. The `externalModules` parameter was renamed to `jsExternalModules`

4. The `ignoredModules` parameter was renamed to `jsIgnoredModules`

**- Test plan**

- Added a test for a module with native bindings
- Introduced a `testBundlers` helper function, which runs a test for each of the specified JS bundlers

feat: add JS bundler fallback

_114306756_gettyimages-482829557

### DIFF
--- a/src/runtimes/node.js
+++ b/src/runtimes/node.js
@@ -18,7 +18,7 @@ const getSrcFiles = async function (options) {
 }
 
 const getSrcFilesAndExternalModules = async function ({
-  jsBundlerVersion,
+  jsBundler,
   jsExternalModules = [],
   srcPath,
   mainFile,
@@ -26,7 +26,7 @@ const getSrcFilesAndExternalModules = async function ({
   stat,
   pluginsModulesPath,
 }) {
-  if (jsBundlerVersion === JS_BUNDLER_LEGACY) {
+  if (jsBundler === JS_BUNDLER_LEGACY) {
     const paths = await listFilesUsingLegacyBundler({ srcPath, mainFile, srcDir, stat, pluginsModulesPath })
 
     return {
@@ -55,7 +55,7 @@ const zipFunction = async function ({
   destFolder,
   extension,
   filename,
-  jsBundlerVersion,
+  jsBundler,
   jsExternalModules: externalModulesFromConfig = [],
   jsIgnoredModules: ignoredModulesFromConfig = [],
   mainFile,
@@ -81,12 +81,12 @@ const zipFunction = async function ({
     srcPath,
     srcDir,
     pluginsModulesPath,
-    jsBundlerVersion,
+    jsBundler,
     jsExternalModules: externalModulesFromConfig,
   })
   const dirnames = srcFiles.map((filePath) => normalize(dirname(filePath)))
 
-  if (jsBundlerVersion === JS_BUNDLER_LEGACY) {
+  if (jsBundler === JS_BUNDLER_LEGACY) {
     await zipNodeJs({
       basePath: commonPathPrefix(dirnames),
       destFolder,
@@ -97,7 +97,7 @@ const zipFunction = async function ({
       srcFiles,
     })
 
-    return destPath
+    return { bundlerVersion: JS_BUNDLER_LEGACY, path: destPath }
   }
 
   const {
@@ -105,7 +105,7 @@ const zipFunction = async function ({
     ignoredModules: ignoredModulesFromSpecialCases,
   } = await getExternalAndIgnoredModulesFromSpecialCases({ srcDir })
 
-  const { bundlePath, cleanTempFiles } = await bundleJsFile({
+  const { bundlePath, data, cleanTempFiles } = await bundleJsFile({
     additionalModulePaths: pluginsModulesPath ? [pluginsModulesPath] : [],
     destFilename: filename,
     destFolder,
@@ -117,6 +117,7 @@ const zipFunction = async function ({
     ignoredModules: [...ignoredModulesFromConfig, ...ignoredModulesFromSpecialCases],
     srcFile: mainFile,
   })
+  const bundlerWarnings = data.warnings.length === 0 ? undefined : [data.warnings]
 
   // We're adding the bundled file to the zip, but we want it to have the same
   // name and path as the original, unbundled file. For this, we use an alias..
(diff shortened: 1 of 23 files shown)


biilmann:
One thing I noticed when I was playing around with React Server Components over the holidays (painful!) was that to get it working with Netlify, I had to reach for the old [Netlify Lambda](https://github.com/netlify/netlify-lambda) library because the server components relied on a webpack build with a custom plugin. I wonder if it would make sense to build in an escape hatch here to specify a custom bundler for functions? Maybe rather than a bundler version (1, 2, ...) a bundler name ("esbuild", "zipnship", "...")?

eduardoboucas:
> One thing I noticed when I was playing around with React Server Components over the holidays (painful!) was that to get it working with Netlify, I had to reach for the old [Netlify Lambda](https://github.com/netlify/netlify-lambda) library because the server components relied on a webpack build with a custom plugin. I wonder if it would make sense to build in an escape hatch here to specify a custom bundler for functions? Maybe rather than a bundler version (1, 2, ...) a bundler name ("esbuild", "zipnship", "...")?

Interesting. I was torn between using version numbers or bundler names, but that feels like a compelling reason to use the names. Thanks @biilmann!

src/runtimes/node.js
```diff
@@ -116,30 +117,47 @@ const zipFunction = async function ({
...
+    return zipFunction(parameters)
+  }
+
+  // Otherwise, we'll try to bundle with v2 and, if that fails, fallback to v1.
+  try {
+    return await zipFunction({ ...parameters, jsBundler: JS_BUNDLER_ESBUILD })
+  } catch (error) {
+    const data = await zipFunction({ ...parameters, jsBundler: JS_BUNDLER_LEGACY })
+
+    return { ...data, bundlerErrors: error.errors }
```
ehmicky, line 159: Are there some `esbuild` exceptions which do not include the `errors` property? If so, do we want to capture those too?

Generally speaking, should we propagate any exception from `zipFunction()`? In other words, returning `error` instead of `error.errors`. Otherwise, we would not get any error message nor stack trace form bugs in `zip-it-and-ship-it`.

src/runtimes/node.js
```diff
@@ -116,30 +117,47 @@ const zipFunction = async function ({
...
+  // If a specific JS bundler version is specified, we'll use it.
+  if (parameters.jsBundler) {
+    return zipFunction(parameters)
+  }
+
+  // Otherwise, we'll try to bundle with v2 and, if that fails, fallback to v1.
+  try {
+    return await zipFunction({ ...parameters, jsBundler: JS_BUNDLER_ESBUILD })
+  } catch (error) {
+    const data = await zipFunction({ ...parameters, jsBundler: JS_BUNDLER_LEGACY })
```
ehmicky, line 157: If both `esbuild` and the legacy bundler fail, this probably mean there is an issue that is not related to the choice of bundler. In that case, would it make sense to re-throw the esbuild `error` (as opposed to the exception thrown by the legacy bundler)?

tests/helpers/test_bundlers.js
```diff
@@ -0,0 +1,19 @@
+const makeTestBundlers = (test) => {
```
ehmicky, line 1: Nice!

ehmicky (approved):
This is looking great!

I added two comments about error handling, but those are mere non-blocking suggestions. :+1:

erezrokah (approved):
What @ehmicky said 🚀

src/runtimes/node.js
eduardoboucas: This ties back to the fact that `zip-it-and-ship-it` still technically supports a binary entry point with a JSON output, so any non-error states should be serialisable to JSON. In this case, we're returning errors from esbuild but the status of the application is not an error state — i.e. esbuild has failed, but we were able to succeed in zipping the functions thanks to the fallback.

esbuild returns an `errors` and `warnings` properties, containing information such as the location (file, line and column) and the description of the event, which I was thinking of surfacing in Netlify Build.

Does that make sense?

src/runtimes/node.js
eduardoboucas: I think that's a good idea. I've done it in https://github.com/netlify/zip-it-and-ship-it/pull/339/commits/ee896198d1a952e83283a8865aa770cf8e6ad8cb#diff-dda86982a0cad48036d5871c2d2f251ac96f1b30089e1e4879cd9903eec8cbcbR153-R164. Let me know if that's what you had in mind.

ehmicky: Yes, exactly! :+1:

src/runtimes/node.js
ehmicky: Yes, thanks for explaining :+1: 

Does `error.stack` contains that information as well? If so, it could also be used, and is JSON-serializable.

eduardoboucas:
I made a change in https://github.com/netlify/zip-it-and-ship-it/pull/339/commits/9201029a10f4ce7ec9d41e56265659267613a573 that changes the default bundler back to ZISI, and introduced a new value for `jsBundler` of `esbuild_zisi`, which contains the fallback behaviour.

The idea is to avoid a major release of `zip-it-and-ship-it` for now. Once we're happy with the esbuild implementation, we can just change the default value of `jsBundler` and cut a major release.

@erezrokah Does that make sense?

erezrokah:
> The idea is to avoid a major release of `zip-it-and-ship-it` for now. Once we're happy with the esbuild implementation, we can just change the default value of `jsBundler` and cut a major release.
> 
> @erezrokah Does that make sense?

Sounds good to me!

eduardoboucas:
@erezrokah I almost forgot your `devDependencies` point from https://github.com/netlify/zip-it-and-ship-it/pull/338#discussion_r578485874! 😱 

I added it in https://github.com/netlify/zip-it-and-ship-it/pull/339/commits/eef3c79d99c2b06632d70ee6750671742fec399f.

PR #347, opened 2021-02-26
fix: treat @prisma/client as an external module

eduardoboucas:
**- Summary**

This PR adds `@prisma/client` as a special case, treating it as an external module.

It also moves the evaluation of special cases a bit upstream, so that their external modules can be part of the traversal procedure.

**- Test plan**

N/A

fix: treat @prisma/client as an external module

download

### DIFF
--- a/tests/main.js
+++ b/tests/main.js
@@ -163,14 +163,12 @@ testBundlers('Throws on invalid package.json', [ESBUILD, ESBUILD_ZISI, DEFAULT],
   const invalidPackageJsonDir = `${fixtureDir}/invalid-package-json`
   const srcPackageJson = `${invalidPackageJsonDir}/package.json.txt`
   const distPackageJson = `${invalidPackageJsonDir}/package.json`
-  const expectedErrorRegex =
-    bundler === DEFAULT ? /invalid JSON/ : /package.json:1:1: error: Expected string but found "{"/
 
   await pRename(srcPackageJson, distPackageJson)
   try {
     await t.throwsAsync(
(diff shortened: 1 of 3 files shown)


tests/main.js
```diff
@@ -163,14 +163,12 @@ testBundlers('Throws on invalid package.json', [ESBUILD, ESBUILD_ZISI, DEFAULT],
...
   const distPackageJson = `${invalidPackageJsonDir}/package.json`
-  const expectedErrorRegex =
-    bundler === DEFAULT ? /invalid JSON/ : /package.json:1:1: error: Expected string but found "{"/
 
   await pRename(srcPackageJson, distPackageJson)
   try {
     await t.throwsAsync(
       zipNode(t, 'invalid-package-json', { opts: { jsBundler: bundler }, fixtureDir }),
-      expectedErrorRegex,
+      /(invalid JSON|package.json:1:1: error: Expected string but found "{")/,
```
ehmicky, line 171: Should we add a test case for Prisma?

ehmicky (approved):

tests/main.js
eduardoboucas: Sorry, I should've added some context about this change. It's not related to Prisma per se. It's about the fact that we might get a slightly different error message for when a `package.json` is invalid, depending on whether or not we have any external modules configured. This is varies with who validates the file: "us" or esbuild.

I haven't changed the Prisma test case because I can't think of a good way of mocking the `postinstall` scenario.

ehmicky: Sounds good :+1:

erezrokah (approved):

PR #353, opened 2021-03-01
fix: detect native bindings and change extension priority

eduardoboucas:
**- Summary**

This PR changes the order of `resolveExtensions` to prefer `.js` files over `.mjs`. This fixes an issue with the `node-fetch` module. More information at https://github.com/evanw/esbuild/issues/363#issuecomment-787978404.

Additionally, it removes an [optimisation for external modules](https://github.com/netlify/zip-it-and-ship-it/commit/bdcf286fb83ee55f955585d0b3a9f7fe85600565#diff-dda86982a0cad48036d5871c2d2f251ac96f1b30089e1e4879cd9903eec8cbcbL73-L80) which, as @erezrokah pointed out, can introduce problems with different versions of the same module. It feels dangerous to be risking problems that are difficult to troubleshoot due to what may well be a premature optimisation.

Finally, it adds an esbuild plugin that detects Node bindings (via a direct `require` call to a `.node` file) and adds the corresponding file to the zip, avoiding a fallback to ZISI.

**- Test plan**

Added a test specifically for `node-fetch` and amended the test for modules with native bindings.

fix: detect native bindings and change extension priority

hamster


src/node_bundler/index.js
```diff
@@ -35,6 +44,8 @@ const bundleJsFile = async function ({
     outfile: bundlePath,
     nodePaths: additionalModulePaths,
     platform: 'node',
+    plugins,
+    resolveExtensions: ['.js', '.jsx', '.mjs', '.cjs', '.json'],
```
ehmicky, line 48: Should `.es6` be added? This is an old extension that was used in the past when ES6 became a thing.

ehmicky (approved):

src/node_bundler/index.js
eduardoboucas: Sure! I'll make a note to do some testing and add that in a subsequent PR.

PR #364, opened 2021-03-08
feat: add support for TypeScript

### LINKED ISSUE #346: Identify naming conflicts ahead of bundling

The following 3 function paths will all map to `{name}.zip`:

- `{name}/{name}.js`
- `{name}/index.js`
- `{name}.js`

We don't detect this situation upfront and we still run the bundling mechanism for all three, except the last one will overwrite the other two.

Once we add support for TypeScript, we'll a `.ts` variation for each of the above.

We detect naming conflicts upfront and skip the bundling logic for functions that will get overwritten.

Checklist:

- [ ] Formalise the order of precedence for each of the paths, with accompanying automated tests
- [ ] Identify naming conflicts upfront and skip the bundler for overwritten functions
- [ ] (Optional) Show a warning when a conflict is found

eduardoboucas:
**- Summary**

This PR introduces some changes in the way the application finds and processes functions, in order to enable a smoother integration of TypeScript support.

The most noteworthy change is that we're now listing all the functions and pre-computing their final names, as well as any naming conflicts, ahead of time. Currently, we're calling `getSrcPaths` to find all files in `srcFolder` and subsequently calling `zipFunction` on each of these files, leaving it up to this function to decide whether a particular item is a function or not.

If the source folder contains multiple files that will yield a function with the same name, like `functions/test.js` and `functions/test` (Go binary), `zip-it-and-ship-it` will still process each of these functions. At the end, one of them will overwrite all the others. Since we're using concurrency, I believe the result of this might even not be deterministic, since it may depend on which of the functions finishes processing first, but I haven't confirmed this.

With this PR, the discovery and de-duping steps take place before any functions are processed. It works as following:

1. When processing a directory with functions, we call [`getFunctionsFromPaths`](https://github.com/netlify/zip-it-and-ship-it/pull/364/files#diff-6f9b6ee29baf15259eed475b0bd9b3397d238e2ae9ff6a71345c3b96a49683b5R48) with the list of paths from the directory
2. This will call the `findFunctionsInPaths` function that each of the runtimes exports. This way, each runtime is responsible for telling the caller which files it wants to take on and generate a function from.
3. The `findFunctionsInPaths` of the next runtime is called with all the paths except the ones that were taken by the runtimes that preceded it and so on, until the last runtime.

With this foundational work in place, we're also adding support for TypeScript functions. We give priority to JavaScript files (e.g. `functions/my-function.js` will take precedence over `functions/my-function.ts`) to try to reduce any compatibility problems. For example, if someone is currently compiling a TypeScript file into a JavaScript file with the same name, we should keep honouring this and process the generated JavaScript file.

Closes #346.

**- Test plan**

- Added tests that capture the order of precedence of various function paths
- Added tests with TypeScript functions

feat: add support for TypeScript

cute-smiling-animals-251


src/main.js
```diff
@@ -19,22 +20,22 @@ const listFunctionsFiles = async function (
...
   )
+
   // TODO: switch to Array.flat() once we drop support for Node.js < 11.0.0
   // eslint-disable-next-line unicorn/prefer-spread
   return [].concat(...listedFunctionsFiles)
 }
 
 const getListedFunction = function ({ runtime, name, mainFile, extension }) {
-  return { name, mainFile, runtime, extension }
+  return { name, mainFile, runtime: runtime.name, extension }
```
eduardoboucas, line 38: The `runtime` object that is passed around used to be a String containing the name of the runtime. It is now an object (or an "instance" of a runtime). We convert it to the string containing the name only at the termination points, like this one.

src/node_bundler/index.js
```diff
@@ -20,9 +20,6 @@ const bundleJsFile = async function ({
   const external = [...new Set([...externalModules, ...ignoredModules])]
   const jsFilename = `${basename(destFilename, extname(destFilename))}.js`
   const bundlePath = join(destFolder, jsFilename)
-  const pluginContext = {
```
eduardoboucas, line 23: We've abandoned the idea of using an esbuild plugin to handle `.node` files.

src/runtimes/detect_runtime.js
```diff
@@ -0,0 +1,24 @@
...
+const { promisify } = require('util')
+
+const { detect, Runtime } = require('elf-cam')
+
+const { RUNTIME_GO, RUNTIME_RUST } = require('../utils/consts')
+
+const pReadFile = promisify(readFile)
+
+// Try to guess the runtime by inspecting the binary file.
+const detectBinaryRuntime = async function (path) {
```
eduardoboucas, line 11: The logic for this is not new. It was simply moved from `src/runtime.js`.

src/runtimes/index.js
```diff
@@ -1,8 +1,82 @@
...
+ * Finds functions for a list of paths using a specific runtime. The return
+ * value is an object containing an array of the functions found (`functions`)
+ * and an array with the paths that haven't been recognized by the runtime
+ * (`remainingPaths`).
+ *
+ * @param   {Array<String>} paths
+ * @param   {Object} runtime
+ * @returns {Promise<Object>}
+ */
+const findFunctionsInRuntime = async function ({ dedupe = false, paths, runtime }) {
```
eduardoboucas, line 17: The `dedupe` parameter dictates whether `functions/test.js` and `functions/test/index.js` will yield two entries or merged into one (according to the precedence rules). I've added this because methods like `listFunctions` and `listFunctionFiles` are still returning all the functions and corresponding files regardless of naming conflicts simply to maintain backwards compatibility, but this feels like something we should revisit.

src/runtimes/index.js
```diff
@@ -1,8 +1,82 @@
...
+  const functions = await runtime.findFunctionsInPaths(paths)
+
+  // If `dedupe` is true, we use the function name (`filename`) as the map key,
+  // so that `function-1.js` will overwrite `function-1.go`. Otherwise, we use
+  // `srcPath`, so that both functions are returned.
+  const key = dedupe ? 'name' : 'srcPath'
+
+  // Augmenting the function objects with additional information.
+  const augmentedFunctions = functions.map((func) => [
```
eduardoboucas, line 26: These parameters could easily be returned by each runtime, but for convenience we compute them upstream so that runtimes don't have to return them, for convenience. They are generic parameters that can be inferred from the caller.

src/runtimes/index.js
```diff
@@ -1,8 +1,82 @@
...
+    func[key],
+    {
+      ...func,
+      extension: extname(func.mainFile),
+      filename: basename(func.srcPath),
+      runtime,
+    },
+  ])
+  const usedPaths = new Set(augmentedFunctions.map(([path]) => path))
+  const remainingPaths = paths.filter((path) => !usedPaths.has(path))
```
eduardoboucas, line 36: This will ensure that once a path has been "taken" by a runtime, it will not be sent to the next runtime, since it will add unnecessary computations.

src/utils/fs.js
```diff
@@ -0,0 +1,17 @@
+const { readdir } = require('fs')
+const { join } = require('path')
+const { promisify } = require('util')
+
+const pReaddir = promisify(readdir)
+
+const listFunctionsDirectory = async function (srcFolder) {
```
eduardoboucas, line 7: This previously existed as `listFilenames` in `src/info.js`.

src/zip.js
```diff
@@ -22,28 +26,41 @@ const zipFunctions = async function (
...
+  const paths = await listFunctionsDirectory(srcFolder)
+  const functions = await runtimes.getFunctionsFromPaths(paths, { dedupe: true })
+  const pluginsModulesPath = await getPluginsModulesPath(srcFolder)
   const zipped = await pMap(
-    srcPaths,
-    (srcPath) =>
-      zipFunction(srcPath, destFolder, {
+    functions.values(),
+    async (func) => {
+      const zipResult = await func.runtime.zipFunction({
```
eduardoboucas, line 41: Rather than calling the `zipFunction` function defined in this file, we're directly calling the `zipFunction` defined by each runtime. We can do this because during the discovery phase (i.e. lines 35-36) we've already collected all the information we need about the functions we're going to process.

The `zipFunction` function defined below remains as a top-level, entry point function for the application, so it includes the discovery step itself, in order to retrieve information about the function at the specific path. Doing this again from `zipFunctions` would be redundant.

src/zip_binary.js
```diff
@@ -0,0 +1,11 @@
+const { startZip, addZipFile, addZipContent, endZip } = require('./archive')
+
+// Zip a binary function file
+const zipBinary = async function ({ srcPath, destPath, filename, stat, runtime }) {
```
eduardoboucas, line 4: This function was moved from `src/runtime.js`.

tests/main.js
```diff
@@ -402,12 +402,12 @@ testBundlers(
     t.deepEqual(
       functions,
       [
-        { name: 'four', mainFile: 'four.js/four.js.js', runtime: 'js', extension: '.js' },
```
eduardoboucas, line 405: I've updated this test because even though the listed functions are the same, the order in which they're listed has changed, which shouldn't matter for the purposes of `listFunctions()`.

tests/main.js
```diff
@@ -422,6 +422,8 @@ testBundlers(
     t.deepEqual(
       functions,
       [
+        { name: 'test', mainFile: 'test.js', runtime: 'js', extension: '.js', srcFile: 'test.js' },
```
eduardoboucas, line 425: I've updated this test because even though the listed function files are the same, the order in which they're listed has changed, which shouldn't matter for the purposes of `listFunctionFiles()`.

src/main.js
```diff
@@ -19,22 +20,22 @@ const listFunctionsFiles = async function (
   srcFolder,
   { jsBundler = JS_BUNDLER_ZISI, jsExternalModules, jsIgnoredModules } = {},
 ) {
-  const [functionInfos, pluginsModulesPath] = await Promise.all([
-    getFunctionInfos(srcFolder),
-    getPluginsModulesPath(srcFolder),
-  ])
+  const paths = await listFunctionsDirectory(srcFolder)
+  const functions = await runtimes.getFunctionsFromPaths(paths)
+  const pluginsModulesPath = await getPluginsModulesPath(srcFolder)
```
ehmicky, line 25: Should `listFunctionsDirectory()` and `runtimes.getFunctionsFromPaths()` be performed in parallel to `getPluginsModulesPath()`? (for performance)

eduardoboucas: Good catch!

src/runtimes/index.js
```diff
@@ -1,8 +1,82 @@
...
+  const augmentedFunctions = functions.map((func) => [
+    func[key],
+    {
+      ...func,
+      extension: extname(func.mainFile),
+      filename: basename(func.srcPath),
+      runtime,
+    },
+  ])
+  const usedPaths = new Set(augmentedFunctions.map(([path]) => path))
```
ehmicky, line 35: Would it make sense to use the following line instead?

```js
  const usedPaths = new Set(functions.map((func) => func[key]))
```

This would allow `argumentedFunctions` to be directly an array of objects:

```js
const augmentedFunctions = functions.map((func) => ({
  ...func,
  extension: extname(func.mainFile),
  filename: basename(func.srcPath),
  runtime,
}))
```

This, in turn, would remove the need to use a `Map()` in `getFunctionsFromPaths()`, returning an array instead. This would remove the need for `map.values()` and iterator methods in consumers.

What do you think?

eduardoboucas: `augmentedFunctions` is returning an array where the first element is the function key and the second element is the function object. This syntax is compatible with [`Map`'s constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/Map#creating_a_new_map), so that by the time we've cycled through all run times and we construct a new `Map` at the end of `getFunctionsFromPaths`, we've de-duped any functions that share the same name.

If `augmentedFunctions` returns an array of objects, this de-duping would not be possible, unless I misunderstood your suggestion.

ehmicky: Yes, you're right, my suggestion would not work :+1:

src/runtimes/index.js
```diff
@@ -1,8 +1,82 @@
...
+
+      return {
+        functions: [...aggregateFunctions, ...runtimeFunctions],
+        remainingPaths: runtimePaths,
+      }
+    },
+    { functions: [], remainingPaths: paths },
+  )
+
+  return new Map(functions)
```
ehmicky, line 75: Every consumer of this function is iterating over it with `map.values()`.
Would it make sense to directly return `[...new Map(functions).values()]`?

src/main.js
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/364/commits/493e24550351a8124874c9462f2a4e2513815a0e. Thanks for the suggestion!

src/runtimes/index.js
eduardoboucas: I'm a bit torn on this, actually.

On the one hand, it would simplify the consumers.

On the other hand, we'd be "losing information" by converting it to an array prematurely — e.g. we lose the ability to query a specific function by name, if we encounter that use case. Also, converting the Map to an array comes with a performance price which we don't always have to pay. For example, [here](https://github.com/netlify/zip-it-and-ship-it/blob/feat/typescript-support/src/zip.js#L86) we can simply access the right element in the Map by making use of the iterator, without having to convert to an array.

I don't have any strong preferences though, so I'm happy to make the change if you do!

ehmicky: This makes sense, good point :+1:

src/main.js
```diff
@@ -1,16 +1,17 @@
 const { extname } = require('path')
 
 require('./utils/polyfills')
-const { getFunctionInfos } = require('./info')
 const { getPluginsModulesPath } = require('./node_dependencies')
 const runtimes = require('./runtimes')
```
ehmicky, line 5: [sand] Should the new `getFunctionsFromPaths()` be added to a separate file from the one importing each individual runtime type? Those look somewhat different. Also, this would also keeping `runtimes/index.js` as a "map" of languages/runtimes.

src/runtimes/node/finder.js
```diff
@@ -0,0 +1,68 @@
...
+const getMainFile = function (srcPath, filename, stat) {
+  if (stat.isDirectory()) {
+    return locatePath(
+      [
+        join(srcPath, `${filename}.js`),
+        join(srcPath, 'index.js'),
+        join(srcPath, `${filename}.ts`),
+        join(srcPath, 'index.ts'),
+      ],
+      { type: 'file' },
```
ehmicky, line 57: Note: `type: 'file'` is the default value.

src/runtimes/go.js
```diff
@@ -1,8 +1,38 @@
...
 
 const cpFile = require('cp-file')
 
-const { zipBinary } = require('../runtime')
+const { RUNTIME_GO } = require('../utils/consts')
+const { zipBinary } = require('../zip_binary')
+
+const { detectBinaryRuntime } = require('./detect_runtime')
+
+const findFunctionsInPaths = async function (paths) {
```
ehmicky, line 14: This function is pretty similar from the Rust one, should those share some common code?

src/zip.js
```diff
@@ -22,28 +26,43 @@ const zipFunctions = async function (
     jsExternalModules = [],
     jsIgnoredModules = [],
     parallelLimit = DEFAULT_PARALLEL_LIMIT,
-    skipGo,
+    skipGo = true,
     zipGo,
   } = {},
 ) {
-  const [srcPaths, pluginsModulesPath] = await Promise.all([getSrcPaths(srcFolder), getPluginsModulesPath(srcFolder)])
+  await makeDir(destFolder)
```
ehmicky, line 33: Should this be run in parallel to `listFunctionsDirectory()`? For performance.

src/zip.js
```diff
@@ -22,28 +26,43 @@ const zipFunctions = async function (
...
+        mainFile: func.mainFile,
         pluginsModulesPath,
-        skipGo,
-        zipGo,
-      }),
+        runtime: func.runtime,
+        srcDir: func.srcDir,
+        srcPath: func.srcPath,
+        stat: func.stat,
+        zipGo: zipGo === undefined ? !skipGo : zipGo,
```
ehmicky, line 56: Is the new default value for `skipGo` and this line needed (as opposed to forwarding `skipGo` and `zipGo` as is? Considering `zipFunction()` default values already assign those parameters.

tests/main.js
```diff
@@ -573,6 +572,60 @@ testBundlers(
   },
 )
 
+testBundlers(
```
ehmicky, line 575: :heart: the tests

ehmicky (approved):
This looks good to me (none of my comments are blocking) :+1:

src/runtimes/index.js
```diff
@@ -1,8 +1,82 @@
...
+}
+
 module.exports = {
+  getFunctionsFromPaths,
   go: goRuntime,
```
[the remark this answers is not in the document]
eduardoboucas: Good point! Turns out the individual runtimes don't need to be exported, because we're no longer accessing them directly outside of this file. We're always using `getFunctionsFromPaths` as a proxy, so I've kept it as the only export and we can re-evaluate whenever we need to export the individual runtimes too.

Done in https://github.com/netlify/zip-it-and-ship-it/pull/364/commits/00604ef8dc6ca80cb7959a4b74952180de9a10d0.

src/zip.js
eduardoboucas: I've added it because `zipFunctions` is no longer calling the `zipFunction` from this file as it used to. It's now directly calling the `zipFunction` method of the corresponding runtime, where the default for `zipGo` is not defined.

src/zip.js
eduardoboucas: Definitely! Done in https://github.com/netlify/zip-it-and-ship-it/pull/364/commits/c0fe24f5650b41742b2fae70d1a4c02eb8b4eb82.

src/runtimes/go.js
ehmicky: Sounds good :+1:

src/zip.js
ehmicky: Makes sense. :+1: 
We should eventually remove `skipGo` and/or `zipGo`, because the default value is starting to be confusing to guess.

ehmicky (approved):

src/zip.js
eduardoboucas: Agree!

PR #373, opened 2021-03-11
chore!: remove `zipGo`/`skipGo` options

### LINKED ISSUE #372: Remove `skipGo`/`zipGo` parameters
The `skipGo`/`zipGo` parameter does not appear to be used by any user (including us), excepting for the CLI, which uses the default value.

https://github.com/netlify/cli/blob/3d8f1793ed90099152246fd4f69d44864762b55f/src/commands/functions/build.js#L56

We should remove those parameters. This would be a breaking change, i.e. require a major release.

ehmicky:
Fixes #372.

The `skipGo`/`zipGo` parameter does not appear to be used by any user (including us), excepting for the CLI, which uses the default value.

https://github.com/netlify/cli/blob/3d8f1793ed90099152246fd4f69d44864762b55f/src/commands/functions/build.js#L56

We should remove those parameters. This would be a breaking change, i.e. require a major release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ Creates Zip `archives` from Node.js, Go, and Rust programs. Those `archives` are
 - Sub-directories with a main file called either `index.js` or `{dir}.js` where `{dir}` is the sub-directory name.
 - `.js` files (Node.js)
 - `.zip` archives with Node.js already ready to upload to AWS Lambda.
-- Go programs already compiled. If the `options.zipGo` is `true`, those are zipped. Otherwise, those are copied as is.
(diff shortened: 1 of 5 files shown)


erezrokah (approved):

eduardoboucas (approved):

ehmicky:
We should hold on merging and releasing this until we decided to make a major release. What do you think @eduardoboucas?

eduardoboucas:
> We should hold on merging and releasing this until we decided to make a major release. What do you think @eduardoboucas?

Yes, I think that's probably best!

PR #374, opened 2021-03-12
fix: make JS files take precedence over TS files with the same name

eduardoboucas:
**- Summary**

When we started introducing support for TS files, the goal was for them to take precedence over JS functions with the same name. This was the case when the functions were in a sub-directory (e.g. `functions/sub-dir/function.js` vs. `functions/sub-dir/function.ts`). Unfortunately, this wasn't the case when the functions were at the root level (e.g. `functions/function.js` vs `functions/function.ts`). This PR addresses that.

**- Test plan**

- Added test to assert that `.js` files take precedence over `.ts` files
- Added test to assert that  `.js` files take precedence over `.zip` files (I verified this was the current behaviour) before we introduced these changes

fix: make JS files take precedence over TS files with the same name

images


tests/main.js
```diff
@@ -402,11 +402,11 @@ testBundlers(
     t.deepEqual(
       functions,
       [
-        { name: 'test', mainFile: 'test.js', runtime: 'js', extension: '.js' },
```
eduardoboucas, line 405: This changes the order of these files once again. 😞

tests/main.js
```diff
@@ -402,8 +402,8 @@ testBundlers(
...
         { name: 'test', mainFile: 'test.zip', runtime: 'js', extension: '.zip' },
+        { name: 'test', mainFile: 'test.js', runtime: 'js', extension: '.js' },
```
eduardoboucas, line 406: The reason for this changing is that we're now prioritising `.js` files over `.zip` files, which was the existing behaviour before #364.

src/runtimes/node/finder.js
```diff
@@ -6,19 +6,42 @@ const locatePath = require('locate-path')
...
   // eslint-disable-next-line fp/no-mutating-methods
-  const sortedFunctions = functions
-    .filter(Boolean)
-    .sort((fA, fB) => Number(fA.stat.isDirectory()) - Number(fB.stat.isDirectory()))
+  const sortedFunctions = functions.filter(Boolean).sort((fA, fB) => {
+    // We first sort the functions array to put directories first. This is so
+    // that `{name}/{name}.js` takes precedence over `{name}.js`.
+    const directorySort = Number(fA.stat.isDirectory()) - Number(fB.stat.isDirectory())
+
+    if (directorySort) {
```
ehmicky, line 33:
```suggestion
    if (directorySort !== 0) {
```

Should we explicit about the typecasting?

ehmicky (approved):

src/runtimes/node/finder.js
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/374/commits/4e6e530a5e1d417baf4a566e52f89d87899b5c05.

eduardoboucas: (Oh, just realised you made an actual suggestion which I could've committed directly!)

PR #376, opened 2021-03-12
feat!: add config property

eduardoboucas:
**- Summary**

This PR introduces a `config` property that can be passed to both `zipFunction` and `zipFunctions`. This is an object where keys are glob-like expressions and values are configuration objects, which will be passed to each function whose name matches the glob.

This property will enable the functions configuration API, and it will be supplied by `@netlify/build`.

Currently, the following properties are pulled from this object:

- `jsExternalModules`
- `jsIgnoredModules`

Expressions are sorted based on the wildcards that they use. For example, a function named `netlify` is matched by `net*`, `netli*` and `netlify`, but they have different levels of precedence, according to how close the expression is to the actual function name. An exact match is the highest level of precedence.

When multiple configuration objects match a function, they will be deep-merged (and arrays will be concatenated). For example, take the following `config` object:

```json
{
  "net*": {
    "imaginaryProperty": 1,
    "jsExternalModules": ["one"]
  },
  "netli*": {
    "imaginaryProperty": 2,
    "jsExternalModules": ["two"]
  },
  "netlify": {
    "imaginaryProperty": 3,
    "jsExternalModules": ["three"]
  }
}
```

When applied to a function named `netlify`, the resulting configuration object will be:

```json
{
  "imaginaryProperty": 3,
  "jsExternalModules": ["one", "two", "three"]
}
```

**- Test plan**

Tests added.

feat: add config property

ss-160826-twip-05_8cf6d4cb83758449fd400c7c3d71aa1f fit-760w


package.json
```diff
@@ -63,6 +63,8 @@
     "junk": "^3.1.0",
     "locate-path": "^5.0.0",
     "make-dir": "^3.1.0",
+    "merge-options": "^3.0.4",
+    "minimatch": "^3.0.4",
```
ehmicky, line 67: `minimatch` is unmaintained.
Should we use a different library, e.g. https://github.com/micromatch/micromatch / https://github.com/mrmlnc/fast-glob ?

src/zip.js
```diff
@@ -85,16 +80,15 @@ const zipFunction = async function (
...
 
-  const { extension, filename, mainFile, runtime, srcDir, stat } = functions.values().next().value
+  const { config, extension, filename, mainFile, runtime, srcDir, stat } = functions.values().next().value
   const pluginsModulesPath =
     defaultModulesPath === undefined ? await getPluginsModulesPath(srcPath) : defaultModulesPath
 
   await makeDir(destFolder)
 
   const zipResult = await runtime.zipFunction({
+    config,
```
ehmicky, line 90: We should make a major release for this PR since it is technically a breaking change.
Should we release #373 at the same time?

package.json
eduardoboucas: I chose `minimatch` because `micromatch` is like three times larger. `fast-glob` uses `micromatch` itself. But you're right, the state of `minimatch` doesn't look good.

src/config.js
```diff
@@ -0,0 +1,33 @@
...
+  const matches = Object.keys(config)
+    .filter((expression) => minimatch(func.name, expression))
+    .map((expression) => {
+      const wildcardCount = [...expression].filter((char) => char === '*').length
+
+      // The weight increases with the number of hardcoded (i.e. non-wildcard)
+      // characters — e.g. "netlify" has a higher weight than "net*". We do a
+      // subtraction of 1 if there is at least one wildcard character, so that
+      // "netlify" has a higher weight than "netlify*".
+      const weight = expression.length - wildcardCount - (wildcardCount ? 1 : 0)
```
ehmicky, line 20:
```suggestion
      const weight = expression.length - wildcardCount - (wildcardCount !== 0 ? 1 : 0)
```

src/config.js
```diff
@@ -0,0 +1,33 @@
...
+
+      return {
+        expression,
+        weight,
+      }
+    })
+    .sort((fA, fB) => fA.weight - fB.weight)
```
ehmicky, line 27: When two functions have the same `weight`, should we sort them according to a second criteria?

src/config.js
```diff
@@ -0,0 +1,33 @@
...
+    .map(({ expression }) => config[expression])
+
+  return mergeOptions.apply({ concatArrays: true }, matches)
```
ehmicky, line 30: Will the `config` option include nested objects? If not, would a shallow merge work `Object.assign({}, ...matches)`?

Also, do we want to concatenate arrays instead of overriding? I think some users might expect simple overrides. Also, this removes the possibility for one globbing pattern to "undo" an array property from another globbing pattern.

src/config.js
eduardoboucas: When they have the same weight, we return `0` in the sorting function, which leaves the elements unchanged. Do you feel that is not sufficient?

src/zip.js
eduardoboucas: Yes, good call!

package.json
ehmicky: We can maybe choose another library than both of those.
But in general, I think the fact it has been mostly unmaintained for 4 years might become a problem: bugs and security fixes are not being applied, new Node.js major releases might break, etc.

src/config.js
ehmicky: Is the original sorting order based on the return value from `fs.readdir()`?

Since the sorting order is significant due to the merging, I am wondering whether we should have a more explicit/defined second sorting order?

src/config.js
eduardoboucas: > Will the config option include nested objects?

I believe it will, yes. We've discussed the ability to define an `environment` object in the configuration objects, which would need to be deep-merged.

> Also, do we want to concatenate arrays instead of overriding? I think some users might expect simple overrides. Also, this removes the possibility for one globbing pattern to "undo" an array property from another globbing pattern.

In the case of external/ignored modules specifically, I think it's important that we concatenate. Imagine that the Next.js plugin specifies an array of external modules for all the plugin-generated functions. If you then add your own modules in the config, you would be unknowingly overriding what the plugin has defined and potentially breaking the build.

In general, I'd say that if someone defined a module as external, there's a specific reason behind it and we should be careful about overriding that behaviour. If we do think it's important that an expression is able to unset an external module defined by a less-specific glob, I would rather use the `!` notation (e.g. `externalModules: ["one", "!two"]`).

What do you think?

package.json
eduardoboucas: You're right. I'll make the change.

ehmicky: Also: we use `minimatch` in `@netlify/cache-utils`, `netlify-cli` (indirectly) and `@netlify/git-utils` (directly), so this would not change a difference in npm package size when `@netlify/zip-it-and-ship-it` is included in one of those. 
(Also this might provide with consistency)

src/config.js
eduardoboucas: > Is the original sorting order based on the return value from fs.readdir()?

The original sorting order is defined by the order of the keys of the configuration object.

(We're comparing expression objects, not functions,  which made me realise that the `fA` and `fB` variable names are misleading. I'll update that.)

As for a second sorting order, what do you have in mind?

package.json
ehmicky: (Note: I meant `micromatch` above, editing my comment)

`minimatch` works :+1:

src/config.js
ehmicky: The order of the keys of the configuration object sounds like a good "second" sorting order, what do you think? This is probably what the user would expect.

If this is the current behavior, then we're good :+1:

src/config.js
ehmicky: Yes, this makes sense (on both points). :+1:

I am assuming `!` is already handled thanks to the globbing library, is that correct?

src/zip.js
ehmicky: I updated the title of the PR to add `!` for conventional commits.

src/config.js
eduardoboucas: Agreed! 👍🏻

src/config.js
eduardoboucas: > I am assuming ! is already handled thanks to the globbing library, is that correct?

Not yet. The globbing libary will operate on the function expressions alone, not on the values of the configuration properties. If we want to introduce support for the `!` syntax, we'll need to handle that ourselves.

ehmicky: Should we create an issue to follow-up on this?

ehmicky (approved):

src/config.js
eduardoboucas: Yes! https://github.com/netlify/zip-it-and-ship-it/issues/381

README.md
```diff
@@ -88,30 +88,34 @@ Maximum number of Functions to bundle at the same time.
...
+
+An object matching glob-like expressions to objects containing configuration properties. Whenever a function name
+matches one of the expressions, it inherits the configuration properties.
 
-_Type_: `array<string>`\
-_Default value_: `[]`
+The following properties are accepted:
 
-List of Node modules to include separately inside a node_modules directory.
+- `jsExternalModules`
```
ehmicky, line 108: Should we consider renaming those properties (in a separate PR) to match how they are called in `@netlify/config`?

eduardoboucas: > Should we consider renaming those properties (in a separate PR) to match how they are called in `@netlify/config`?

You mean the camel case vs. snake case difference? In https://github.com/netlify/build/pull/2414 I had Build doing the translation. Do you think ZISI should just handle the snake case properties?

ehmicky: No, it probably should keep using camelCase.
I meant, the new names suggested here: https://github.com/netlify/build/pull/2390#issuecomment-799535006

eduardoboucas: > No, it probably should keep using camelCase.
> 
> I meant, the new names suggested here: https://github.com/netlify/build/pull/2390#issuecomment-799535006

Ah, yes! Definitely. I was just waiting for a 👍/👎 from @verythorough before making the change.

eduardoboucas:
@ehmicky FYI I've renamed the properties in https://github.com/netlify/zip-it-and-ship-it/pull/376/commits/5e361419896892546f3b50b3bba5764a36b7a7ba.

src/config.js
erezrokah: > In general, I'd say that if someone defined a module as external, there's a specific reason behind it and we should be careful about overriding that behaviour. If we do think it's important that an expression is able to unset an external module defined by a less-specific glob, I would rather use the `!` notation (e.g. `externalModules: ["one", "!two"]`).

I think the negate approach is a bit problematic since users won't have any idea what external modules will be added by plugins, so they won't know which ones to negate.

How about a `force` syntax like we have with `redirects`?
This can work at a single property level (e.g. to avoid array concatenation) and at a glob pattern level (to avoid a merge).

erezrokah (approved):
This looks great.
I have one small comment about merging/concatenation of properties/arrays. I don't think it's a blocker for this PR, but I think we should give it some thought.

eduardoboucas:
Thanks @erezrokah!

> I think the negate approach is a bit problematic since users won't have any idea what external modules will be added by plugins, so they won't know which ones to negate.
>
> How about a force syntax like we have with redirects?
> This can work at a single property level (e.g. to avoid array concatenation) and at a glob pattern level (to avoid a merge).

That makes sense. Would you mind adding that feedback in https://github.com/netlify/zip-it-and-ship-it/issues/381?

Issue #384, opened 2021-03-16
Flag bundling errors as user errors


eduardoboucas:
**- Do you want to request a _feature_ or report a _bug_?**

🐞 

**- What is the current behavior?**

Errors resulting from bundling a function with esbuild are flagged as unhandled errors.

**- What is the expected behavior?**

These errors should be flagged as user errors.

**- Please mention your node.js, and operating system version.**

N/A

ehmicky:
Progress at https://github.com/netlify/build/pull/2419 and https://github.com/netlify/build/pull/2420

PR #404, opened 2021-04-05
feat: add errorInfo property to esbuild bundling errors

eduardoboucas:
**- Summary**

This PR adds an `errorInfo` property to any errors resulting from bundling with esbuild, so that they're flagged as user errors in our build system.

Closes https://github.com/netlify/zip-it-and-ship-it/issues/384.

**- Test plan**

Added a test.

feat: add errorInfo property to esbuild bundling errors

p091prnq


src/runtimes/node/bundler.js
```diff
@@ -46,7 +35,25 @@ const bundleJsFile = async function ({
...
+      outfile: bundlePath,
+      nodePaths: additionalModulePaths,
+      platform: 'node',
+      resolveExtensions: ['.js', '.jsx', '.mjs', '.cjs', '.json'],
+      target: ['es2017'],
+    })
+
+    return { bundlePath, cleanTempFiles, data }
+  } catch (error) {
+    error.errorInfo = { type: 'functionsBundling' }
```
ehmicky, line 53: At the moment, the `functionsBundling` error type has a required field `location.functionName` which is used in the error message.

https://github.com/netlify/build/blob/fb47c2059d7ad46c48df5d715de759de38c0712d/packages/build/src/error/type.js#L86

I.e. this would need to be `error.errorInfo = { type: 'functionsBundling', location: { functionName } }`.

ehmicky, line 53: The field has been renamed to `customErrorInfo`.

https://github.com/netlify/build/blob/fb47c2059d7ad46c48df5d715de759de38c0712d/packages/build/src/error/info.js#L40

eduardoboucas: Ah, of course! 🤦🏻‍♂️  Fixed in https://github.com/netlify/zip-it-and-ship-it/pull/404/commits/920ee876b5874a8f4348e4ba15d9ff5451fcce8c.

eduardoboucas: Ah, I didn't know this. Thanks for flagging! Done in https://github.com/netlify/zip-it-and-ship-it/pull/404/commits/920ee876b5874a8f4348e4ba15d9ff5451fcce8c.

ehmicky (approved):

PR #410, opened 2021-04-06
feat: add archiveFormat configuration property

### LINKED ISSUE #409: Introduce archiveFormat configuration property

Feature. 🚀 

The application creates a ZIP archive for each bundled function.

Introducing a new configuration property that allows users to configure the output format.

- Proposed name: `archiveFormat`
- Possible values: `zip`, `none`
- Default value: `zip`

If the value is `none`, each function generates a directory containing all the bundled files.

This option would be used by the CLI for serving bundled functions without the need to zip/unzip files.

eduardoboucas:
**- Summary**

This PR introduces a new `archiveFormat` configuration property.

Closes #409.

**- Test plan**

Added 2 new tests.

feat: add archiveFormat configuration property

b3382984d7818ed5b99ed899e98c9913


README.md
```diff
@@ -84,20 +93,19 @@ The following properties are accepted:
...
-  _Type_: `array<string>`\
+  _Type_: `array<string>`
 
   List of Node modules to keep out of the bundle.
 
 - `nodeBundler`
 
-  _Type_: `string`\
-  _Default value_: `zisi`
+  _Type_: `string` _Default value_: `zisi`
```
ehmicky, line 108: I believe this might have been accidental.

src/runtimes/node/index.js
```diff
@@ -53,6 +55,7 @@ const zipFunction = async function ({
   }
 
   return zipEsbuild({
+    archiveFormat,
     config,
     destFolder,
     destPath,
```
ehmicky, line 61: This line can be removed.

src/zip_node.js
```diff
@@ -1,16 +1,73 @@
...
+  const functionFolder = join(destFolder, basename(filename, extension))
+
+  // Creating the function folder.
+  await makeDir(functionFolder)
+
+  // Writing entry file.
+  await pWriteFile(join(functionFolder, entryFilename), entryContents)
+
+  // Copying source files.
+  await pMap(
```
ehmicky, line 45: Would [`p-limit`](https://github.com/sindresorhus/p-limit) be better here, since we don't use the return value of `p-map`?

src/zip_node.js
```diff
@@ -1,16 +1,73 @@
...
 const { startZip, addZipFile, addZipContent, endZip } = require('./archive')
+const { ARCHIVE_FORMAT_ZIP } = require('./utils/consts')
 
 const pStat = promisify(fs.stat)
+const pWriteFile = promisify(fs.writeFile)
 
-// Zip a Node.js function file
-const zipNodeJs = async function ({ basePath, destPath, filename, mainFile, pluginsModulesPath, aliases, srcFiles }) {
+// Taken from https://www.npmjs.com/package/cpy.
+const COPY_FILE_CONCURRENCY = os.cpus().length === 0 ? 2 : os.cpus().length * 2
```
ehmicky, line 19: Could `cpy` be used directly?
One potential issue is the globbing, maybe it can be disabled?

ehmicky (approved):

README.md
eduardoboucas: It was! Thanks. Fixed in https://github.com/netlify/zip-it-and-ship-it/pull/410/commits/b29fd723a3db7b82cbeb128c1ea70714a441042c.

src/runtimes/node/index.js
eduardoboucas: Absolutely. The path is now computed in the individual zip functions. Fixed in https://github.com/netlify/zip-it-and-ship-it/pull/410/commits/b29fd723a3db7b82cbeb128c1ea70714a441042c.

src/zip_node.js
eduardoboucas: I tried using `cpy` initially, but I wasn't able to correctly preserve the hierarchy of files in the target directory.
The library offers the [`parents` option](https://github.com/sindresorhus/cpy#parents), but it doesn't seem to do exactly what we need, i.e. it doesn't seem to compute the relative path like we're doing.

ehmicky: I've had the same problem pretty often when using that library. The `parents` option is not intuitive.
Usually, I manage to copy recursively using the following (odd) syntax:

```js
await cpy('**', destDir, { cwd: srcDir, parents: true })
```

eduardoboucas: Yes, I've considered that. We're using that approach somewhere in the ZISI tests.

However, I don't think it'll work here because we don't want to copy all files under a certain directory. We only want to copy the files contained in the array.

Also, because of the `aliases` object, we need the ability to modify individual paths before copying them, which we wouldn't be able to do using `cpy`.

Finally, `cpy` is currently under `devDependencies`, and by moving it to `dependencies` we'd be adding to the package size. I would rather avoid that, if possible.

ehmicky: This makes sense, thanks for explaining :+1:

tests/main.js
```diff
@@ -406,10 +410,26 @@ testBundlers('Can reduce parallelism', [ESBUILD, ESBUILD_ZISI, DEFAULT], async (
...
+        return ESBUILD
+
+      case DEFAULT:
+        return JS_BUNDLER_ZISI
+
+      default:
+        return bundler
+    }
+  }
+  const outBundler = getOutBundler()
```
ehmicky, line 428: Would an object work in this case?

```js
const outBundlers = { [ESBUILD_ZISI]: ESBUILD, [DEFAULT]: JS_BUNDLER_ZISI }
const outBundler = outBundlers[bundler] || bundler
```

eduardoboucas: Sure! Done in https://github.com/netlify/zip-it-and-ship-it/pull/410/commits/2826d4bb1632130f7e458c8857346fd0882b4321.

Issue #399, opened 2021-04-01
Add benchmarks to CI


eduardoboucas:
We should performance benchmarks as part of the CI flow, giving us visibility of whether a PR is about to introduce any performance degradation.

eduardoboucas:
Done in https://github.com/netlify/zip-it-and-ship-it/pull/411.

PR #411, opened 2021-04-06
chore: add benchmark

eduardoboucas:
**- Summary**

First attempt at https://github.com/netlify/zip-it-and-ship-it/issues/399.

chore: add benchmark

snails-af1f9bf


benchmarks/run.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+(cd benchmarks/fixtures && npm ci);
```
ehmicky, line 2: Would `npm ci --prefix benchmarks/fixtures` work?

benchmarks/run.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
eduardoboucas, line 1: I've opted for a bash script just to avoid polluting `package.json` with dependencies that are only used in the benchmark fixtures, but I'm open to alternative ideas.

ehmicky (approved):

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/setup-node@v2.1.5
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run benchmarks
+        run: npm run benchmark
+
+      - name: Delta-T
+        uses: eduardoboucas/delta-t-action@main
```
eduardoboucas, line 26: We can transfer this to the `netlify` org if/when we're happy with this approach.

benchmarks/run.sh
eduardoboucas: Absolutely, thanks! Done in https://github.com/netlify/zip-it-and-ship-it/pull/411/commits/91b08cf7dd9b50186d4c410a99bfdf6fc2cec6e9.

PR #418, opened 2021-04-13
feat: add plugin to detect native modules

eduardoboucas:
**- Summary**

This PR adds an esbuild plugin to detect modules with native bindings and automatically mark them as external.

**- Test plan**

Added new tests.

feat: add plugin to detect native modules

unnamed


src/runtimes/node/native_modules/detector.js
```diff
@@ -0,0 +1,16 @@
+// eslint-disable-next-line complexity
+const isNativeModule = ({ binary, dependencies = {}, devDependencies = {}, gypfile }) =>
```
eduardoboucas, line 2: I've opted for inlining this check over using [`is-native-module`](https://www.npmjs.com/package/is-native-module) because we also want to look for certain packages in `devDependencies`.

src/runtimes/node/native_modules/plugin.js
```diff
@@ -0,0 +1,68 @@
...
+  return cache[packageJsonPath]
+}
+
+const externalNativeModulesPlugin = (externalizedModules) => ({
+  name: 'external-native-modules',
+  setup(build) {
+    const cache = {}
+
+    // eslint-disable-next-line complexity, max-statements
+    build.onResolve({ filter: /^([^./]*)$/ }, async (args) => {
```
erezrokah, line 27: [sand] can we extract the regex to a const with a meaningful name so it is clear what we're filtering?

src/runtimes/node/native_modules/plugin.js
```diff
@@ -0,0 +1,68 @@
...
+      const package = packageName.exec(args.path)
+
+      if (!package) return
+
+      // eslint-disable-next-line fp/no-let
+      let directory = args.resolveDir
+
+      // eslint-disable-next-line fp/no-loops
+      while (true) {
```
erezrokah, line 36: [sand] Should we put the stop condition (`parentDirectory !== directory`) here?

erezrokah (changes requested):
@eduardoboucas nice work.

Some comments, nothing major

src/runtimes/node/native_modules/plugin.js
ehmicky, line 27: Would merging the `packageName` regExp to `filter` improve performance?

src/runtimes/node/native_modules/detector.js
eduardoboucas: > Can we merge dependencies and devDependencies and pass the merged object to is-native-module?

Since this is such a hot function, I'd prefer to avoid the cost of creating a new object each time, if we can help it. The impact should be really minimal, but we're trying to heavily optimise this function.

> We could also submit a PR to is-native-module to check for dev dependencies too?

Sounds good! I'll do that.

src/runtimes/node/native_modules/plugin.js
eduardoboucas: That means having `parentDirectory` outside the scope of the loop and initialising them both together, which feels a bit odd? What do you think?

src/runtimes/node/native_modules/plugin.js
eduardoboucas: Absolutely!

eduardoboucas: From my tests, no. I think this is because the regex is more complex, which gets more expensive since it'll be executed for each import. Also, the regex we're currently using for filtering already filters out relative and absolute paths, so the number of callback invokations that we'd reduce by using `packageName` would be minimal.

eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/418/commits/f47f22206c4e80d9e921b7b1e2af96d8dc95d379.

eduardoboucas: And we'd still need to run `packageName.exec` inside the callback, because we need to extract the directory from the package name.

src/runtimes/node/native_modules/plugin.js
erezrokah: Good point. Since `directory` is already in the outer scope we could be consistent with `parentDirectory` too.
I'm good either way.

src/runtimes/node/native_modules/detector.js
erezrokah: I'm good with inlining to unblock and using the new `is-native-module` once the PR lands

src/runtimes/node/native_modules/plugin.js
ehmicky: :+1:

ehmicky (approved):

src/runtimes/node/native_modules/plugin.js
eduardoboucas: It's just that placing `parentDirectory` in the outer scope means duplicating the `path.dirname` assignment: once outside, to satisfy the loop condition, and a second time inside to calculate the parent directory of each iteration, unless I'm missing what you're suggesting.

erezrokah: Yes, you're right. We can keep it as is 👍

erezrokah (approved):

PR #425, opened 2021-04-16
feat: add nodeVersion config property

### LINKED ISSUE #400: Adjust esbuild target to AWS_LAMBDA_JS_RUNTIME

Feature.

Currently, esbuild is configured to use `es2017` as the target. We should make it this configurable, and we should take the `AWS_LAMBDA_JS_RUNTIME` environment variable into account for this.

eduardoboucas:
**- Summary**

Introduces a new configuration property called `nodeVersion`, which specifies the version of Node.js that esbuild will use as the compilation target.

It supports values like `10.x`, as well as the `nodejs10.x` format used by [AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html).

The idea is for consumers of `zip-it-and-ship-it` to supply this configuration property based on the environment it's running on. For example, Netlify Build could pass along the value of the `AWS_LAMBDA_JS_RUNTIME` environment variable, whereas Netlify CLI could supply the local version of Node.js when serving bundled functions in `netlify dev`.

Closes #400.

**- Test plan**

Added one new test.

feat: add nodeVersion config property


README.md
```diff
@@ -112,6 +112,18 @@ The following properties are accepted:
...
+
+  _Type_: `string`\
+  _Default value_: `12.x`
+
+  The version of Node.js to use as the compilation target. Possible values:
+
+  - `8.x` (or `nodejs8.x`)
+  - `10.x` (or `nodejs10.x`)
+  - `12.x` (or `nodejs12.x`)
+  - `14.x` (or `nodejs14.x`)
```
ehmicky, line 125: Would it be simpler to allow only the AWS values, i.e. `nodejs8.x` but not `8.x`?

tests/fixtures/node-module-optional-catch-binding/function.js
```diff
@@ -0,0 +1,12 @@
+const noop = () => {}
+
+module.exports.handler = () => {
+  try {
+    noop()
+    // eslint-disable-next-line node/no-unsupported-features/es-syntax
```
ehmicky, line 6: Would using `} catch (error) {` remove the linting error?

ehmicky (approved):
:tada:

README.md
eduardoboucas: It's simpler, but from an API perspective it feels redundant to include `nodejs` in the value of a property that relates specifically to the Node.js version (e.g. `nodeVersion: "nodejs10.x"`), so I thought it'd be nice to allow the shorter syntax too.

It's not essential though, so I'm happy to revisit if you think this is a bad idea.

ehmicky: :+1:

PR #426, opened 2021-04-16
feat: add config property to CLI entry point

eduardoboucas:
**- Summary**

The `config` property was only available via the programmatic entry point. This PR adds it to the CLI interface.

Usage example:

```sh
zip-it-and-ship-it --config.\*.nodeBundler=esbuild functions/ functions-out/
```

**- Test plan**

N/A

feat: add config property to CLI entry point

DZW-S51W0AA2vQV


tests/main.js
```diff
@@ -477,25 +477,21 @@ testBundlers('Can use zipFunction()', [ESBUILD, ESBUILD_ZISI, DEFAULT], async (b
...
-        { name: 'test', mainFile: 'test.js', runtime: 'js', extension: '.js' },
-        { name: 'four', mainFile: 'four.js/four.js.js', runtime: 'js', extension: '.js' },
-        { name: 'one', mainFile: 'one/index.js', runtime: 'js', extension: '.js' },
-        { name: 'two', mainFile: 'two/two.js', runtime: 'js', extension: '.js' },
-        { name: 'test', mainFile: 'test', runtime: 'go', extension: '' },
-      ].map(normalizeFiles.bind(null, fixtureDir)),
-    )
-  },
-)
+test('Can list function main files with listFunctions()', async (t) => {
```
eduardoboucas, line 480: Unrelated to the feature itself, but I realised that this test was being ran once for each bundler despite the fact that the bundler doesn't influence the output.

src/bin.js
```diff
@@ -30,6 +30,12 @@ const OPTIONS = {
     default: ARCHIVE_FORMAT_ZIP,
     describe: 'Format of the archive created for each function',
   },
+  config: {
+    object: true,
```
ehmicky, line 34: I think `object` is not a known property with Yargs (although it might ignore unknown properties).
Yargs automatically parses dot notation CLI flags to objects (unless the `dot-notation` parser property is used).

ehmicky (approved):

src/bin.js
eduardoboucas: Thanks! Removed in https://github.com/netlify/zip-it-and-ship-it/pull/426/commits/966e8f25fac8248f6e4d56a9f0d226cdced78672.

PR #433, opened 2021-04-20
feat: add `nativeNodeModules` property to output

eduardoboucas:
**- Summary**

This PR adds a `nativeNodeModules` property to the return value of `zipFunction` and `zipFunctions` with any native modules detected as part of the logic added in #418. This object has the following shape:

```js
{
  "nativeNodeModules": {
    "module name": {
      "/full/path/to/the/module": "version (e.g. 1.0.0)"
    }
  }
}
```

With this property, we'll be able to show warnings in the CLI whenever the user is about to locally build and deploy a function with native modules.

Note that this currently only works when building with esbuild. Given that we're making it the new default in less than a month, I'm not sure the effort of adding this detection to the legacy bundler is worth the trouble, but I'm happy to discuss this.

**- Test plan**

Adjusted existing tests.

giant_panda_mei_xiang_enjoys_her_afternoon_nap_at_the_national_zoo_in_washington_on_aug_23_2007 _reuters


ehmicky (approved):
:+1: 

Note: CI tests are failing on Node 8

eduardoboucas:
> 👍
> 
> Note: CI tests are failing on Node 8

Working on that. ⏳

PR #439, opened 2021-04-23
feat: add inputs property

eduardoboucas:
**- Summary**

Adds an `inputs` property to the output of `zipFunction` and `zipFunctions`. This contains the (absolute) paths that contributed to the bundle generation, which means the paths that were visited as part of the tree-shaking process.

For the legacy bundler, this value is the same as the internal `srcFiles`, since node modules aren't subject to tree-shaking.

This will allow us to have a much more performant file watching mechanism in CLI, since we'll be able to determine exactly which functions were affected by a given file change and thus avoiding the need to re-bundle all functions.

**- Test plan**

Added a new test.

2027805_052317-kgo-baby-rhino-img_Image_00-00-12,01


src/runtimes/node/bundler.js
```diff
@@ -50,15 +50,17 @@ const bundleJsFile = async function ({
...
-      outfile: bundlePath,
+      metafile: true,
       nodePaths: additionalModulePaths,
+      outfile: bundlePath,
       platform: 'node',
       plugins: supportsAsyncAPI ? plugins : [],
       resolveExtensions: ['.js', '.jsx', '.mjs', '.cjs', '.ts', '.json'],
       target: [nodeTarget],
     })
+    const inputs = Object.keys(data.metafile.inputs).map((path) => resolve(process.cwd(), path))
```
ehmicky, line 61: [sand]

```suggestion
    const inputs = Object.keys(data.metafile.inputs).map((path) => resolve(path))
```

ehmicky (approved):

src/runtimes/node/bundler.js
eduardoboucas: Thanks!

PR #441, opened 2021-04-23
feat: add support for included_files config property

eduardoboucas:
**- Summary**

Adds support for an `includedFiles` property supplied as part of the `config` object. Any paths that match one of the globs found in `included_files` will be added to the function bundle. This follows the spec in https://github.com/netlify/pod-workflow/issues/189.

Currently, an `includedFilesBasePath` property needs to be supplied, so that `zip-it-and-ship-it` knows where to resolve the globs from. This will be removed in a future major version, favouring a setup where `@netlify/config` resolves the globs and supplies `zip-it-and-ship-it` an array of absolute paths.

**- Test plan**

Added a new test.

unnamed


src/runtimes/node/src_files.js
```diff
@@ -1,6 +1,44 @@
...
+    },
+    { include: [], exclude: [] },
+  )
+  const pathGroups = await Promise.all(
+    include.map((expression) => pGlob(expression, { absolute: true, cwd: basePath, ignore: exclude })),
+  )
+
+  // `pathGroups` is an array containing the paths for each expression in the
+  // `include` array. We flatten it into a single dimension.
+  const paths = pathGroups.reduce((acc, group) => [...acc, ...group], [])
```
ehmicky, line 37: I think we're using the flat polyfill?

```suggestion
  const paths = pathGroups.flat()
```

tests/fixtures/included_files/netlify/functions/func1.js
```diff
@@ -0,0 +1,24 @@
...
+    const data = await pReadFile(filePath, 'utf8')
+
+    return {
+      statusCode: 200,
+      body: data,
+    }
+  } catch (_) {
+    return {
+      statusCode: 500,
+      body: 'Uh-oh',
```
ehmicky, line 21: :-p

ehmicky (approved):

src/runtimes/node/src_files.js
eduardoboucas: Ah, perfect! That's much better, thanks.

PR #447, opened 2021-04-30
feat: add support for sourcemaps

### LINKED ISSUE #414: Add support for source maps

Feature.

When using functions created by `zip-it-and-ship-it` for local development, it would be useful to have access to source maps to ease debugging.

We could add a `sourcemap` configuration property, which could potentially be enabled by default when `archiveType` is `none`.

This would improve the experience of serving bundled functions in the CLI (https://github.com/netlify/cli/pull/2094).

eduardoboucas:
**- Summary**

Adds support for sourcemaps via a new configuration property:

- `nodeSourcemap` (Boolean): Whether to generate a sourcemap for a given function. Defaults to `false`.

Closes #414.

**- Test plan**

New tests added.

download


src/runtimes/node/bundler.js
```diff
@@ -56,16 +50,51 @@ const bundleJsFile = async function ({
...
+      warnings,
+    }
   } catch (error) {
     error.customErrorInfo = { type: 'functionsBundling', location: { functionName: name } }
 
     throw error
   }
 }
 
+const getCleanupFunction = (bundlePath, sourcemapPath) => async () => {
```
eduardoboucas, line 80: Extracted into a separate function to make the linter happy.

src/runtimes/node/bundler.js
```diff
@@ -8,9 +8,17 @@ const semver = require('semver')
 
 const { getBundlerTarget } = require('./bundler_target')
 const { externalNativeModulesPlugin } = require('./native_modules/plugin')
+const { processSourcemap } = require('./sourcemap')
 
 const pUnlink = promisify(fs.unlink)
 
+const supportsAsyncAPI = semver.satisfies(process.version, '>=9.x')
```
eduardoboucas, line 15: No functionality has changed here. Extracted from `bundleJsFile` to conform to the linter rules.

src/runtimes/node/zip_esbuild.js
```diff
@@ -9,6 +9,19 @@ const { zipNodeJs } = require('../../zip_node')
 const { bundleJsFile } = require('./bundler')
 const { getSrcFilesAndExternalModules } = require('./src_files')
 
+const getAliases = ({ bundlePath, mainFile, sourcemapPath, srcDir }) => {
```
eduardoboucas, line 12: Extracted from `zipEsbuild` to conform with linter rules. The only functionality that changed here was that `aliases` is now a `Map` instead of the object, so that we can conditionally add elements to it.

src/runtimes/node/bundler.js
```diff
@@ -56,16 +48,47 @@ const bundleJsFile = async function ({
...
     throw error
   }
 }
 
+const getCleanupFunction = (bundlePath, sourcemapPath) => async () => {
+  try {
+    await pUnlink(bundlePath)
+
+    if (sourcemapPath) {
+      await pUnlink(sourcemapPath)
```
ehmicky, line 79: [sand] As a tiny performance optimization, should this be done in parallel. Also, should the function be generalized so it can be re-used? Something along the lines of:

```js
const getCleanupFunction = (...paths) => async () => {
  await Promise.all(paths.filter(Boolean).map(cleanupFile))
}

const cleanupFile = async (path) => {
  try {
    await pUnlink(path)
  } catch (_) {}
}
```

src/runtimes/node/sourcemap.js
```diff
@@ -0,0 +1,46 @@
...
+const processSourcemap = async ({ bundlePath, pathFormat: pathFormatConfig, sourcemapPath, srcDir }) => {
+  if (!sourcemapPath) {
+    return
+  }
+
+  const bundleDir = dirname(bundlePath)
+  const pathFormat = getPathFormat(pathFormatConfig)
+  const data = await pReadFile(sourcemapPath, 'utf8')
+  const sourcemap = JSON.parse(data)
+  const newSources = sourcemap.sources.map((path) => {
```
ehmicky, line 27: Is this required even when setting the right `sourceRoot` in the source map file?

eduardoboucas: Oh no! This entire file is not needed anymore. I should've removed it in the last commit. I'm really sorry! 🤦🏻 

Done in https://github.com/netlify/zip-it-and-ship-it/pull/447/commits/68253759c76c0b07e1da1f73a932c76f6bd0a892.

src/runtimes/node/sourcemap.js
```diff
@@ -0,0 +1,46 @@
...
+    const absolutePath = resolve(bundleDir, path)
+
+    if (pathFormat === PATH_FORMAT_ABSOLUTE) {
+      return absolutePath
+    }
+
+    const relativePath = relative(srcDir, absolutePath)
```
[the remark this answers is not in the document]
ehmicky: Please disregard (since the file was removed)

ehmicky (approved):

src/runtimes/node/bundler.js
eduardoboucas: I did this in https://github.com/netlify/zip-it-and-ship-it/pull/447/commits/d9260d3bed870aed02eb4a1a63681e0f6ba02db0. Does that look okay to you?

ehmicky: Yes! :+1:

eduardoboucas: Great! Thanks for the tip.

PR #472, opened 2021-05-18
chore: run CI on `latest` Node.js

ehmicky:
Part of https://github.com/netlify/team-dev/issues/19

This runs the CI on the `latest` Node.js version.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [10.3.0, 16.x]
+        node-version: [10.3.0, '*']
         exclude:
           - os: macOS-latest
             node-version: 10.3.0
@@ -27,11 +27,12 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
+          check-latest: true
       - name: Install dependencies
         run: npm ci
       - name: Linting
         run: npm run format:ci
-        if: "${{ matrix.node-version == '16.x' }}"
+        if: "${{ matrix.node-version == '*' }}"
       - name: Tests
         run: npm run test:ci
       - name: Get test coverage flags
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -16,7 +16,8 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@v2
         with:
-          node-version: '16.x'
+          node-version: '*'
+          check-latest: true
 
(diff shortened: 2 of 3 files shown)


eduardoboucas:
We should probably update https://github.com/netlify/zip-it-and-ship-it/blob/main/.github/workflows/benchmark.yml#L19 too. There's also https://github.com/netlify/zip-it-and-ship-it/blob/main/.github/workflows/release-please.yml#L25, but I wonder if we want to update that one too?

ehmicky:
Good point, I missed those. Fixed!
Changing all of them would make sense with the goal of not having to update them every 6 months.

.github/workflows/workflow.yml
```diff
@@ -27,11 +27,12 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
+          check-latest: true
```
eduardoboucas, line 30: What happens if we don't use `check-latest` here? From https://github.com/actions/setup-node:

> When set to false, the action tries to first resolve a version of node from the local cache. (...) The local version of Node in cache gets updated every couple of weeks. If unable to find a specific version in the cache, the action will then attempt to download a version of Node.

And:

> Setting check-latest to true has performance implications as downloading versions of Node is slower than using cached versions

I wonder about these performance implications as we try to speed up the CI runs. If we use `*` without `check-latest`, does it mean we may only get an updated version after GitHub updates the image? If so, how do we feel about this trade-off?

erezrokah: Looks like download time is a few seconds
Linux is 4s, MacOS is 5s and Windows is 9s (based on this run).
This is a single sample, but I don't expect it to be much higher.

I've also subscribed to https://github.com/actions/setup-node/issues/26 so we might be able to use `lts` in the future

eduardoboucas: Cool, in case that this looks good to me! 🚀

eduardoboucas (approved):

erezrokah (approved):

.github/workflows/workflow.yml
ehmicky: I added `check-latest: true` because without it, `*` is tried against the cached Node.js versions which seemed to be 10, 12 and 14. So without it, `*` was resolving to `14.*.*`. I have experimented with several other ranges, and none worked: `latest`, `current`, `*.*.*`. The only one which did work was to use something like `>1` but I was not 100% sure how this would behave.

ehmicky: Note: there is also this [ongoing PR](https://github.com/actions/setup-node/pull/239) which is actually adding a library of mine to `actions/setup-node` and would allow using `latest` without needing `check-latest: true`.

PR #474, opened 2021-05-19
feat: remove src directory from function bundle

eduardoboucas:
**- Summary**

This PR removes the `src/` sub-directory from Node.js function deployments. This means that an input file at `{FUNCTIONS DIR}/file.js` will be placed at `{LAMBDA ROOT}/file.js` instead of `{LAMBDA ROOT}/src/file.js`.

The only exception to this is when there's a potential for a naming conflict, which might happen if there's a file at the root of the project with the same name as one of the functions _and_ that file is being implicitly (via `require`/`import`) or explicitly (via `functions.included_files`) included in the bundle. When this happens, `zip-it-and-ship-it` falls back to the current behaviour of placing all user-defined files inside a `src/` sub-directory.

Additionally, to ensure that paths are consistent between local and production environments, `zip-it-and-ship-it` now accepts an optional `basePath` property, which contains the shortest path that will be included in the bundle (in other words, the root path of the lambda, from which `process.cwd()` will be computed from).

_Example:_

```
/
  projects/
    my-site/ <---- PROJECT ROOT
      netlify/
        functions/
          func1.js
      netlify.toml
```

Currently, the base path will be `/projects/my-site/netlify/functions/func1`, so the function will live at `{LAMBDA ROOT}/src/func1.js` in production, whereas locally it will live at `{PROJECT ROOT}/netlify/functions/func1.js`.

With this PR, one can set the base path to be the project root directory (i.e. `/projects/my-site`), so the function file will live at `{LAMBDA ROOT}/netlify/functions/func1.js`, the same as locally.

Implements https://github.com/netlify/pod-workflow/issues/214.

**- Test plan**

Added a new test.

ba27b70e9e58fcbcf26a14578047cbb1


tests/fixtures/base_path/content/post3.md
```diff
@@ -0,0 +1 @@
+# To tell you I'm sorry for everything that I've done
```
ehmicky, line 1: You're forgiven @eduardoboucas :pray:

src/runtimes/node/zip_esbuild.js
```diff
@@ -22,6 +22,19 @@ const getAliases = ({ bundlePath, mainFile, sourcemapPath, srcDir }) => {
...
 
+const getFunctionBasePath = ({ basePathFromConfig, mainFile, supportingSrcFiles }) => {
+  // If there is a base path defined in the config, we use that.
+  if (basePathFromConfig !== undefined) {
+    return basePathFromConfig
+  }
+
+  // If not, the base path is the common path prefix between all the supporting
+  // files and the main file.
+  const dirnames = supportingSrcFiles.map((filePath) => normalize(dirname(filePath)))
```
ehmicky, line 33:
```suggestion
  const dirnames = [...supportingSrcFiles, mainFile].map((filePath) => normalize(dirname(filePath)))
```

[sand] Since `mainFile` has uses the `dirname()` + `normalize()`

ehmicky (approved):

eduardoboucas:
I realised that while this is not a breaking change in the sense that we're not violating any of our public contracts, it does technically mean that any functions that are relying on the assumption that files will be placed under a `src/` directory might break.

For this reason, I've introduced an `experimentalHandlerV2` configuration property, which we can then hook up to `@netlify/config`, an environment variable or a feature flag (or a combination of these). Unless this property is set to `true`, the old behaviour is preserved, i.e. everything except the entry file will be placed under a `src/` directory.

This allows us to gradually roll this out and reduce any risks associated.

tests/main.js
```diff
@@ -1020,6 +1020,73 @@ testBundlers(
...
+  'Places all user-defined files at the root of the target directory if the `experimentalHandlerV2` configuration property is defined',
+  [ESBUILD, ESBUILD_ZISI, DEFAULT],
+  async (bundler, t) => {
+    const fixtureName = 'base_path'
+    const { tmpDir } = await zipNode(t, `${fixtureName}/netlify/functions1`, {
+      opts: {
+        basePath: join(FIXTURES_DIR, fixtureName),
+        config: {
+          '*': {
+            experimentalHandlerV2: true,
```
ehmicky, line 1033: Should we add a test checking that `experimentalHandlerV2` defaults to `false` and has the impact of using the old mechanism?

ehmicky (approved):
This makes sense!
A LaunchDarkly feature flag would be nice since it would allow rolling this out progressively.

tests/main.js
eduardoboucas: Definitely! Done in https://github.com/netlify/zip-it-and-ship-it/pull/474/commits/99b4ff8770c8e35fc68c7b49ce7b83eccae23657.

PR #478, opened 2021-05-21
feat: remove Node 8 support

### LINKED ISSUE #294: Chore: drop Node.js 8 support
Following up with https://github.com/netlify/cli/issues/1568, we should be good now with dropping Node.js 8 support for the `zip-it-and-ship-it`.

eduardoboucas:
**- Summary**

Officially removes support for Node 8.

Closes #294.

**- Test plan**

N/A

a-beary-sincere-goodbye

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "tmp-promise": "^3.0.2"
   },
   "engines": {
-    "node": ">=8.3.0"
+    "node": ">=10.3.0"
   },
   "ava": {
     "files": [
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "tmp-promise": "^3.0.2"
   },
   "engines": {
-    "node": ">=10.3.0"
+    "node": ">=10.18.0"
   },
   "ava": {
     "files": [
--- a/renovate.json5
+++ b/renovate.json5
@@ -6,32 +6,9 @@
   automerge: false,
   packageRules: [
     {
-      // Those cannot be upgraded to a major version until we drop support for Node 8
-      packageNames: [
-        'archiver',
-        'ava',
-        'cp-file',
-        'del',
-        'eslint',
-        'execa',
-        'find-up',
-        'get-stream',
-        'locate-path',
-        'p-map',
-        'pkg-dir',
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -94,7 +94,7 @@
     "tmp-promise": "^3.0.2"
   },
   "engines": {
-    "node": ">=8.3.0"
+    "node": ">=10.3.0"
```
eduardoboucas, line 97: I was torn on whether to use `10.0.0`. which we technically should support, or `10.3.0`, the lowest version we test against in the CI. @ehmicky what do you think?

ehmicky, line 97: `cli` is using `>=10.18.0`

https://github.com/netlify/cli/blob/2a662e4546520f31e003306b2d3e120d021f8ce4/package.json#L14

[Link to builds usage data](https://github.com/netlify/engineering/issues/50#issuecomment-704451333) from 2 months ago. Looking at it, I am wondering whether we should also pick `10.18.0`?

cc @JGAntunes too

renovate.json5
```diff
@@ -5,33 +5,10 @@
   dependencyDashboard: true,
   automerge: false,
   packageRules: [
-    {
-      // Those cannot be upgraded to a major version until we drop support for Node 8
-      packageNames: [
```
eduardoboucas, line 10: There's a possibility that some of these packages will also have problems in Node 10, but I think we can wait for Renovate to create the PRs and we evaluate them one-by-one. Unless anyone thinks otherwise?

ehmicky, line 10: :tada:

package.json
ehmicky: I answered in the other comment. :+1:

renovate.json5
ehmicky: Absolutely :+1: 

Many packages from Sindre Sorhus have dropped support for Node 10 recently, but we could check one by one.

package.json
eduardoboucas: That works for me. I'll update.

ehmicky (approved):

PR #501, opened 2021-05-31
chore: add packageRules to Renovate config

eduardoboucas:
**- Summary**

Some packages can't be upgraded until we drop support for Node 10. This PR updates the Renovate config to that effect.

**- Test plan**

N/A

🐘

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -8,6 +8,9 @@
     {
       // These cannot be upgraded to a major version until we drop support for Node 10
       packageNames: ['p-map', 'yargs'],
+      major: {
+        enabled: false,
+      },
     },
     {
       // Fake dependencies used in tests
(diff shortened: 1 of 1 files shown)


erezrokah (approved):
FYI this will auto close https://github.com/netlify/zip-it-and-ship-it/pull/494 and https://github.com/netlify/zip-it-and-ship-it/pull/497, but we should still update `p-map` to `v4` and `yargs` to `v16`

erezrokah:
We can use something like https://github.com/netlify/renovate-config/blob/21eae43b07e39ac00251a6411fe297e01e658fbe/default.json#L27

eduardoboucas:
Oh, good point! I'll address that separately. Thanks!

eduardoboucas:
@erezrokah Added that in https://github.com/netlify/zip-it-and-ship-it/pull/501/commits/6d97ebcf576aa129a9393d06d1cd5c465bbfc44b.

erezrokah (approved):
🚀

PR #504, opened 2021-05-31
chore(renovate): remove redundant rule

erezrokah:
We don't need this rule as we have more granular ones.

Sorry @eduardoboucas I missed this in the PR review


eduardoboucas:
Ah, good point! Thanks @erezrokah.

eduardoboucas (approved):

PR #521, opened 2021-06-07
feat: handle dynamic imports

eduardoboucas:
**- Summary**

This PR takes advantage of the new `onDynamicImport` plugin hook of esbuild to intercept imports with dynamic expressions and include in the bundle the files necessary to make the function work at runtime.

For example, when parsing the following statement, all `.json` files in the `./lang` directory will be included in the bundle so that the `require` call can work at runtime.

```js
require(`./lang/${req.params.lang}.json`)
```

This PR includes support for parsing template literals only. Other expression types will be added in subsequent PRs.

**- Test plan**

Added a new test.

istockphoto-962467250-612x612


src/runtimes/node/bundler.js
```diff
@@ -36,20 +39,39 @@ const bundleJsFile = async function ({
...
   const nodeTarget = getBundlerTarget(config.nodeVersion)
 
-  // esbuild will format `sources` relative to the sourcemap file, which is a
-  // sibling of `bundlePath`. We use `sourceRoot` to establish that relation.
-  // They are URLs, so even on Windows they should use forward slashes.
-  const sourceRoot = dirname(bundlePath).replace(/\\/g, '/')
+  // esbuild will format `sources` relative to the sourcemap file, which lives
+  // in `destFolder`. We use `sourceRoot` to establish that relation. They are
+  // URLs, not paths, so even on Windows they should use forward slashes.
+  const sourceRoot = destFolder.replace(/\\/g, '/')
```
eduardoboucas, line 74: `bundlePath` was being computed with `join(destFolder, jsFilename)`, so it doesn't make sense to be getting its dirname here. We can use `destFolder` directly.

src/runtimes/node/bundler.js
```diff
@@ -88,18 +114,40 @@ const bundleJsFile = async function ({
...
+  // entry of the `outputs` object.
+  Object.entries(outputs).forEach(([path, output]) => {
+    const filename = basename(path)
+    const extension = extname(path)
+    const absolutePath = join(destFolder, filename)
 
-  if (relativePath) {
-    return resolve(relativePath)
-  }
+    if (output.entryPoint && basename(output.entryPoint) === basename(srcFile)) {
```
eduardoboucas, line 136: This detects the main function file, i.e. the one that needs to be the input file of esbuild.

src/runtimes/node/bundler.js
```diff
@@ -88,18 +114,40 @@ const bundleJsFile = async function ({
...
+      // Ensuring the main file has a `.js` extension.
+      const normalizedSrcFile = getPathWithExtension(srcFile, '.js')
+
+      bundlePaths.set(absolutePath, normalizedSrcFile)
+    } else if (extension === '.js' || filename === `${bundleFilename}.map`) {
```
eduardoboucas, line 141: This looks for any additional files, as long as they are JS files (e.g. dynamic import shims) or sourcemap files. If a file doesn't fall into this criteria, we don't care about it.

tests/helpers/main.js
```diff
@@ -31,6 +32,11 @@ const zipFixture = async function (t, fixture, { length, fixtureDir, opts = {} }
   const { path: tmpDir } = await getTmpDir({
     prefix: `zip-it-test-bundler-${bundlerString}`,
   })
+
+  if (env.ZISI_KEEP_TEMP_DIRS !== undefined) {
```
eduardoboucas, line 36: When the `ZISI_KEEP_TEMP_DIRS` variable is set, we likely want to inspect the temp directories after the tests, so it makes sense to print their path too.

src/runtimes/node/dynamic_imports/parser.js
```diff
@@ -0,0 +1,140 @@
+const { join, relative, resolve } = require('path')
+
+const acorn = require('acorn')
+
+const ECMA_VERSION = 2017
```
ehmicky, line 5: Why not 2021?

tests/main.js
```diff
@@ -1224,6 +1224,38 @@ test('Returns an empty list of modules with dynamic imports if the modules are m
...
+    opts: { basePath: join(FIXTURES_DIR, fixtureName), config: { '*': { nodeBundler: ESBUILD } } },
+  })
+
+  // eslint-disable-next-line import/no-dynamic-require, node/global-require
+  const func = require(`${tmpDir}/function.js`)
+
+  t.deepEqual(func('en')[0], ['yes', 'no'])
+  t.deepEqual(func('en')[1], ['yes', 'no'])
+  t.deepEqual(func('pt')[0], ['sim', 'não'])
+  t.deepEqual(func('pt')[1], ['sim', 'não'])
```
ehmicky, line 1255: :portugal:

src/runtimes/node/dynamic_imports/parser.js
```diff
@@ -0,0 +1,140 @@
...
+
+      case 'Literal':
+        return operand.value
+
+      default:
+        throw new Error('Expression member not supported')
+    }
+  })
+
+  return operands.flat()
```
ehmicky, line 99: [sand] `flatMap()` could be used instead of `map()` + `flat()`

ehmicky (approved):

src/runtimes/node/dynamic_imports/parser.js
eduardoboucas: Good catch! Done in https://github.com/netlify/zip-it-and-ship-it/pull/521/commits/eab6fdb2ef7cf70883ba9eec81cd7f2cf473450b. Thanks. 🙌🏻

src/runtimes/node/dynamic_imports/parser.js
eduardoboucas: I was aligning this with the default esbuild target, but these are separate things in reality. There's no harm in allowing acorn to parse the latest ES version. Done in https://github.com/netlify/zip-it-and-ship-it/pull/521/commits/9717f5b2f20fea75c06fddef23c78f45ff3505c8. Thanks!

PR #537, opened 2021-06-17
feat: add support for multiple source directories

eduardoboucas:
**- Summary**

Makes `zipFunctions` capable of bundling functions from multiple directories if the first argument is an array. When this happens, the order of the directories in the arrays matters, as rightmost directories will take precedence in case of naming conflicts.

**- Test plan**

Added a new test.

armadillo

(cc @ascorbic)


ascorbic:
Nice! Have we decided how we'll expose this in builds? Just add an automatic `.netlify/functions` dir, or allow plugins/toml to add multiple funcitons folders?

eduardoboucas:
> Nice! Have we decided how we'll expose this in builds? Just add an automatic `.netlify/functions` dir, or allow plugins/toml to add multiple funcitons folders?

I think the former would be simpler, but I'm keeping the implementation in zip-it-and-ship-it generic enough that it can support both scenarios.

src/utils/fs.js
```diff
@@ -11,6 +11,30 @@ const safeUnlink = async (path) => {
...
+// filenames within those directories, if at least one of the directories
+// exists. If not, an error is thrown.
+const listFunctionsDirectories = async function (srcFolders) {
+  const filenamesByDirectory = await Promise.all(
+    srcFolders.map(async (srcFolder) => {
+      try {
+        const filenames = await listFunctionsDirectory(srcFolder)
+
+        return filenames
+      } catch (error) {
```
ehmicky, line 24: Instead of ignoring exceptions, should we check for `srcFolder` existence before running `listFunctionsDirectory()`?

ehmicky (approved):
:tada:

src/utils/fs.js
eduardoboucas: `listFunctionsDirectory` itself follows this logic of treating exceptions from `pReaddir` as signals that the directory doesn't exist, so we'll only ever fall into the `catch` branch here if one of the directories doesn't exist. Do you have a specific case in mind for also checking for the existence of the directory, which means another I/O call?

ehmicky: Good point about the additional I/O call :+1: 
I am thinking of the case where `readdir` fails for another reason than `ENOENT`, e.g. `EPERM`. In that case, it might be good to propagate the error?

eduardoboucas: Sorry for the late reply!

I see what you mean, but I think that's more about changing the existing `listFunctionsDirectory`, which is currently not propagating the error (or, more specifically, is swallowing the error message). Would you be okay with merging this as is and then figuring out how best to propagate errors in a subsequent PR?

ehmicky: Sounds good :+1:

PR #546, opened 2021-06-23
feat: add more expressions to dynamic import parser

eduardoboucas:
**- Summary**

Adds support for more expression types when parsing dynamic import expressions.

**- Test plan**

Adjusted existing tests. Also added a new test which asserts that we leave alone any imports whose expressions we can't convert at build time.

### DIFF
--- a/src/runtimes/node/dynamic_imports/parser.js
+++ b/src/runtimes/node/dynamic_imports/parser.js
@@ -26,21 +26,50 @@ const getAbsoluteGlob = ({ basePath, globNodes, resolveDir }) => {
   return absoluteGlob
 }
 
+// Returns GLOB_WILDCARD for AST nodes that are accepted as part of a dynamic
+// expression and convertable to a wildcard character. This determines whether
+// we convert an expression or leave it alone. For example:
+//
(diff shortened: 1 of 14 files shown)


src/runtimes/node/dynamic_imports/parser.js
```diff
@@ -26,21 +26,50 @@ const getAbsoluteGlob = ({ basePath, globNodes, resolveDir }) => {
...
+const getWildcardFromASTNode = (node) => {
+  switch (node.type) {
+    case 'CallExpression':
+    case 'ConditionalExpression':
+    case 'Identifier':
+    case 'MemberExpression':
+      return GLOB_WILDCARD
+
+    default:
+      throw new Error('Expression member not supported')
```
erezrokah, line 51: This is handled in the `parseExpression` `try/catch` correct?

erezrokah (approved):

src/runtimes/node/dynamic_imports/parser.js
eduardoboucas: Yes, exactly!

PR #548, opened 2021-06-24
fix: use temporary directory for esbuild

### LINKED ISSUE #547: Naming collisions in dynamic import shims
When generating shim files for dynamic imports in esbuild, it's possible that naming collisions occur, since all files will be written to the same directory. As a solution, we should make esbuild write files to a temporary directory, namespaced by function.

eduardoboucas:
**- Summary**

Uses a temporary directory as the esbuild target, namespaced by function, avoiding any naming collisions between files generated by different functions.

Closes #547.

**- Test plan**

Added a new test, which includes 30 functions with equal names and dynamic import expressions. This allowed me to replicate the issue described in #547, and verify that this PR fixes it.


erezrokah (approved):

src/runtimes/node/bundler.js
```diff
@@ -26,17 +27,21 @@ const supportsAsyncAPI = semver.satisfies(process.version, '>=9.x')
...
-  destFolder,
   externalModules = [],
   ignoredModules = [],
   name,
   srcDir,
   srcFile,
 }) {
+  // We use a temporary directory as the destination for esbuild files to avoid
+  // any naming conflicts with files generated by other functions.
+  const targetDirectory = await tmpName()
```
ehmicky, line 43: Just double-checking: are we creating `targetDirectory` intermediary directories if they do not exist?
If not, `tmp-promise` `dir()` method can be used for that purpose.

eduardoboucas: > Just double-checking: are we creating `targetDirectory` intermediary directories if they do not exist?
> If not, `tmp-promise` `dir()` method can be used for that purpose.

Yes, esbuild takes care of creating all the intermediary directories, so I don't think we need to do that ourselves here.

PR #554, opened 2021-06-26
chore: add Next.js E2E test workflow

eduardoboucas:
Adds an end-to-end testing workflow with a Next.js site.


CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
 ### Features
 
-* add more expressions to dynamic import parser ([#546](https://www.github.com/netlify/zip-it-and-ship-it/issues/546)) ([f29156e](https://www.github.com/netlify/zip-it-and-ship-it/commit/f29156ef5395795ad40e9df9556908b58aec6ae9))
+* add support for more expression types to dynamic import parser ([#546](https://www.github.com/netlify/zip-it-and-ship-it/issues/546)) ([f29156e](https://www.github.com/netlify/zip-it-and-ship-it/commit/f29156ef5395795ad40e9df9556908b58aec6ae9))
```
eduardoboucas, line 15: This change was so that the E2E flow was triggered: https://github.com/netlify/zip-it-and-ship-it/pull/554/files#diff-d396261d6ce3045ca78e2b914f9b95deb0b675668f8f4866293b2e6451019bfcR7

.github/workflows/e2e-nextjs.yml
```diff
@@ -0,0 +1,83 @@
...
+          ''utf-8''));config.functions = { ...config.functions, node_bundler: ''${{ matrix.bundler }}''
+          };writeFileSync(''netlify.toml'', stringify(config));"'
+        working-directory: netlify-plugin-nextjs/demo
+      - name: Symlinking zip-it-and-ship-it (1/2)
+        run:
+          rm -rf $GITHUB_WORKSPACE/netlify-cli/node_modules/@netlify/zip-it-and-ship-it && ln -s
+          $GITHUB_WORKSPACE/zip-it-and-ship-it $GITHUB_WORKSPACE/netlify-cli/node_modules/@netlify/zip-it-and-ship-it
+      - name: Symlinking zip-it-and-ship-it (2/2)
+        run:
+          rm -rf $GITHUB_WORKSPACE/netlify-cli/node_modules/@netlify/build/node_modules/@netlify/zip-it-and-ship-it &&
```
erezrokah, line 57: Should we do this for `functions-utils` too?

Also should we use `npm link` instead of manually linking?

.github/workflows/e2e-nextjs.yml
```diff
@@ -0,0 +1,83 @@
...
+          path: netlify-plugin-nextjs
+      - name: Checking out netlify-cli
+        uses: actions/checkout@v2
+        with:
+          repository: netlify/cli
+          path: netlify-cli
+      - name: Installing Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
erezrokah, line 33: Do we want to use the [latest](https://github.com/netlify/zip-it-and-ship-it/blob/c7b62ca7080b6a1bbe77483fde32e9bcde326855/.github/workflows/workflow.yml#L16) Node.js version?

.github/workflows/e2e-nextjs.yml
```diff
@@ -0,0 +1,83 @@
...
+        working-directory: zip-it-and-ship-it
+      - name: Building and deploying site
+        run:
+          NETLIFY_EXPERIMENTAL_FUNCTION_HANDLER_V2=true $GITHUB_WORKSPACE/netlify-cli/bin/run deploy --build --json
+          --site ${{ secrets.NETLIFY_SITE_ID }} --auth ${{ secrets.NETLIFY_TOKEN }} --functions .netlify/functions >
+          .netlify-deploy-log.json
+        working-directory: netlify-plugin-nextjs/demo
+      - name: Parsing deploy result
+        run: |
+          node -e "console.log('deploy_log_url=' + require('./.netlify-deploy-log.json').logs)" >> $GITHUB_ENV
```
erezrokah, line 71: Instead of inlining JavaScript, could we use something like
`echo deploy_log_url=$(cat .netlify-deploy-log.json | jq -r .logs) >> $GITHUB_ENV`

.github/workflows/e2e-nextjs.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Checking out netlify-cli
+        uses: actions/checkout@v2
+        with:
+          repository: netlify/cli
+          path: netlify-cli
+      - name: Installing Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: Installing netlify-cli
```
erezrokah, line 34: The CLI is already a local dependency of the plugin:
https://github.com/netlify/netlify-plugin-nextjs/blob/1c06d1323312bf7b4ff7f4c228978b96a73f2917/package.json#L90
so we might be able to save some time and skip this part (and the checkout).

.github/workflows/e2e-nextjs.yml
```diff
@@ -0,0 +1,83 @@
...
+        working-directory: netlify-cli
+      - name: Installing netlify-plugin-nextjs
+        run: npm install
+        working-directory: netlify-plugin-nextjs
+      - name: Installing netlify-plugin-nextjs demo
+        run: npm install
+        working-directory: netlify-plugin-nextjs/demo
+      - name: Patching netlify.toml to change the bundler
+        if: "${{ matrix.bundler != 'default' }}"
+        run:
```
erezrokah, line 45: What do you think on extracting this into a dedicate JS file?

erezrokah (approved):
This is 🚀  @eduardoboucas 

Added some comments, none of them should block this in my opinion.

.github/workflows/e2e-nextjs.yml
eduardoboucas: > Should we do this for functions-utils too?

Yes, good call!

> Also should we use npm link instead of manually linking?

I did try with `npm link` but it wasn't working as expected. How would it work when linking transitive dependencies? Would we run `npm link @netlify/zip-it-and-ship-it` from within the various packages in `node_modules`?

.github/workflows/e2e-nextjs.yml
eduardoboucas: Yes! I'll change.

.github/workflows/e2e-nextjs.yml
eduardoboucas: Oh, I didn't know that. It's definitely worth exploring.

.github/workflows/e2e-nextjs.yml
eduardoboucas: Will `jq` be available? After committing this I realised there's a `fromJSON` utility in GitHub Actions: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#fromjson. That could do the trick?

.github/workflows/e2e-nextjs.yml
eduardoboucas: I agree that it looks a bit ugly, but I do like the idea of having the entire flow contained in this file. I'm happy to revisit, though.

.github/workflows/e2e-nextjs.yml
erezrokah: > Would we run `npm link @netlify/zip-it-and-ship-it` from within the various packages in `node_modules`?

Yes, I think we would need to that for each copy

.github/workflows/e2e-nextjs.yml
erezrokah: `jq` should be available for linux https://github.com/actions/virtual-environments/blob/c3f3d8325f86e969eba47d14d6484f752d37b973/images/linux/Ubuntu2004-README.md#tools.

`fromJSON` seem much more elegant though

.github/workflows/e2e-nextjs.yml
erezrokah: I think we can revisit if the script gets more complicated

.github/workflows/e2e-nextjs.yml
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/554/commits/8b0c35fa76d232e6a37772a422ee449d77555021.

eduardoboucas:
@erezrokah Since your suggestions are not blocking, I think we can merge this as is and then iterate on the improvements. I created an issue to track that progress: https://github.com/netlify/zip-it-and-ship-it/issues/558.

PR #559, opened 2021-06-28
feat: add processDynamicNodeImports config property

eduardoboucas:
**- Summary**

Puts processing of dynamic import expressions behind a `processDynamicNodeImports` configuration property, so that we can roll out the feature more gracefully.

**- Test plan**

Added new test and adjusted existing tests.

### DIFF
--- a/src/runtimes/node/dynamic_imports/plugin.js
+++ b/src/runtimes/node/dynamic_imports/plugin.js
@@ -13,11 +13,12 @@ const { parseExpression } = require('./parser')
 // issues. Secondly, it parses the dynamic expressions and tries to include in
 // the bundle all the files that are possibly needed to make the import work at
 // runtime. This is not always possible, but we do our best.
-const getDynamicImportsPlugin = ({ basePath, includedPaths, moduleNames, srcDir }) => ({
+const getDynamicImportsPlugin = ({ basePath, includedPaths, moduleNames, processImports, srcDir }) => ({
   name: 'dynamic-imports',
   setup(build) {
     const cache = new Map()
 
+    // eslint-disable-next-line complexity
     build.onDynamicImport({}, async (args) => {
       const { expression, resolveDir } = args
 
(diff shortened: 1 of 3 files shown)


src/runtimes/node/dynamic_imports/plugin.js
```diff
@@ -13,19 +13,22 @@ const { parseExpression } = require('./parser')
...
       const { expression, resolveDir } = args
 
       await registerModuleWithDynamicImports({ cache, moduleNames, resolveDir, srcDir })
 
       // Don't attempt to parse the expression if the base path isn't defined,
       // since we won't be able to generate the globs for the included paths.
-      if (!basePath) {
+      // Also don't parse the expression if we're not interested in processing
+      // the dynamic import expressions.
+      if (!basePath || !processImports) {
```
erezrokah, line 31: Should we not register the plugin at all if the flag is off?

Asking mostly out of curiosity

erezrokah (approved):

src/runtimes/node/dynamic_imports/plugin.js
eduardoboucas: Great question! The plugin is actually doing two things, as described [here](https://github.com/netlify/zip-it-and-ship-it/blob/main/src/runtimes/node/dynamic_imports/plugin.js#L9-L15):

1. Detects dynamic imports
2. Tries to resolve dynamic imports

We need 1, otherwise we won't be able to show a warning message telling people that they have modules with dynamic imports and offering a suggestion for fixing potential issues. This has been in production for a while and hasn't caused any problems.

If we register the plugin but make it return `undefined`, which is what we're doing with the flag, we tell esbuild to leave the dynamic imports alone, i.e. don't inject a runtime shim.

eduardoboucas: Note that registering the plugin and always returning `undefined` is what we have in version 4.4.2, the version that is currently in production: https://github.com/netlify/zip-it-and-ship-it/blob/v4.4.2/src/runtimes/node/dynamic_imports.js#L11-L21

PR #567, opened 2021-07-04
feat: add support for feature flags

eduardoboucas:
**- Summary**

Unlike Netlify Build, zip-it-and-ship-it doesn't have an official feature flag mechanism. We've been using the functions configuration object to that effect, but that falls short in some situations.

For example, the config is only evaluated when a function is being processed, but we might want to have a feature flag that determines whether a function is even created or not, meaning the feature flag needs to come earlier in the process.

This PR adds a `featureFlags` property that is passed to both `zipFunction` and `zipFunctions` and subsequently supplied to  `getFunctionsFromPaths`.

**- Test plan**

New tests added.

main_xkdhX3G

### DIFF
--- a/src/feature_flags.js
+++ b/src/feature_flags.js
@@ -1,0 +1,13 @@
+// List of supported flags and their default value.
+const FLAGS = {}
+
+const getFlags = (input = {}, flags = FLAGS) =>
+  Object.entries(flags).reduce(
+    (result, [key, defaultValue]) => ({
+      ...result,
+      [key]: input[key] === undefined ? defaultValue : input[key],
+    }),
+    {},
+  )
+
+module.exports = { FLAGS, getFlags }
--- a/src/zip.js
+++ b/src/zip.js
@@ -3,6 +3,7 @@ const { resolve } = require('path')
 const makeDir = require('make-dir')
 const pMap = require('p-map')
 
+const { getFlags } = require('./feature_flags')
 const { getPluginsModulesPath } = require('./node_dependencies')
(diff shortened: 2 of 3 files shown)


src/feature_flags.js
```diff
@@ -0,0 +1,13 @@
+// List of supported flags and their default value.
+const FLAGS = {}
+
+const getFlags = (input = {}, flags = FLAGS) =>
```
ehmicky, line 4: [sand] Would `(input, flags = FLAGS) => ({ ...flags, ...input })` (or `(input, flags = FLAGS) => ({ ...flags, ...removeFalsy(input) })` if `input` might contain `undefined` keys) be simpler? 
One difference is that if `input` contains unknown flags, those would not be filtered, but that should not be a problem?

ehmicky (approved):

src/feature_flags.js
eduardoboucas: I would actually like `FLAGS` to contain all the feature flags that ZISI is currently accepting, and make it so that any supplied flags that don't belong to that list are ignored (e.g. https://github.com/netlify/zip-it-and-ship-it/pull/567/files#diff-84cbdd132bc7ff6db27e194fa841b40864de4a8f661e0a79273ff9b2477e9e3dR11-R15). I think this is important to ensure that any flags used throughout the codebase are declared in one place for greater visibility.

eduardoboucas: I'll merge this as is for now, but I'm happy to rework this later if you think we should.

ehmicky: Sounds good :+1:

PR #576, opened 2021-07-08
feat: build Go functions from source

eduardoboucas:
**- Summary**

Builds Go functions from source if the `buildGoSource` feature flag is enabled.

A Go function be its own sub-directory with a `main.go` file inside. `zip-it-and-ship-it` will compile it using `go build -ldflags "-s -w"`, with `GOOS` set to `linux`.

We'll likely want to allow parts of the build command (like `ldflags`) to be customisable via the `config` property. We can gradually add these over time.

**- Test plan**

Added new tests.

gophers_eating


erezrokah:
Once this is merged, published and added to the CLI - it would be great to suggest it as a solution to https://github.com/netlify/cli/issues/1147.

I have a repo that reproduces that issue in https://github.com/erezrokah/go-netlify-test

src/runtimes/go/builder.js
```diff
@@ -0,0 +1,33 @@
...
+const build = async ({ destPath, mainFile, srcDir }) => {
+  const functionName = basename(srcDir)
+
+  try {
+    await runCommand('go', ['build', '-o', destPath, '-ldflags', '-s -w'], {
+      cwd: srcDir,
+      env: {
+        GOOS: 'linux',
+      },
+    })
```
mraerino, line 15: consider setting `CGO_ENABLED=0` so that people get a failure when trying to include native libs which likely won't work well on lambda.

giving an option to enable if they are sure might be nice.
or if you allow people to pass arbitrary env vars to this command, consider setting `CGO_ENABLED=0` if there is no `CGO_ENABLED` given

src/runtimes/go/index.js
```diff
@@ -0,0 +1,86 @@
...
+const detectGoFunction = async ({ fsCache, path }) => {
+  const stat = await cachedLstat(fsCache, path)
+
+  if (!stat.isDirectory()) {
+    return false
+  }
+
+  const files = await cachedReaddir(fsCache, path)
+
+  return files.includes(MAIN_FILE_NAME)
```
mraerino, line 22: you could also use `go.mod` / `go.sum` as an indicator

mraerino (commented):
i don't know a lot about this service, but left some notes

src/runtimes/go/index.js
eduardoboucas: Would you make that a requirement? i.e. Forcing functions to have both a `main.go` _and_ a `go.mod` + `go.sum`? Could you not have a `vendor` directory instead?

src/runtimes/go/builder.js
eduardoboucas: I'd like to keep this initial PR simple and introduce any configuration properties in follow up PRs. Perhaps we can hardcode `CGO_ENABLED=0` for now, and then study the best way to make this configurable. Does that sound reasonable?

eduardoboucas:
> i don't know a lot about this service, but left some notes

@mraerino Sorry for throwing this on you without much context. I was precisely after your insights from the Go compilation side — thanks for that! 🙌🏻

src/runtimes/go/index.js
mraerino: not a requirement, just a signal

src/utils/fs.js
```diff
@@ -1,10 +1,29 @@
...
 const { promisify } = require('util')
 
+const pLstat = promisify(lstat)
 const pReaddir = promisify(readdir)
+const pReadFile = promisify(readFile)
 const pUnlink = promisify(unlink)
 
+// This caches multiple FS calls to the same path. It creates a cache key with
+// the name of the function and the path (e.g. "readdir:/some/directory").
+const cachedIOFunction = (func, cache, path, ...args) => {
```
erezrokah, line 12: When are we expecting to get cache hits?

Aren't the paths passed to `findFunctionsInPaths` distinct?

erezrokah (approved):
This looks good from a JS standpoint.

I'm not sure when the cache is needed though.

src/utils/fs.js
eduardoboucas: > Aren't the paths passed to findFunctionsInPaths distinct?

No. Each runtime will get the list of paths that have not been claimed by previous runtimes. So if you have a directory that doesn't contain any valid functions, it'll get passed to each runtime's `findFunctionsInPaths` and each runtime will try to read the directory and its contents to determine whether it contains a valid function, so that directory will go through a `stat` call (and a few other FS calls) for each runtime.

That's the duplication of work we're trying to avoid.

The JS runtime is not yet taking full advantage of this caching, because it's using `locate-path`, which we have no control over. I'll get to that in a future PR.

src/runtimes/go/builder.js
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/576/commits/23afdea7c21455bfb7397d275b799ce95762d97e.

src/utils/fs.js
erezrokah: > No. Each runtime will get the list of paths that have not been claimed by previous runtimes.

Cool, that makes sense. Thanks for clarifying

PR #580, opened 2021-07-14
fix: add basePath parameter to zipFunction

eduardoboucas:
**- Summary**

The `basePath` parameter, which is required for dynamic import processing to work, is present in `zipFunctions` but missing in `zipFunction`. This PR fixes that.

**- Test plan**

New test added.

🐞


erezrokah (approved):
🎸

PR #582, opened 2021-07-15
feat: use double asterisk in dynamic import shim glob

eduardoboucas:
**- Summary**

When processing dynamic import expressions, we create glob expressions by replacing variables with a wildcard character. For example:

- `./lang/${variable}.json` -> `./lang/*.json`
- `./scripts/${variable}` -> `./scripts/*`

By using `*` as the wildcard character, it means we'll only include the first level of files. So in the second example above, a file under a nested directory (e.g. `scripts/nested/foo.js`) would be left out.

This PR changes the wildcard character to `**` so that nested directories are also included.

**- Test plan**

Added new tests

🦁

### DIFF
--- a/src/runtimes/node/src_files.js
+++ b/src/runtimes/node/src_files.js
@@ -30,7 +30,7 @@ const getPathsOfIncludedFiles = async (includedFiles, basePath) => {
     { include: [], exclude: [] },
   )
   const pathGroups = await Promise.all(
-    include.map((expression) => pGlob(expression, { absolute: true, cwd: basePath, ignore: exclude })),
+    include.map((expression) => pGlob(expression, { absolute: true, cwd: basePath, ignore: exclude, nodir: true })),
   )
 
   // `pathGroups` is an array containing the paths for each expression in the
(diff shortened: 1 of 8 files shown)


src/runtimes/node/src_files.js
```diff
@@ -30,7 +30,7 @@ const getPathsOfIncludedFiles = async (includedFiles, basePath) => {
     { include: [], exclude: [] },
   )
   const pathGroups = await Promise.all(
-    include.map((expression) => pGlob(expression, { absolute: true, cwd: basePath, ignore: exclude })),
+    include.map((expression) => pGlob(expression, { absolute: true, cwd: basePath, ignore: exclude, nodir: true })),
```
eduardoboucas, line 33: This ignores directories, which in practice means that a directory will only be included if it has files inside it.

For example, imagine:

```
- full
    - file.js
- empty
```

The glob `**` without `nodir: true` would yield `full`, `full/file.js` and `empty`. We can't copy directories when using `archiveFormat: "none"`, so it'd throw an error when handling `full` and `empty`.

By using `nodir: true`, the glob would yield simply `full/file.js`. Upon copying, it would be placed in the right location (i.e. any parent directories would be created automatically).

erezrokah (approved):

PR #587, opened 2021-07-20
feat: build Rust functions from source

eduardoboucas:
**- Summary**

Builds Rust functions from source if the `buildRustSource` feature flag is enabled.

A Rust function be its own sub-directory with a `Cargo.toml` and `src/main.rs` files inside. `zip-it-and-ship-it` will compile it to the `x86_64-unknown-linux-musl` target.

We'll likely want to allow parts of the build command to be customisable via the `config` property. We can gradually add these over time.

**- Test plan**

Added new tests.

good-for-you-crab-1170x781


erezrokah (approved):
🔥

src/runtimes/rust/builder.js
```diff
@@ -0,0 +1,49 @@
...
+
+const tmp = require('tmp-promise')
+const toml = require('toml')
+
+const { lstat } = require('../../utils/fs')
+const { runCommand } = require('../../utils/shell')
+
+const { BUILD_TARGET, MANIFEST_NAME } = require('./constants')
+
+const build = async ({ srcDir }) => {
```
eduardoboucas, line 15: @mraerino @fasterthanlime Does this look okay to you from a Rust perspective? 🦀 

(You don't need to look at the whole PR, this file should contain most of the Rust-specific logic.)

src/runtimes/rust/builder.js
```diff
@@ -0,0 +1,49 @@
...
+  // We compile the binary to a temporary directory so that we don't pollute
+  // the user's functions directory.
+  const { path: targetDirectory } = await tmp.dir()
+  const functionName = basename(srcDir)
+
+  try {
+    await runCommand('cargo', ['build', '--target', BUILD_TARGET, '--release'], {
+      cwd: srcDir,
+      env: {
+        CARGO_TARGET_DIR: targetDirectory,
```
mraerino, line 25: you'll want to think about caching this to reduce build times in the future.
we do this for builds if the root of the repo is a Cargo project: https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh#L701-L704

building a rust project from scratch is usually quite expensive, so levering the cache should be a big win for most people.

for our own projects we use sccache with an S3 bucket, but that seems a bit overkill since it doesn't integrate well with our regular caching strategy.

mraerino (approved):
lgtm. have a consideration for caching, but not blocking

src/runtimes/rust/builder.js
eduardoboucas: I did think about that and I'm glad you brought it up.

I'm using a temporary directory here because zip-it-and-ship-it might be running on people's machines, in which case we don't want to be injecting files into their project directories.

But we could use a similar strategy to the one we use for builds. When running on buildbot, zip-it-and-ship-it could adopt a deterministic target directory (e.g. `.netlify/functions-rust-cache/{FUNCTION-NAME}`) and we'd add `.netlify/functions-rust-cache` to the list of directories that we cache in between builds.

And now that I think of it, this behaviour would also make sense in the local development environment, since we want functions to build as quickly as possible in response to changes. The key is that we use a directory that we know will be git-ignored (like `.netlify/`).

What do you think?

src/runtimes/rust/builder.js
```diff
@@ -0,0 +1,67 @@
...
+      },
+    })
+  } catch (error) {
+    const hasToolchain = await checkRustToolchain()
+
+    if (!hasToolchain) {
+      throw new Error(
+        'There is no Rust toolchain installed. Visit https://ntl.fyi/missing-rust-toolchain for more information.',
```
eduardoboucas, line 33: This would point to https://docs.netlify.com/configure-builds/manage-dependencies/#rust.

eduardoboucas:
In https://github.com/netlify/zip-it-and-ship-it/pull/587/commits/d8722cab922e1f7f3d13de238a3e80533065ab33 and https://github.com/netlify/zip-it-and-ship-it/pull/587/commits/8dcd1c31ecf98c6c5b71b4092189ff5793854d90 I've added a check for a missing Rust toolchain, throwing an error with an informative error message.

@erezrokah mind having a final look, please?

erezrokah (approved):

PR #601, opened 2021-08-02
feat: add support for manifest file

eduardoboucas:
**- Summary**

Adds a new `manifest` property to `zipFunctions`. When set, it defines the path where a JSON-formatted file will be created, containing the list of all created functions as well as additional metadata properties.

This file can then be used by the CLI to know whether to reuse previously bundled functions as part of `build` and `deploy` commands.

**- Test plan**

New test added.

1501921785-baby-turtles-9

### DIFF
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -1,3 +1,4 @@
+const { resolve } = require('path')
 const { arch, platform } = require('process')
 
 const { safeWriteFile } = require('./utils/fs')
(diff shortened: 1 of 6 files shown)


src/manifest.js
```diff
@@ -0,0 +1,22 @@
...
+const createManifest = async ({ functions, path }) => {
+  const formattedFunctions = functions.map(formatFunction)
+  const payload = {
+    functions: formattedFunctions,
+    system: { arch, platform },
+    timestamp: Date.now(),
+    version: MANIFEST_VERSION,
+  }
+
+  await safeWriteFile(path, JSON.stringify(payload))
```
ehmicky, line 17: Should exceptions be propagated instead?

ehmicky (approved):
:+1:

src/manifest.js
eduardoboucas: I'm glad you brought this up, because I was on the fence about it.

I did it this way because the existence of the manifest doesn't impact the correctness of the bundle so I thought the operation should still succeed. However, since the manifest creation is opt-in, it makes more sense for any errors to be propagated, as you say.

Done in https://github.com/netlify/zip-it-and-ship-it/pull/601/commits/3c53e0f124d58ee673847a63a28808d9e803090f.

Issue #617, opened 2021-08-20
Detect functions with ES Modules


eduardoboucas:
Currently, functions with ES Modules are supported only when using esbuild. It would be great if zip-it-and-ship-it could detect when a function uses ES Modules and automatically use esbuild to bundle it, just like what happens with TypeScript.

charliegroll:
POC for possible module detection: https://github.com/netlify/zip-it-and-ship-it/pull/625

eduardoboucas:
(Adding some loose thoughts here.)

My main concern with this feature is the performance implications of doing an additional parse of each function just to determine whether it uses ESM or not. When using the legacy bundler, we're already parsing files with `acorn` (via [`precinct`](https://www.npmjs.com/package/precinct)), but I don't think we can tap into that data.

This feels particularly wasteful when processing functions generated dynamically by build plugins (e.g. Essential Next.js). Plugin authors (i.e. Netlify, in the case of Essential plugins) are in control of the format used in these functions, so it feels redundant to be detecting it by performing an expensive operation. If plugins want to generate JavaScript functions that use ESM, they can simply generate `.mjs` files (once https://github.com/netlify/zip-it-and-ship-it/pull/619 lands).

These functions are loaded from an internal functions directory (i.e. `.netlify/functions-internal`), which is one of the sources supplied to `zipFunctions`. We could look into doing the automatic detection only for user-defined functions.

ehmicky:
:100: on what @eduardoboucas is saying about the performance cost of parsing or lexing JavaScript for every build with functions.

I am wondering if a solution based on string search might be good enough for 99% of cases (while still imperfect)? Something like:
  - Read file (or only read first few lines if we are concerned about the time to read big files)
  - Look for the `import ` or `export ` string (including final space) at the beginning of the line. Using a string search instead of RegExps (for performance)

This does feel rough and improper compared to using a lexer. However, it might be much faster.

charliegroll:
@eduardoboucas @ehmicky perf is the concern I had in mind when I introduced https://github.com/guybedford/es-module-lexer in my draft PR, since it runs inline wasm and claims to be highly performant at just doing its one task. If adding _any_ checks is still a concern, though, I can shelf my POC work.

eduardoboucas:
Thanks for that additional context, @charliegroll. If that library brings considerable performance improvements over `acorn`, I'm actually interested in finding out whether we could use it _instead_ of `acorn` for the esbuild parsing.

[...]

PR #625, opened 2021-08-23
feat: use esbuild by default for es modules, if detected

charliegroll:
**- Summary**

It'd be great if we could detect ES Module functions and automagically use esbuild as the bundler, similar to the way Typescript detection works.

**- Test plan**

`TODO` (I've added a simple test for POC)

cat spinning

Closes https://github.com/netlify/zip-it-and-ship-it/issues/617


package.json
```diff
@@ -63,6 +63,7 @@
     "del": "^6.0.0",
     "elf-cam": "^0.1.1",
     "end-of-stream": "^1.4.4",
+    "es-module-lexer": "^0.7.1",
```
charliegroll, line 66: the node docs reference https://github.com/guybedford/cjs-module-lexer, which led me to this (https://github.com/guybedford/es-module-lexer)

src/runtimes/node/detect_es_module.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs')
+
+const { init, parse } = require('es-module-lexer')
+
+const detectEsModule = async ({ srcFile }) => {
+  // console.log('srcFile: ', srcFile)
```
charliegroll, line 6: not quite sure yet how to accurately supply this param until I figure out exactly what the params to `zipFunction` each mean

src/runtimes/node/detect_es_module.js
```diff
@@ -0,0 +1,22 @@
...
+  if (!srcFile) {
+    return false
+  }
+
+  await init
+
+  // Would love to just use await fs.promises.readFile, but it's not available in our version of Node.
+  // eslint-disable-next-line node/no-sync
+  const srcFileContents = fs.readFileSync(srcFile, { encoding: 'utf8' })
```
charliegroll, line 15: this could probably use some `try/catch` safety with a default to `false` to safely maintain the status quo

src/runtimes/node/detect_es_module.js
```diff
@@ -0,0 +1,22 @@
...
+
+const detectEsModule = async ({ srcFile }) => {
+  // console.log('srcFile: ', srcFile)
+  if (!srcFile) {
+    return false
+  }
+
+  await init
+
+  // Would love to just use await fs.promises.readFile, but it's not available in our version of Node.
```
ehmicky, line 13: `utils.promisify()` is available though.

src/runtimes/node/index.js
```diff
@@ -35,7 +50,8 @@ const zipFunction = async function ({
   srcPath,
   stat,
 }) {
-  const bundler = config.nodeBundler || getDefaultBundler({ extension })
+  // how do you construct the srcFile?
+  const bundler = config.nodeBundler || getDefaultBundler({ extension, srcFile: join(srcPath, filename) })
```
eduardoboucas, line 54: `getDefaultBundler` returns a Promise, so depending on which branch of we fall into, `bundler` will be either `string` or `Promise<string>`, which may be problematic. How do you feel about:

```js
const bundler = config.nodeBundler || (await getDefaultBundler({ extension, srcFile: join(srcPath, filename) }))
```

src/runtimes/node/index.js
```diff
@@ -4,14 +4,29 @@ const cpFile = require('cp-file')
...
 
 // We use ZISI as the default bundler until the next major release, with the
 // exception of TypeScript files, for which the only option is esbuild.
-const getDefaultBundler = ({ extension }) => (extension === '.ts' ? JS_BUNDLER_ESBUILD : JS_BUNDLER_ZISI)
+const getDefaultBundler = ({ extension, srcFile }) => {
+  if (extension === '.ts') {
+    return JS_BUNDLER_ESBUILD
+  }
+
+  if (srcFile) {
```
eduardoboucas, line 20: Do we need this check here, since `zipFunction` is the only function calling this and `join()` will never be falsy?

src/runtimes/node/detect_es_module.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs')
+
+const { init, parse } = require('es-module-lexer')
```
eduardoboucas, line 3: We already have a JavaScript parser ([`acorn`](https://www.npmjs.com/package/acorn)) as a dependency, which should be able to tell us whether a file uses ESM or not. If it is, I would try to use it before adding a new dependency.

src/runtimes/node/index.js
eduardoboucas: Also, I believe you can use `mainFile` instead of `join(srcPath, filename)`.

charliegroll: I had tried mainFile but I kept getting results where it looked like mainFile wasn't the actual file being investigated? is it possible `getDefaultBundler` will be called repeatedly for the same mainFile but with different dependencies its evaluating?

src/runtimes/node/detect_es_module.js
charliegroll: sure sure, I'll take a look 👍

ehmicky: I am wondering about the performance difference there considering:
  - `acorn` is implemented in JavaScript while `es-module-lexer` is in WebAssembly
  - `acorn` does lexing+parsing while `es-module-lexer` does lexing only
  - `acorn` parses all possible JavaScript syntax while `es-module-lexer` only looks for imports/exports

The author of `es-module-lexer` actually [claims the following](https://github.com/guybedford/es-module-lexer#es-module-lexer):

> For an example of the performance, Angular 1 (720KiB) is fully parsed in 5ms, in comparison to the fastest JS parser, Acorn which takes over 100ms.

tests/main.js
```diff
@@ -272,16 +272,14 @@ testBundlers('Can use dynamic import() with esbuild', [ESBUILD, ESBUILD_ZISI], a
   await zipNode(t, 'dynamic-import', { opts: { config: { '*': { nodeBundler: bundler } } } })
 })
 
-testBundlers('Bundling does not crash with dynamic import() with zisi', [DEFAULT], async (bundler, t) => {
-  await t.throwsAsync(zipNode(t, 'dynamic-import', { opts: { config: { '*': { nodeBundler: bundler } } } }), {
```
charliegroll, line 276: this is no longer a reachable flow, since we'll detect `import`s/`export`s and default to `esbuild`

charliegroll:
I ran some preliminary benchmarks using the existing fixtures and got the following: 

```sh
# run.sh
#!/usr/bin/env bash
exec 3>>.delta.largeDepsZisi
exec 4>>.delta.largeDepsEsbuild
exec 5>>.delta.largeDepsZisi.withFlag
exec 6>>.delta.largeDepsEsbuild.withFlag

npm ci --prefix benchmarks/fixtures
node benchmarks/zisi.js >&3
node benchmarks/esbuild.js >&4
NETLIFY_EXPERIMENTAL_DEFAULT_ES_MODULES_TO_ES_BUILD=true node benchmarks/zisi.js >&5
NETLIFY_EXPERIMENTAL_DEFAULT_ES_MODULES_TO_ES_BUILD=true node benchmarks/esbuild.js >&6
```

image

src/runtimes/node/detect_es_module.js
```diff
@@ -0,0 +1,25 @@
...
+
+const detectEsModule = async ({ mainFile }) => {
+  if (!mainFile) {
+    return false
+  }
+
+  try {
+    const [mainFileContents] = await Promise.all([pReadFile(mainFile, { encoding: 'utf8' }), init])
```
eduardoboucas, line 14: Tiny detail, just for consistency with the rest of the codebase:

```suggestion
    const [mainFileContents] = await Promise.all([pReadFile(mainFile, 'utf8'), init])
```

src/runtimes/node/detect_es_module.js
```diff
@@ -0,0 +1,25 @@
...
+
+    const [imports, exports] = parse(mainFileContents)
+
+    return imports.length !== 0 || exports.length !== 0
+  } catch {
+    // if there are any problems with init or parsing, assume it's not an ES module
```
eduardoboucas, line 20: 🤓 

```suggestion
    // If there are any problems with init or parsing, assume it's not an ES module
```

src/runtimes/node/index.js
```diff
@@ -4,20 +4,34 @@ const cpFile = require('cp-file')
...
 // We use ZISI as the default bundler, except for certain extensions, for which
 // esbuild is the only option.
-const getDefaultBundler = ({ extension }) =>
-  ['.mjs', '.ts'].includes(extension) ? JS_BUNDLER_ESBUILD : JS_BUNDLER_ZISI
+const getDefaultBundler = async ({ extension, mainFile, featureFlags }) => {
+  if (['.mjs', '.ts'].includes(extension)) {
+    return JS_BUNDLER_ESBUILD
+  }
+
+  if (featureFlags && featureFlags.defaultEsModulesToEsBuild) {
```
eduardoboucas, line 20:
```suggestion
  if (featureFlags && featureFlags.defaultEsModulesToEsbuild) {
```

tests/main.js
```diff
@@ -272,16 +272,34 @@ testBundlers('Can use dynamic import() with esbuild', [ESBUILD, ESBUILD_ZISI], a
...
+    await zipNode(t, 'local-require-esm', {
+      length: 3,
+      opts: { featureFlags: { defaultEsModulesToEsBuild: true }, config: { '*': { nodeBundler: bundler } } },
+    })
+  },
+)
+
+testBundlers(
+  'Can bundle functions with `.js` extension using ES Modules and feature flag OFF',
+  [DEFAULT],
```
eduardoboucas, line 292: There's no need to use `testBundlers` if we're running the test with a single bundler. In this case, I would either use a plain `test()` or, even better, test with all bundlers and check that the operation succeeds or fails based on the value of `bundler`. Does that make sense?

src/runtimes/node/index.js
```diff
@@ -4,20 +4,34 @@ const cpFile = require('cp-file')
...
 const { findFunctionsInPaths } = require('./finder')
 const { getSrcFiles } = require('./src_files')
 const { zipEsbuild } = require('./zip_esbuild')
 const { zipZisi } = require('./zip_zisi')
 
 // We use ZISI as the default bundler, except for certain extensions, for which
 // esbuild is the only option.
-const getDefaultBundler = ({ extension }) =>
-  ['.mjs', '.ts'].includes(extension) ? JS_BUNDLER_ESBUILD : JS_BUNDLER_ZISI
+const getDefaultBundler = async ({ extension, mainFile, featureFlags }) => {
```
eduardoboucas, line 15: [dust] If we add a default value to `featureFlags` we can probably lose the `&&` in line 20. What do you think?

```suggestion
const getDefaultBundler = async ({ extension, mainFile, featureFlags = {} }) => {
```

eduardoboucas (approved):
This looks great, thanks @charliegroll! I've added super minor comments.

PR #654, opened 2021-09-13
chore: pin version of adm-zip

eduardoboucas:
**- Summary**

Version 0.5.6 of `adm-zip` causes tests to fail (see https://github.com/netlify/zip-it-and-ship-it/pull/651/checks?check_run_id=3587767454). This locks the version to <=0.5.5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
   },
   "devDependencies": {
     "@netlify/eslint-config-node": "^3.2.6",
-    "adm-zip": "^0.5.0",
+    "adm-zip": "0.5.5",
     "ava": "^3.0.0",
     "cpy": "^8.0.0",
     "get-stream": "^6.0.0",
--- a/renovate.json5
+++ b/renovate.json5
@@ -13,6 +13,12 @@
       matchPackageNames: ['yargs'],
       allowedVersions: '<17',
     },
+    // Version 0.5.6 of adm-zip is breaking tests. See:
+    // https://github.com/netlify/zip-it-and-ship-it/pull/651/checks?check_run_id=3587767454
+    {
+      matchPackageNames: ['adm-zip'],
+      allowedVersions: '<=0.5.5',
+    },
     {
       // Fake dependencies used in tests
       packageNames: ['test'],


ehmicky (commented):
Should the version be pinned in `package.json` too?

erezrokah (approved):

eduardoboucas:
> Should the version be pinned in `package.json` too?

That's what I started doing first, but then @erezrokah mentioned the Renovate file instead. It feels redundant to do both — i.e. if it's pinned in `package.json`, Renovate wouldn't try to upgrade it anyway?

ehmicky:
When we run `npm install` locally, we would get `0.5.6`, making local tests fail, unless it is pinned in `package.json`.

ehmicky (approved):

eduardoboucas:
Those are great points. Also, every pinned version in the Renovate file is also reflected in `package.json`, so not doing it for this would be inconsistent.

Done in https://github.com/netlify/zip-it-and-ship-it/pull/654/commits/e068dbf148950593100c10aed09726b197a64bf3.

Thanks @ehmicky! 🙌🏻

erezrokah:
> That's what I started doing first, but then @erezrokah mentioned the Renovate file instead. It feels redundant to do both — i.e. if it's pinned in `package.json`, Renovate wouldn't try to upgrade it anyway?

I think we need both:
1. So users will get the pinned versioned
2. So renovate will stop creating PRs for the broken version

If you pin a dependency in `package.json` I believe renovate will still create a PR to upgrade it by default, but will create a pinned update, e.g. from `"adm-zip": "0.5.5"` to `"adm-zip": "0.5.6"`. If you have a range, e.g. `"adm-zip": "^0.5.5"` renovate will create a PR to update only the lock file such as https://github.com/netlify/zip-it-and-ship-it/pull/649/files, unless the update requires a range update (like a major one), then it will update `package.json` too.

> FYI, for Netlify packages we use a different update strategy where we always bump both the lock file and the `package.json` file https://github.com/netlify/zip-it-and-ship-it/pull/653/files

More [here](https://docs.renovatebot.com/configuration-options/#rangestrategy)

PR #655, opened 2021-09-13
chore(ci): fix our npm caching strategy for our e2e nextjs tests

JGAntunes:
**- Summary**

Similar to - https://github.com/netlify/cli/pull/3277 - for our e2e tests seems like we clone a total of 3 projects. We need to make use of the `cache-dependency-path`property to point to each of the lockfiles individually (see [here](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies)).

**- Test plan**

- https://github.com/netlify/zip-it-and-ship-it/runs/3588927992?check_suite_focus=true

wombat-chore

### DIFF
--- a/.github/workflows/e2e-nextjs.yml
+++ b/.github/workflows/e2e-nextjs.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: [main, 'chore/fix-ci-npm-cache']
     # Run on release PRs only, or changes to this same file
-    paths: ['CHANGELOG.md', '.github/workflows-e2e-nextjs.yml']
+    paths: ['CHANGELOG.md', '.github/workflows/e2e-nextjs.yml']
 
 jobs:
   integration:
--- a/.github/workflows/e2e-nextjs.yml
+++ b/.github/workflows/e2e-nextjs.yml
@@ -2,7 +2,7 @@ name: Next.js Integration test
 
 on:
   pull_request:
-    branches: [main, 'chore/fix-ci-npm-cache']
+    branches: [main, 'chore/ci-e2e-nextjs']
     # Run on release PRs only, or changes to this same file
     paths: ['CHANGELOG.md', '.github/workflows/e2e-nextjs.yml']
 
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-nextjs.yml
```diff
@@ -2,9 +2,9 @@ name: Next.js Integration test
 
 on:
   pull_request:
-    branches: [main]
-    # Run on release PRs only
-    paths: ['CHANGELOG.md']
+    branches: [main, 'chore/fix-ci-npm-cache']
```
eduardoboucas, line 5: Should we remove this change before merging?

.github/workflows/e2e-nextjs.yml
```diff
@@ -2,9 +2,9 @@ name: Next.js Integration test
...
+    # Run on release PRs only, or changes to this same file
+    paths: ['CHANGELOG.md', '.github/workflows/e2e-nextjs.yml']
```
eduardoboucas, line 7: Should we remove this change before merging?

eduardoboucas: Actually, it kind of makes sense to run the CI when the workflow file changes, so maybe we should leave it?

JGAntunes: If folks agree with it I would keep both these changes ☝️ similart to what we have with `fossa` -https://github.com/netlify/zip-it-and-ship-it/blob/main/.github/workflows/fossa.yml#L6-L7 - this way we have the ability to always test this during dev if needed be?

JGAntunes: We could probably change the branch name though, maybe to something like `chore/ci-e2e-nextjs`? What do you think @eduardoboucas?

.github/workflows/e2e-nextjs.yml
JGAntunes: We could maybe keep this but under a different name, what do you think? 👇 

```suggestion
    branches: [main, 'chore/ci-e2e-nextjs']
```

.github/workflows/e2e-nextjs.yml
eduardoboucas: Oh, I see. So you'll use that branch whenever you want to test the flow? Sounds good to me!

eduardoboucas (approved):

Issue #636, opened 2021-08-31
Error: Cannot find module '@prisma/client' in netlify functions when using prisma + next + rushjs + pnpm


abereghici:
**- Do you want to request a _feature_ or report a _bug_?**

🐞

**- What is the current behavior?**

In a rush monorepo with pnpm  `@prisma/client` and `.prisma` files aren't included in function node_modules and throw the error:

```
{
errorType: "Runtime.ImportModuleError",
errorMessage: "Error: Cannot find module '@prisma/client' Require stack: - /var/task/.netlify/functions-internal/next_api_views/next_api_views.js - /var/task/next_api_views.js - /var/runtime/UserFunction.js - /var/runtime/index.js",
trace: [
"Runtime.ImportModuleError: Error: Cannot find module '@prisma/client'",
"Require stack:",
"- /var/task/.netlify/functions-internal/next_api_views/next_api_views.js",
"- /var/task/next_api_views.js",
"- /var/runtime/UserFunction.js",
"- /var/runtime/index.js",
" at _loadUserApp (/var/runtime/UserFunction.js:100:13)",
" at Object.module.exports.load (/var/runtime/UserFunction.js:140:17)",
" at Object.<anonymous> (/var/runtime/index.js:43:30)",
" at Module._compile (internal/modules/cjs/loader.js:999:30)",
" at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)",
" at Module.load (internal/modules/cjs/loader.js:863:32)",
" at Function.Module._load (internal/modules/cjs/loader.js:708:14)",
" at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)",
" at internal/main/run_main_module.js:17:47"
]
}
```

**- If the current behavior is a bug, please provide the steps to reproduce.**

- Create a rush monorepo with pnpm as package manager.
- Add a nextjs app with prisma inside of monorepo
- Set next target to `experimental-serverless-trace`
- Run prisma generate && netlify-cli build && netlify-cli deploy
- Unzip a function that uses prisma from `.netlify/functions/` and see that `@prisma/client` and `.prisma` are not included in node_modules

If you change the package manager from pnpm to npm then is working as expected.

I created a repository where you can reproduce the issue: https://github.com/abereghici/prisma-next-netlify-rush 

**- What is the expected behavior?**
`@prisma/client` and `.prisma` files should be included in `node_modules` when using pnpm.

[...]

PR #657, opened 2021-09-16
feat: pnpm-style symlinking (#636)

netlify-team-account-1:
Contains a reproduction test for pnpm-style symlinking, reported non-working in #636.


eduardoboucas:
@netlify-team-account-1 what's the status of this one?

netlify-team-account-1:
This is the test that was originally added a while ago to reproduce #636 (the issue we paired on). We found that this reproduction doesn't correctly capture the real issue - this test may still be useful, so why not merge it.

eduardoboucas:
> This is the test that was originally added a while ago to reproduce #636 (the issue we paired on). We found that this reproduction doesn't correctly capture the real issue - but the test may still be useful to prevent future bugs, so why not merge it.

Sounds good!

tests/main.js
```diff
@@ -1976,3 +1976,14 @@ test('Creates a manifest file with the list of created functions if the `manifes
     t.is(fn.path, file.path)
   })
 })
+
+testBundlers('Correctly follows node_modules via symlink', [ESBUILD, ESBUILD_ZISI], async (bundler, t) => {
+  const { tmpDir } = await zipNode(t, 'node-module-symlinks', {
+    opts: { config: { '*': { nodeBundler: bundler } } },
+  })
+
+  // eslint-disable-next-line import/no-dynamic-require, node/global-require
```
eduardoboucas, line 1985: We'll need to remove these once we land https://github.com/netlify/zip-it-and-ship-it/pull/682.

eduardoboucas (approved):

tests/main.js
netlify-team-account-1: #682 is only about fixtures, not about test code. We could add `// eslint-disable-file import/no-dynamic-require, node/global-require` to the top of this file instead of line-by-line, though!

erlandsona:
Curious if this also solves for this issue with regular `npm`?

Skn0tt:
Not sure what issue you're referring to - if there's already an issue for it in this repository, could you link it? If not, feel free to open one :)

PR #672, opened 2021-09-24
feat: add `size` property to output

### LINKED ISSUE #620: Add `size` property to output of `zipFunction` and `zipFunctions`

A feature. 🚀 

zip-it-and-ship-it should return the size of the archives it creates, allowing consumers (e.g. Netlify Build) to warn users whenever the size of a function is above a pre-configured threshold. This is particularly relevant to flag problems with functions that exceed the 50MB limit of AWS Lambda.

eduardoboucas:
**- Summary**

Adds a `size` property to the output of `zipFunction` and `zipFunctions`, containing the number of bytes of the generated archive.

Closes #620.

**- Test plan**

Added new test.

Chinchillas-Great-Pets

### DIFF
--- a/src/zip.js
+++ b/src/zip.js
@@ -1,4 +1,4 @@
-const { resolve } = require('path')
+const { extname, resolve } = require('path')
 
 const makeDir = require('make-dir')
 const pMap = require('p-map')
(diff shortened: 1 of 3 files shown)


src/zip.js
```diff
@@ -8,15 +8,23 @@ const { createManifest } = require('./manifest')
...
+// Returns the input object with an additional `size` property containing the
+// size of the file at `path` when it is a ZIP archive.
+const addArchiveSize = async (result) => {
+  const { path } = result
+
+  if (extname(path) !== '.zip') {
+    return result
   }
+
+  const { size } = await lstat(path)
```
ehmicky, line 25: Don't we want to use `fs.stat()` instead of `fs.lstat()` here?
I'm not sure it can actually ever be a symlink there, but if it was, would we want the size of the symlinked path, or the symlinked content?

ehmicky (approved):

src/zip.js
eduardoboucas: It won't ever be a symlink in this case because we're opening the ZIP file we're creating ourselves, but I agree that there's no reason to use `lstat`. I replaced it with `stat` in https://github.com/netlify/zip-it-and-ship-it/pull/672/commits/6a75c748242bde193d473b2041c4c26fe07dca60.

PR #675, opened 2021-09-24
chore: add TypeScript build step

netlify-team-account-1:
**- Summary**

- sets up typescript
- configures `ava` to auto-compile typescript files during testing
- migrates the first files (as example)

**- Test plan**


tests/bin.ts
```diff
@@ -1,6 +1,6 @@
 const { join } = require('path')
 
-const test = require('ava')
+import test from "ava"
```
Skn0tt, line 3: In .ts-files, you need to use `import` to get typings.

tests/main.js
```diff
@@ -14,14 +14,14 @@ const sinon = require('sinon')
 const { dir: getTmpDir, tmpName } = require('tmp-promise')
 const unixify = require('unixify')
 
+const { zipFunction, listFunctions, listFunctionsFiles } = require('..')
 // We must require this file first because we need to stub it before the main
 // functions are required.
 // eslint-disable-next-line import/order
-const shellUtils = require('../src/utils/shell')
+const shellUtils = require('../dist/utils/shell')
```
Skn0tt, line 21: Mocks need to be made on the dist files.

tests/bin.ts
netlify-team-account-1, line 3: In .ts-files, you'll need to use import to get typings.

tests/main.js
netlify-team-account-1, line 21: Mocks need to be made on the dist files.

tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+
+    /* Interop Constraints */
+    // "isolatedModules": true,                          /* Ensure that each file can be safely transpiled without relying on other imports. */
+    // "allowSyntheticDefaultImports": true,             /* Allow 'import x from y' when a module doesn't have a default export. */
+    "esModuleInterop": true /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables `allowSyntheticDefaultImports` for type compatibility. */,
+    // "preserveSymlinks": true,                         /* Disable resolving symlinks to their realpath. This correlates to the same flag in node. */
+    "forceConsistentCasingInFileNames": true /* Ensure that casing is correct in imports. */,
+
+    /* Type Checking */
+    "strict": false /* Enable all strict type-checking options. */,
```
netlify-team-account-1, line 76: we'll start by disabling all of the typescript checks, and enable them step-by-step.

netlify-team-account-1:
There's this one test that asserts ESM-imports to fail with the default bundler: https://github.com/netlify/zip-it-and-ship-it/blob/setup-typescript/tests/main.js#L330-L345

This PR adds the `esm` helper for tests, which makes Node.js capable of understanding `import`. It's required to make typescript test files (like [tests/bin.ts](https://github.com/netlify/zip-it-and-ship-it/pull/675/files#diff-2b8a49e916fa4c1d3feee7d83f39c71278d9c98e8dcb1b88bbf2f528a1c0b6bc)) work. It breaks that test, though: https://github.com/netlify/zip-it-and-ship-it/runs/3732129960#step:6:370

What should we do? Remove the test?

.eslintrc.js
```diff
@@ -13,4 +13,5 @@ module.exports = {
       },
     },
   ],
+  ignorePatterns: ['README.md', 'tests/fixtures'],
```
eduardoboucas, line 16: Any reason to exclude `README.md` here? Linting the README is intentional (it's part of our [Node project template](https://github.com/netlify/node-template/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/package.json#L27)), and it allows us to not only have a consistent Markdown syntax, but also ensure that any code samples within it are also following our conventions.

package.json
```diff
@@ -2,19 +2,21 @@
...
   "description": "Zip it and ship it",
-  "main": "./src/main.js",
+  "main": "./dist/main.js",
   "bin": {
-    "zip-it-and-ship-it": "src/bin.js"
+    "zip-it-and-ship-it": "dist/bin.js"
   },
   "files": [
-    "src/**/*.js"
+    "dist"
```
eduardoboucas, line 10: Will `dist` only contain the transpiled JavaScript files? For example, if we have a non-TypeScript file (e.g. JSON) somewhere in `src`, will it end up in `dist` or only the files that `tsc` processes? I'm asking this because the current `files` glob purposely excludes any extraneous files from the npm package, and I'd like to keep that behaviour.

package.json
```diff
@@ -2,19 +2,21 @@
...
   ],
   "scripts": {
     "prepublishOnly": "npm ci && npm test",
+    "prepack": "npm run build",
+    "build": "tsc && chmod +x dist/bin.js",
```
eduardoboucas, line 15: Is there a cross-OS way to ensure the right permissions for the binary?

package.json
```diff
@@ -88,24 +90,36 @@
     "yargs": "^16.0.0"
   },
   "devDependencies": {
-    "@netlify/eslint-config-node": "^3.2.8",
+    "@ava/typescript": "^2.0.0",
+    "@netlify/eslint-config-node": "github:netlify/eslint-config-node#feat/typescript",
     "adm-zip": "0.5.5",
     "ava": "^3.0.0",
     "cpy": "^8.0.0",
+    "esm": "^3.2.25",
```
eduardoboucas, line 98: Can you clarify why this is needed?

tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    /* Projects */
+    "incremental": true /* Enable incremental compilation */,
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "es5" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
eduardoboucas, line 14: Any reason for us to be transpiling down to ES5? The minimum version supported by zip-it-and-ship-it is [10.18.0](https://github.com/netlify/zip-it-and-ship-it/blob/main/package.json#L102), so we should be fine with `"es6"`? On that note, our linter currently warns you whenever you're trying to use a feature that isn't supported by the version in `engines.node`, which is quite useful — what will be the state of that with the introduction of TypeScript and the build step? Will it be able to know what things will be transpiled _to_ and warn you if there's a compatibility issue?

eduardoboucas:
> There's this one test that asserts ESM-imports to fail with the default bundler: https://github.com/netlify/zip-it-and-ship-it/blob/setup-typescript/tests/main.js#L330-L345
> 
> This PR adds the `esm` helper for tests, which makes Node.js capable of understanding `import`. It's required to make typescript test files (like [tests/bin.ts](https://github.com/netlify/zip-it-and-ship-it/pull/675/files#diff-2b8a49e916fa4c1d3feee7d83f39c71278d9c98e8dcb1b88bbf2f528a1c0b6bc)) work. It breaks that test, though: https://github.com/netlify/zip-it-and-ship-it/runs/3732129960#step:6:370
> 
> What should we do? Remove the test?

I'd try to keep the test, if possible, but it's not the end of the world if we need to remove it. But more importantly, one thing I worry about the proposed setup is that we'll no longer be testing the code we'll actually ship, but the pre-transpilation version. Would testing the `dist/` files be a viable option?

package.json
netlify-team-account-1: `tsc` will replicate the structure in `src`, but only with regards to source files. AFAIK, a JSON file would be copied over, but only if it's imported from one of the TS files. But I'll verify that.

package.json
netlify-team-account-1: Adding that permission is only required for the `execa` call in `tests/bin.ts`. Now that I think of it, we should be able to remove the `chmod +x` call altogether if we run `execa("node dist/bin.js")` instead of `execa("dist/bin.js")`.

tsconfig.json
netlify-team-account-1: ES5 is the default, I didn't change anything about it. Will update to `es6` 👍

netlify-team-account-1:
> Would testing the dist/ files be a viable option?

The `@ava/typescript` plugin [rewrites](https://github.com/avajs/typescript#enabling-typescript-support) every import from `src/` to `dist/`, so we should already be testing the `dist/` files AFAICT. Will verify.

eduardoboucas:
> The `@ava/typescript` plugin [rewrites](https://github.com/avajs/typescript#enabling-typescript-support) every import from `src/` to `dist/`, so we should already be testing the `dist/` files AFAICT. Will verify.

Oh! So the test fails not because we're testing the source instead of the compiled files, but because the test runner injects into the runtime the ability to run ESM files?

netlify-team-account-1:
Correct! Couldn't have said it better :)

eduardoboucas:
> Correct!

Got it! I misunderstood.

package.json
netlify-team-account-1: https://github.com/netlify/zip-it-and-ship-it/pull/675#issuecomment-930057320

netlify-team-account-1: This is needed, because `@ava/typescript` doesn't transpile the "import {} from ..." calls in our test files. So we need to teach Node.js what to do with them, and `esm` does that.

package.json
netlify-team-account-1: If we enabled tsc's `resolveJsonModule` option, it would copy over directly-imported JSON files. Since we don't do that, it will not generate anything into `dist` except `.js`, `.d.ts` and `.js.map`.

package.json
```diff
@@ -88,24 +90,36 @@
     "yargs": "^16.0.0"
   },
   "devDependencies": {
-    "@netlify/eslint-config-node": "^3.2.8",
+    "@ava/typescript": "^2.0.0",
+    "@netlify/eslint-config-node": "github:netlify/eslint-config-node#feat/typescript",
```
eduardoboucas, line 94: The CI seems to be failing because of this. Would you like to cut a release candidate of `@netlify/eslint-config-node`?

Skn0tt: Yes, please!

netlify-team-account-1: Yes, that's a good idea!

.eslintrc.js
netlify-team-account-1: fixed with 1bfdfeb

package.json
netlify-team-account-1: esm and @ava/typescript Were removed in 1842943

package.json
netlify-team-account-1: 7f7b607

netlify-team-account-1:
Tests are green! yay 🥳

tests/bin.js
```diff
@@ -8,45 +8,47 @@ const { version } = require('../package.json')
 
 const { FIXTURES_DIR } = require('./helpers/main')
 
-const BINARY_PATH = `${__dirname}/../src/bin.js`
+const BINARY_PATH = join(__dirname, '..', 'dist', 'bin.js')
+
+const exec = (args, options) => execa('node', [BINARY_PATH, ...args], options)
```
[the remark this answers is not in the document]
eduardoboucas: We need the file to be executable. I don't think npm automatically sets execution permissions on files specified in `bin`, so presumably this works right now because we set those permissions on the file ourselves. If the file is going to be created as part of the build step, how do we handle that? 🤔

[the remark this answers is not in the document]
netlify-team-account-1: Hmm, I'd have assumed that specifying it under package.json's `bin` field is enough, and NPM will automatically make it executable on install. I'll cut a prerelease of this branch to verify that.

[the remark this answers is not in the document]
netlify-team-account-1:
```bash
❯ npm i --save-global @netlify/zip-it-and-ship-it@4.23.3-setup-typescript-1  
...
❯ zip-it-and-ship-it --version
4.23.3-setup-typescript-1
```

Works on my machine 😄

netlify-team-account-1:
pre-released to `@netlify/zip-it-and-ship-it-4.23.4-setup-typescript-2`

eduardoboucas:
> pre-released to `@netlify/zip-it-and-ship-it@4.23.4-setup-typescript-2`

We have a 3.8x increase in number of files and 2.1x increase in package size. I would expect an increase due to the `.d.ts` files, but this feels a bit too much? Can we omit the sourcemap files from the package?

netlify-team-account-1:
I published another prerelease to `@netlify/zip-it-and-ship-it@4.23.4-setup-typescript-omit-sourcemaps`, that has neither `.d.ts.map` nor `.js.map` files. That should be fine since our emitted `.js` files are somewhat readable. Can you check that release, too?

eduardoboucas:
Looks much better! Can we add `tsconfig.tsbuildinfo` to `.gitignore`? It's currently the largest file in the package.

netlify-team-account-1:
Yes, omitted `tsconfig.tsbuildinfo`. Sorry for not checking the file list myself, that could've saved us some back-and-forth.

eduardoboucas:
> Yes, omitted `tsconfig.tsbuildinfo`. Sorry for not checking the file list myself, that could've saved us some back-and-forth.

No problem!

I see that you chose to exclude it from the package rather than adding it to `.gitignore`. Is there any value in having this file in version control?

netlify-team-account-1:
`tsc` apparently outputs the `.tsbuildinfo` file both into the project root, and into the `dist/` directory. The project root one is gitignored, the dist directory one is excluded from NPM publishing.

eduardoboucas:
> `tsc` apparently outputs the `.tsbuildinfo` file both into the project root, and into the `dist/` directory.

Are you sure about that? I just ran `npm run build` and a single file was created, inside `dist/`, which is the expected behaviour according to [the docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html). I don't think the `/tsconfig.tsbuildinfo` entry in `.gitignore` is doing anything.

I think what's happening is that including `dist/` in `files` is overriding the `dist/` entry in `.gitignore`, so it'll include everything from `dist/`. I worry that in the future this starts including files we don't want to ship, without us realising.

We can probably change files to only includes `dist/**/*.js` and `dist/**/*.d.ts`. What do you think?

netlify-team-account-1:
Agreed! Implemented in 622d7f0. Running `npm pack` shows that we still include the same files.

eduardoboucas:
> Agreed! Implemented in [622d7f0](https://github.com/netlify/zip-it-and-ship-it/commit/622d7f0ed07fca9250d00dcc5ceae02958e7dea8). Running `npm pack` shows that we still include the same files.

Would you mind running another pre-release (hopefully the last one! 🤞🏻), just to be sure?

eduardoboucas (approved):
🎉

PR #681, opened 2021-09-27
fix: force esbuild target in zisi

eduardoboucas:
**- Summary**

When parsing files with esbuild to find dependencies, we rely on the default `target` property set by esbuild, which is `esnext`. However, this is overridden by any `compilerOptions.target` property set in a `tsconfig.json`. When this value is set to `es5`, parsing with esbuild can fail because some expressions can't be lowered to ES5.

This PR explicitly sets `target: "esnext"` in esbuild so that no transpiling is required.

**- Test plan**

Added a new test.

f03a349219e07c262f961a9afefb9a66

### DIFF
--- a/tests/fixtures/local-node-module-destructure-require/node_modules/@netlify/mock-package/stack.js
+++ b/tests/fixtures/local-node-module-destructure-require/node_modules/@netlify/mock-package/stack.js
@@ -1,0 +1,1 @@
+module.exports = 'jam'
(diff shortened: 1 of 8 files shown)


ehmicky (approved):

tests/fixtures/local-node-module-destructure-require/node_modules/@netlify/mock-package/stack.js
```diff
@@ -0,0 +1 @@
+module.exports = 'jam'
```
netlify-team-account-1, line 1: 🇯🇲

netlify-team-account-1 (approved):

PR #682, opened 2021-09-28
chore: exclude fixtures from eslint

netlify-team-account-1:
**- Summary**

All of our fixture files are littered with eslint-ignore directives, to the point where they're more noise than useful (it's fixtures, the whole point is that we can test against them, especially against non-optimal code!). This PR ignores the `tests/fixtures` for linting, and removes a plethora of eslint directives from fixture files.

**- Test plan**


.eslintrc.js
```diff
@@ -13,4 +13,5 @@ module.exports = {
       },
     },
   ],
+  ignorePatterns: ['tests/fixtures/**/*'],
```
eduardoboucas, line 16: Instead of adding an `ignorePatterns` rule, could we not remove `tests` from [here](https://github.com/netlify/zip-it-and-ship-it/blob/main/package.json#L28)?

netlify-team-account-1: I think we should keep actual test files linted, as tests are code and should be maintainable. The fixtures are more resources than code, which is why I think we shouldn't be linting them.

Curious, what's the reason we have eslint config in `package.json`? Instead of adding the ignore pattern there, we could also think about moving that config into `.eslintrc.js`.

eduardoboucas: > I think we should keep actual test files linted, as tests are code and should be maintainable. The fixtures are more resources than code, which is why I think we shouldn't be linting them.

Doh! That makes sense, ignore my comment.

> 
> Curious, what's the reason we have eslint config in `package.json`? Instead of adding the ignore pattern there, we could also think about moving that config into `.eslintrc.js`.

I'm not sure, to be honest. That's the pattern we use in our [Node project template](https://github.com/netlify/node-template/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/package.json#L26-L29). I'm not against changing it, but perhaps we should discuss that with the guild so that we use a unified approach.

eduardoboucas (approved):

eduardoboucas:
@netlify-team-account-1 btw the fixtures in `benchmarks/fixtures/` are not covered by our linter, so the comments you actually come from the tools that generated the functions. It shouldn't make a difference, but conceptually we should probably keep them because they would exist in a real scenario.

tests/fixtures/node-module-next10-critters-exact/function/renderNextPage.js
```diff
@@ -1,5 +1,4 @@
 /* eslint-disable unicorn/filename-case */
```
eduardoboucas, line 1: @netlify-team-account-1 I think you missed `/* eslint-disable` rules?

PR #695, opened 2021-10-04
chore: refactor `testMany` test util

eduardoboucas:
**- Summary**

The test suite uses a `testBundlers` helper that allows us to run a single test multiple times, one for each Node bundler. This is very useful, but having it constrained to variations in the bundler is very limiting, as we may wish to test other variations (e.g. with or without a feature flag).

This PR renames the `testBundlers` helper to `testMany` and introduces a new syntax.

We first define a matrix of all the possible variations when we instantiate the helper:

```js
const testMany = makeTestMany(test, {
  bundler_default: {
    nodeBundler: undefined,
  },
  bundler_esbuild: {
    nodeBundler: ESBUILD,
  },
})
```

And, for each test, we supply the set of keys to be used as variations:

```js
testMany('Zips Node.js function files', ['bundler_default', 'bundler_esbuild'], async (options, t) => {
  // ...
})
```

If we want to test a feature flag, we can just add a key (e.g. `bundler_esbuild_xyz_flag`) to the matrix and include that key in the tests we want to test against.

images


eduardoboucas:
Side note: in the future we can easily extend this to support a glob syntax, so if we wanted to run both `bundler_esbuild` and `bundler_default` we'd be able to do:

```js
testMany('Some test, ['bundler_*'], async (options, t) => {
  // ...
})
```

tests/main.js
```diff
@@ -70,24 +64,31 @@ test.afterEach(() => {
...
+  bundler_default: {
+    nodeBundler: undefined,
+  },
+  bundler_esbuild: {
+    nodeBundler: ESBUILD,
+  },
+})
 
-testBundlers('Zips Node.js function files', [ESBUILD, ESBUILD_ZISI, DEFAULT], async (bundler, t) => {
+testMany('Zips Node.js function files', ['bundler_default', 'bundler_esbuild'], async (options, t) => {
```
netlify-team-account-1, line 76: We're replacing `[ESBUILD, ESBUILD_ZISI, DEFAULT]` (3 entries) with `['bundler_default', 'bundler_esbuild']` (2 entries, `ESBUILD_ZISI` missing). It looks like we're effectively losing a third of our tests - am I missing smth?

netlify-team-account-1 (commented):
Looks mostly good to me. Due to repetitiveness and size, I skimmed through the changes.

CodeCov reports lost coverage for a couple of lines: https://codecov.io/gh/netlify/zip-it-and-ship-it/compare/1efb897295154fb01c54b449d6299c1938e8b237...d91a6346dca38a21d6253795192b15c78a9b1350/changes in `runtimes/node/index.js`, do you have an idea what this may be caused by @eduardoboucas?

tests/main.js
eduardoboucas: Great catch, thanks! Added back in https://github.com/netlify/zip-it-and-ship-it/pull/695/commits/0b187129cab43be9bf80ff1c10642a4e549bd0f4.

netlify-team-account-1 (approved):
Coverage is at +0.00% again 👍 Nice work @eduardoboucas! 🥳

PR #698, opened 2021-10-04
chore: add test for `node:` prefix

eduardoboucas:
Adds a test asserting that we can handle imports with the `node:` prefix.


netlify-team-account-1 (approved):

netlify-team-account-1 (commented):
Taking back my approval here - are we sure that we don't need to change `package.json`? If this package is installed by netlify/build, it won't take the `package-lock` into account, and potentially install precinct@8.0 or precinct@8.1.

PR #700, opened 2021-10-04
fix: set `customErrorInfo` on zisi/esbuild errors

eduardoboucas:
**- Summary**

Adds a `customErrorInfo` property to errors arising from esbuild in ZISI.

### DIFF
--- a/src/runtimes/node/bundler.js
+++ b/src/runtimes/node/bundler.js
@@ -3,7 +3,7 @@ const { basename, dirname, extname, resolve, join } = require('path')
 const esbuild = require('@netlify/esbuild')
 const { tmpName } = require('tmp-promise')
 
-const { RUNTIME_JS } = require('../../utils/consts')
+const { JS_BUNDLER_ESBUILD, RUNTIME_JS } = require('../../utils/consts')
 const { getPathWithExtension, safeUnlink } = require('../../utils/fs')
 
 const { getBundlerTarget } = require('./bundler_target')
@@ -106,7 +106,10 @@ const bundleJsFile = async function ({
       warnings,
     }
   } catch (error) {
-    error.customErrorInfo = { type: 'functionsBundling', location: { functionName: name, runtime: RUNTIME_JS } }
+    error.customErrorInfo = {
+      type: 'functionsBundling',
+      location: { bundler: JS_BUNDLER_ESBUILD, functionName: name, runtime: RUNTIME_JS },
+    }
 
     throw error
   }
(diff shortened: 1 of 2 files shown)


src/runtimes/node/bundler.js
```diff
@@ -106,7 +106,10 @@ const bundleJsFile = async function ({
       warnings,
     }
   } catch (error) {
-    error.customErrorInfo = { type: 'functionsBundling', location: { functionName: name, runtime: RUNTIME_JS } }
+    error.customErrorInfo = {
+      type: 'functionsBundling',
+      location: { bundler: JS_BUNDLER_ESBUILD, functionName: name, runtime: RUNTIME_JS },
```
ehmicky, line 111: Do we want to show the bundler and/or the runtime in the error message title?

https://github.com/netlify/build/blob/a6ab7b9dbfebd21bcedcb5dbada4d1620cd6192a/packages/build/src/error/type.js#L86

Please note we do show the `location` fields in the error details, which is printed in the build logs. Also, the `location` fields are always sent to Bugsnag. Just wondering what you think about the error message title itself (which is shown in the error header and in the deploy summary).

ehmicky (approved):

src/runtimes/node/bundler.js
eduardoboucas: I personally don't think we need to show the runtime in the message because it probably offers less clarity than the noise it introduces. I think that field is useful mostly for Bugsnag. 

However, on that link I do see that we show `functionName`, which means that the field is mandatory. We don't have the function name when using esbuild from the ZISI flow. Adding it will mean passing that parameter around in a lot of functions, which I'm not super excited about, but it'll have to be done. 

I'll add to this PR.

ehmicky: Makes sense! :+1:

eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/700/commits/94c32481a17aaa73083b65a8691be483d0cef689. Some of these functions are called recursively, and without types it's a bit difficult to ensure all cases are covered, but I think I've done it!

ehmicky: Looks good!

ehmicky (approved):

src/runtimes/node/bundler.js
eduardoboucas: Thanks for your help! 🙌🏻

PR #702, opened 2021-10-05
chore: allow testMany args to accept glob patterns

netlify-team-account-1:
**- Summary**

This PR allows `testMany` variations to be passed in using patterns, and also adds a fancy recursive typescript type to give nice autocomplete to it.

**- Test plan**

- clone locally, try out autocomplete
- check if coverage stays the same

### DIFF
--- a/tests/helpers/globify.ts
+++ b/tests/helpers/globify.ts
@@ -1,0 +1,5 @@
+export type Globify<S extends string> = S extends `${infer Head}_${infer Tail}`
+  ? `${Head}*` | `${Head}_${Globify<Tail>}`
+  : S | '*'
+
+type Test = Globify<'foo_bar' | 'foo_bar_baz' | 'yeet'>
--- a/tests/helpers/globify.ts
+++ b/tests/helpers/globify.ts
@@ -1,5 +1,7 @@
 export type Globify<S extends string> = S extends `${infer Head}_${infer Tail}`
-  ? `${Head}*` | `${Head}_${Globify<Tail>}`
+  ? '*' | `${Head}*` | `${Head}_${Globify<Tail>}`
   : S | '*'
 
 type Test = Globify<'foo_bar' | 'foo_bar_baz' | 'yeet'>
+
+type Test2 = Globify<'bundler_default' | 'bundler_default_parse_esbuild' | 'bundler_esbuild' | 'bundler_esbuild_zisi'>
--- a/tests/helpers/test_many.js
+++ b/tests/helpers/test_many.js
@@ -1,5 +1,11 @@
 const { platform } = require('process')
 
+/**
+ * @template M, O
+ * @param {import("ava")} test
+ * @param {Record<M, O>} matrix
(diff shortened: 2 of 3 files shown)


netlify-team-account-1:
another demo:

https://user-images.githubusercontent.com/90322326/136005649-5c923ad1-bd77-4146-8866-a24a2e388fc2.mov

tests/helpers/globify.ts
```diff
@@ -0,0 +1,7 @@
+export type Globify<S extends string> = S extends `${infer Head}_${infer Tail}`
+  ? '*' | `${Head}*` | `${Head}_${Globify<Tail>}`
+  : S | '*'
+
+type Test = Globify<'foo_bar' | 'foo_bar_baz' | 'yeet'>
```
eduardoboucas, line 5: Do we need these?

netlify-team-account-1: not neccessarily - (I used them for verifying that the type works as expected) - but they don't really hurt either, as they'll never be shipped and are used only be the editor

tests/helpers/globify.ts
```diff
@@ -0,0 +1,7 @@
+export type Globify<S extends string> = S extends `${infer Head}_${infer Tail}`
```
eduardoboucas, line 1: Could you add a short comment explaining what this expression is doing, since it's not particularly easy to parse?

tests/helpers/globify.ts
eduardoboucas: Sure, but why keep them if they're not doing anything?

tests/helpers/globify.ts
netlify-team-account-1:
```suggestion
// takes a string like `foo_bar`, recursively splits it up along underscores, and collects all kinds of glob expressions.
// for `foo_bar`, it'd evaluate to `'*' | 'foo*' | 'foo_*' | 'foo_bar'`.
export type Globify<S extends string> = S extends `${infer Head}_${infer Tail}`
```

tests/helpers/globify.ts
```diff
@@ -0,0 +1,9 @@
+// takes a string like `foo_bar`, recursively splits it up along underscores, and collects all kinds of glob expressions.
```
eduardoboucas, line 1:
```suggestion
// Takes a string like `foo_bar`, recursively splits it up along underscores, and collects all kinds of glob expressions.
```

🤓 

(Comment is super helpful, thanks!)

tests/helpers/globify.ts
netlify-team-account-1: good point. I've transformed them into type-level tests: https://github.com/netlify/zip-it-and-ship-it/pull/702/files/7f867da77363718059c2d449e91b8750f9926b86..3ef4abe7fa6e91e9f1657b4e18187ade118971f5

so now they work as documentation & tests :)
(since we haven't integrated our test suite with TypeScript yet, this won't do anything apart from being squiggly in the editor - but that's a start ^^)

netlify-team-account-1:
this will likely need to be updated to support the `todo:` syntax introduced in #713

PR #736, opened 2021-10-11
chore: dont upgrade hand-built fixture modules

netlify-team-account-1:
**- Summary**

excludes tests/fixtures/node_modules from renovate upgrades, since these are hand-crafted mocks and only meant for tests.

cc #733

**- Test plan**

### DIFF
--- a/src/runtimes/rust/builder.js
+++ b/src/runtimes/rust/builder.js
@@ -18,7 +18,7 @@ const build = async ({ config, name, srcDir }) => {
   const functionName = basename(srcDir)
 
   try {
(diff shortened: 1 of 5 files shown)


netlify-team-account-1:
also thought about that. My initial thought was: fixtures are supposed to be *fix* - we don't need to keep them up-to-date, there's neither a security risk nor a maintenance risk connected with it.
OTOH, having Renovate keep those up-to-date is a very cheap way of making sure that ZISI keeps working with newer versions of the dependencies we test against. Makes me think we should keep it that way.

PR #740, opened 2021-10-12
feat!: remove `includedFilesBasePath` config property

eduardoboucas:
**- Summary**

The `includedFilesBasePath` config property was introduced in #441, when we added support for the `includedFiles` property.

A `basePath` property was then introduced in #474, making `includedFilesBasePath` redundant. This PR removes the latter in favour of the former.

This is technically a breaking change.

**- Test plan**

Adjusted existing test.

depositphotos_12883889-stock-photo-wombat-close-up


eduardoboucas:
> > this is technically a breaking change
> 
> are we planning on doing a major bump? or do we expect people to read the patch notes carefully?

We'll do a major bump. That is handled automatically by [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) (i.e. `feat!:` vs `feat:`).

eduardoboucas:
RIP

PR #746, opened 2021-10-14
fix: upgrade `precinct`

### LINKED ISSUE #743: `Cannot find module` error for Node-provided package
**- Do you want to request a _feature_ or report a _bug_?** bug

**- What is the current behavior?** build fails with error `Cannot find module 'stream/package.json'`

**- If the current behavior is a bug, please provide the steps to reproduce.**

- Create a new SvelteKit project with `npm init svelte@next`
- Install the Netlify adapter: https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify#installation
- Create an `.nvmrc` with the contents `16` to use Node 16
- Deploy to Netlify

There should be no need to install a `stream` package because it is provided by Node: https://nodejs.org/api/stream.html#stream_stream

I'm not sure why it's trying to find `stream/package.json` and failing since this is a built-in package

**- Please mention your node.js, and operating system version.**

Node 16. Linux

netlify-team-account-1:
**- Summary**

closes #743 

**- Test plan**


benmccann:
Hmm, I'm not sure if that's generally true. SvelteKit bundles everything with Vite first. It then uses ESBuild for a final bundling

Is there a reason we can't add an option to continue if the package cannot be found? Or alternatively, or in addition, we could assume packages with a `node:` prefix are provided by Node

netlify-team-account-1:
I opened a [PR](https://github.com/dependents/node-precinct/pull/88) on the dependency that does the module detection, to add `node:` as an indicator for built-in packages. Let's see what they say!

benmccann:
Thanks for the fix here! What's needed to get https://github.com/netlify/cli/pull/3533 merged? I notice that there's currently a dozen different PRs open to upgrade different dependencies and am not sure if there's some process you have for batching them all or something along those lines?

PR #759, opened 2021-10-21
feat: add ESM support when bundling with NFT

eduardoboucas:
**- Summary**

Adds support for ESM functions when bundling with NFT. We look for ESM files in the NFT cache and, for each one that is found, we use esbuild to transpile them down to CJS. We then replace the original files with the transpiled ones in the archive.

**- Test plan**

Included the NFT variation in 3 tests.

0cd8d5da58bb8d2f482eec89f968ebe6


eduardoboucas:
> Once ESM files are transpiled, what happens to them? I understand that each ESM file gets a CJS equivalent at some `tmp` destination. I'd assume that these `tmp` files are used to replace the ESM ones at some point, but I can't find where that happens 🤔 

You're absolutely correct. I think the part you're missing happens in a `Map` called `aliases` (which, as most things, could have a better name), which maps paths that should be in the archive to the paths they should take inside the zip.

_Example:_

```
{
  "/tmp/function-1-transpiled.js" => "/netlify/functions/function-1.js"
}
```

With this `aliases` object, when we find `/tmp/function-1-transpiled.js` in the list of files to include, we rename it to `/netlify/functions/function-1.js`, which is the expected path of the original, untranspiled file.

Each bundler is responsible for creating this `aliases` file, which is then used at the archive creation time.

Does that help?

PR #783, opened 2021-11-02
fix: modify `type` field of transpiled ESM packages

eduardoboucas:
**- Summary**

When transpiling ESM packages to CJS, we should also modify their `package.json` so that the `type` field is set to `commonjs`, not `module`.

**- Test plan**

Existing tests reused.

7f175148-2ad8-4170-a34f-6356948a6870-medium16x9_feat_6cd9db36ca604a249b8949fdb2de2df8

### DIFF
--- a/src/runtimes/node/bundlers/index.ts
+++ b/src/runtimes/node/bundlers/index.ts
@@ -1,3 +1,5 @@
+import type { Buffer } from 'buffer'
+
 import type { Message } from '@netlify/esbuild'
 
 import type { NodeBundlerName } from '..'
@@ -26,7 +28,25 @@ type BundleFunction = (
     repositoryRoot?: string
   } & FunctionSource,
 ) => Promise<{
+  // Aliases are used to change the path that a file should take inside the
+  // generated archive. For example:
+  //
+  // "/my-transpiled-function.js" => "/my-function.js"
+  //
+  // When "/my-transpiled-function.js" is found in the list of files, it will
+  // be added to the archive with the "/my-function.js" path.
   aliases?: Map<string, string>
+
+  // Rewrites are used to change the source file associated with a given path.
+  // For example:
+  //
+  // "/my-function.js" => "console.log(`Hello!`)"
+  //
+  // When "/my-function.js" is found in the list of files, it will be added to
+  // the archive with "console.log(`Hello!`)" as its source, replacing whatever
+  // the file at "/my-function.js" contains.
+  rewrites?: Map<string, string>
+
   basePath: string
   bundlerWarnings?: BundlerWarning[]
   cleanupFunction?: CleanupFunction
--- a/src/runtimes/node/bundlers/index.ts
+++ b/src/runtimes/node/bundlers/index.ts
@@ -1,5 +1,3 @@
-import type { Buffer } from 'buffer'
-
 import type { Message } from '@netlify/esbuild'
 
 import type { NodeBundlerName } from '..'
--- a/src/runtimes/node/bundlers/nft/index.ts
+++ b/src/runtimes/node/bundlers/nft/index.ts
@@ -7,20 +7,16 @@ import unixify from 'unixify'
 
 import type { BundleFunction } from '..'
 import type { FunctionConfig } from '../../../../config'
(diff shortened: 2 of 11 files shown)


src/runtimes/node/bundlers/index.ts
```diff
@@ -26,7 +26,25 @@ type BundleFunction = (
...
   } & FunctionSource,
 ) => Promise<{
+  // Aliases are used to change the path that a file should take inside the
+  // generated archive. For example:
+  //
+  // "/my-transpiled-function.js" => "/my-function.js"
+  //
+  // When "/my-transpiled-function.js" is found in the list of files, it will
+  // be added to the archive with the "/my-function.js" path.
   aliases?: Map<string, string>
```
eduardoboucas, line 36: `aliases` existed before, I've just added a comment to clarify what it's for.

src/runtimes/node/bundlers/index.ts
```diff
@@ -26,7 +26,25 @@ type BundleFunction = (
...
+
+  // Rewrites are used to change the source file associated with a given path.
+  // For example:
+  //
+  // "/my-function.js" => "console.log(`Hello!`)"
+  //
+  // When "/my-function.js" is found in the list of files, it will be added to
+  // the archive with "console.log(`Hello!`)" as its source, replacing whatever
+  // the file at "/my-function.js" contains.
+  rewrites?: Map<string, string>
```
eduardoboucas, line 46: `rewrites` is a new concept.

src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -81,10 +72,12 @@ const traceFilesAndTranspile = async function ({
   pluginsModulesPath?: string
 }) {
   const fsCache: FsCache = {}
-  const cache: NftCache = {}
-  const { fileList: dependencyPaths } = await nodeFileTrace([mainFile], {
+  const {
+    fileList: dependencyPaths,
+    esmFileList,
```
eduardoboucas, line 77: We were previously relying on NFT's cache object to inspect which of the paths corresponded to ESM files, but I realised there's a `esmFileList` property that contains just that, so there's no need to muck about with the internal cache.

src/runtimes/node/bundlers/nft/transpile.ts
```diff
@@ -1,32 +1,23 @@
...
 const transpile = async (path: string, config: FunctionConfig) => {
-  const targetPath = await tmpName({ postfix: '.js' })
-  const cleanupFn = () => safeUnlink(targetPath)
-
   // The version of ECMAScript to use as the build target. This will determine
   // whether certain features are transpiled down or left untransformed.
   const nodeTarget = getBundlerTarget(config.nodeVersion)
-
-  await build({
+  const transpiled = await build({
```
eduardoboucas, line 10: We were piping the output of esbuild to a temporary file, but actually we can just get the result loaded into memory by setting `write: false` and avoid creating a file altogether.

src/runtimes/node/utils/zip.ts
```diff
@@ -191,14 +208,19 @@ const zipJsFile = function ({
...
   userNamespace: string
 }) {
-  const filename = aliases.get(srcFile) || srcFile
-  const normalizedFilename = normalizeFilePath({ commonPrefix, path: filename, pluginsModulesPath, userNamespace })
+  const destPath = aliases.get(srcFile) || srcFile
+  const normalizedDestPath = normalizeFilePath({ commonPrefix, path: destPath, pluginsModulesPath, userNamespace })
 
-  addZipFile(archive, srcFile, normalizedFilename, stat)
+  if (rewrites.has(srcFile)) {
+    addZipContent(archive, rewrites.get(srcFile) as string, normalizedDestPath)
```
eduardoboucas, line 220: Because of https://github.com/microsoft/TypeScript/issues/13086.

netlify-team-account-1 (approved):
:rocket:

Hrishikesh-K:
Hey team, I believe this was supposed to be related with this thread: https://answers.netlify.com/t/nextjs-page-with-server-side-logic-throws-error-cannot-find-module-os/46431/? Could someone confirm? Users are still reporting it's not working.

PR #792, opened 2021-11-05
fix: correctly zip symlinks

netlify-team-account-1:
**- Summary**

The PNPM test currently doesn't work with NFT. While looking into the issue, I found this to be caused by ZISI not correctly handling symbolic links. Instead of creating actual symlinks inside the generated ZIP archives, it wrote a link's contents into a *normal* file, breaking their usage as symlinks.

I'm having a light "how did this ever work?" moment, as it seems that our zipping logic has *never* taken symlinks into account correctly, also not with other bundlers. That seems highly unlikely to me, given that we've been able to deploy PNPM projects successfully. @eduardoboucas maybe you have some wise insights here? 

6b9eb01e5597e6bf714337bfbbd1f20fc7b8c56b fixes the observed defect, but also breaks some other tests apparently, which we'll have to look into.

The `adm-zip` tool we used for unzipping files inside the test suite doesn't support symlinks *at all*. [I sent them a reproduction test](https://github.com/cthackers/adm-zip/pull/401), and swapped `adm-zip` out with the `unzip` CLI in the meantime (6b9eb01e5597e6bf714337bfbbd1f20fc7b8c56b). This is likely unsupported on Windows, so we need to find another solution before merging.

**- Test plan**


netlify-team-account-1:
I think this PR should be fine to review locally, I just gotta figure out a way to unzip files cross-platform without adm-zip :/

netlify-team-account-1:
Alright, got it working except for two lonely failing tests on windows. Interestingly enough, running it locally also yields two failing tests on Windows, albeit two different ones and with different error codes 🤔

tests/main.js
```diff
@@ -2253,18 +2253,24 @@ test('Creates a manifest file with the list of created functions if the `manifes
...
-    opts,
-  })
+const skipOnWindows = (name) => (platform === 'win32' ? `skip:${name}` : name)
 
-  const isEven = require(`${tmpDir}/function`)
-  t.is(isEven(2), '2 is even')
-})
+testMany(
+  'Correctly follows node_modules via symlink',
+  ['bundler_esbuild', skipOnWindows('bundler_nft')],
```
eduardoboucas, line 2260: [dust] We have a few places in the test where we change the behaviour based on `platform`, and we're using simple `if` statements. I would rather keep that simplicity than adding an abstraction on top of it, but if we do insist on adding an abstraction, we should probably change the existing tests to use it.

eduardoboucas (approved):
LGTM!

tests/main.js
netlify-team-account-1: our existing usage of if statements skips entire `testMany` blocks, while here I want to only disable one variation. `if` statements would become a bit unwieldly here. I'll see what I can do to clarify things.

eduardoboucas: Tests handlers receive the `variation` parameter. You could skip the test `if (variation === 'bundler_nft' && platform === 'win32')`?

eduardoboucas (dismissed):

tests/main.js
netlify-team-account-1: know what? scrap it. this is an escape hatch, we neither expect nor encourage to use this often. might as well make hard to type. 1f9b1054a85c7cd49e98dbf064a67f494d8c590a

eduardoboucas (approved):

eduardoboucas (approved):

PR #802, opened 2021-11-11
fix: esm imports with node: prefix on esbuild

netlify-team-account-1:
**- Summary**

fixes a bundling error that occurs when esbuild bundles a `node:`-namespaced  esm import.

cc https://github.com/nuxt/framework/issues/1826

**- Test plan**

### DIFF
--- a/src/runtimes/node/bundlers/esbuild/bundler.ts
+++ b/src/runtimes/node/bundlers/esbuild/bundler.ts
@@ -11,6 +11,7 @@ import type { RuntimeName } from '../../../runtime'
 import { getBundlerTarget } from './bundler_target'
 import { getDynamicImportsPlugin } from './plugin_dynamic_imports'
 import { getNativeModulesPlugin } from './plugin_native_modules'
+import { getNodeBuiltinPlugin } from './plugin_node_builtin'
 
 // Maximum number of log messages that an esbuild instance will produce. This
 // limit is important to avoid out-of-memory errors due to too much data being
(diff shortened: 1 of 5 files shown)


src/runtimes/node/bundlers/esbuild/plugin_native_modules.ts
```diff
@@ -39,6 +39,10 @@ const getNativeModulesPlugin = (externalizedModules: NativeNodeModules): Plugin
 
       if (!pkg) return
 
+      if (pkg[1].startsWith('node:')) {
```
eduardoboucas, line 42: Note that "native modules" in this context means Node modules with native binaries, whereas the `node:` prefix denotes built-in Node modules.

By having these together, I worry that making changes to the native module detection might break the built-in module handling in the future.

Have you considered creating a separate plugin for this? Something like:

```js
build.onResolve({ filter: /^node: }, async (args) => {
  return { external: true }
})
```

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -71,6 +72,7 @@ const bundleJsFile = async function ({
       processImports: config.processDynamicNodeImports !== false,
       srcDir,
     }),
+    getNodeBuiltinPlugin(),
```
eduardoboucas, line 75: Might be a micro-optimisation, but I think we'd benefit from running this plugin first, since marking a path as external will avoid running the other two (more expensive) plugins.

eduardoboucas (commented):
Looks great! Added just a tiny comment.

eduardoboucas (approved):
🎸

Hrishikesh-K:
Hey team, is this supposed to be working fine in Netlify CLI?

I was testing a repo and got the exact same error in CLI, except the user is using Vite and not Nuxt.js.

netlify-team-account-1:
It's indeed a ZISI problem, which uses esbuild internally. It's expected that this error occurs in the CLI until we ship a new version of that.

PR #806, opened 2021-11-12
fix: compile away `node:`-namespaced imports

netlify-team-account-1:
**- Summary**

AWS Lambda currently ships v14.17.4. Some libraries start using `require("node:...")`, which was only added in v14.18. This PR updates our esbuild bundler, so that it "compiles away" the `node:` prefix. The legacy / NFT logic behaviour is untouched.

cc https://github.com/netlify/pod-serverless/issues/57

cc https://github.com/nuxt/framework/issues/1826

**- Test plan**


eduardoboucas (approved):
LGTM. Can you change the description so that this doesn't close https://github.com/netlify/pod-serverless/issues/57? The issue is still there when not using esbuild.

eduardoboucas (approved):

PR #807, opened 2021-11-12
fix: only transpile ESM imports when entrypoint is ESM

eduardoboucas:
**- Summary**

The `shouldTranspile` method was previously returning `true` for any path that doesn't have a parent, where in reality it must do so only if the path itself is an ESM file.

**- Test plan**

A new test was added to verify this scenario.

sloth-1-102913


tests/fixtures/node-cjs-importing-mjs/lib/file.mjs
```diff
@@ -0,0 +1,10 @@
+class App {
+  constructor(event, context) {
+    return {
```
netlify-team-account-1, line 3: wait, what? Constructors can return something? this weird

netlify-team-account-1: but fine for a fixture, I guess ^^

netlify-team-account-1 (approved):

tests/fixtures/node-cjs-importing-mjs/lib/file.mjs
eduardoboucas: The idea was to replicate [this scenario](https://stackoverflow.com/questions/66676555/nodejs-14-x-native-aws-lambda-import-export-support/69136438#69136438), but using a class or not is actually irrelevant for the test case here. All that matters is that the file is ESM.

PR #813, opened 2021-11-15
chore: remove unneeded v16 test

netlify-team-account-1:
this test was added as a very specific reproduction case, but since
was superceded by two more general tests.
Removed in a follow-up to https://github.com/netlify/zip-it-and-ship-it/pull/806.

**- Summary**

**- Test plan**


eduardoboucas (approved):
🔪

PR #816, opened 2021-11-17
fix: ensure directories are created before writing files

eduardoboucas:
**- Summary**

When using `archiveFormat: "none"`, any files in `rewrites` are written to disk using `writeFile`. When doing this, we must ensure that the directory structure exists prior to the write operation. This PR addresses that.

**- Test plan**

Added a new test.

749dc9dfe2559f53aa1e466a9af3c609

### DIFF
--- a/src/utils/fs.ts
+++ b/src/utils/fs.ts
@@ -1,6 +1,8 @@
 import { lstat, readdir, readFile, stat, unlink, writeFile } from 'fs'
-import { format, join, parse, resolve } from 'path'
+import { dirname, format, join, parse, resolve } from 'path'
 import { promisify } from 'util'
+
+import makeDir from 'make-dir'
 
 import { nonNullable } from './non_nullable'
 
@@ -86,6 +88,16 @@ const resolveFunctionsDirectories = (input: string | string[]) => {
   return absoluteDirectories
 }
 
+const mkdirAndWriteFile: typeof pWriteFile = async (path, ...params) => {
+  if (typeof path === 'string') {
+    const directory = dirname(path)
+
+    await makeDir(directory)
+  }
+
+  return pWriteFile(path, ...params)
+}
+
 export {
   cachedLstat,
   cachedReaddir,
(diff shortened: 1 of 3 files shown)


src/utils/fs.ts
```diff
@@ -86,6 +88,16 @@ const resolveFunctionsDirectories = (input: string | string[]) => {
   return absoluteDirectories
 }
 
+const mkdirAndWriteFile: typeof pWriteFile = async (path, ...params) => {
+  if (typeof path === 'string') {
```
eduardoboucas, line 92: Checking that path is a string because it can also be a file descriptor.

ehmicky: If it is a file descriptor, shouldn't `pWriteFile()` be used instead of `mkdirAndWriteFile()` then?

ehmicky (approved):
:+1:

src/utils/fs.ts
eduardoboucas: Absolutely, but we still need to account for the possibility that someone decides to use `mkdirAndWriteFile` with a file descriptor. If nothing else, TypeScript forces us to have this check since we're setting `mkdirAndWriteFile` to have the same type as `pWriteFile`.

src/utils/fs.ts
```diff
@@ -86,6 +88,16 @@ const resolveFunctionsDirectories = (input: string | string[]) => {
...
+    const directory = dirname(path)
+
+    await makeDir(directory)
```
lukasholzer, line 95: why can we not use the native fs `mkdirSync(directory, { recursive: true })` with the recursive option as this was introduced in Node.js 10 and should be available no need for async and an own library like `make-dir`

src/utils/fs.ts
ehmicky: :+1:

src/utils/fs.ts
ehmicky: +1 to this :arrow_up:

PR #819, opened 2021-11-17
feat: add new top-level `schedule` property

eduardoboucas:
**- Summary**

Currently, any schedule data is surfaced in the result of `zipFunction` and `zipFunctions` as `config.schedule`. This PR adds a new top-level `schedule` property, which contains the same value.

This property will be useful when we introduce the in-code config declarations, because we'll be able to merge schedule data coming from the two types of sources and decide which one takes priority.

**- Test plan**

New test added.

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -36,12 +36,12 @@ const createManifest = async ({ functions, path }: { functions: FunctionResult[]
   await writeFile(path, JSON.stringify(payload))
 }
 
-const formatFunctionForManifest = ({ config, mainFile, name, path, runtime }: FunctionResult): ManifestFunction => ({
+const formatFunctionForManifest = ({ mainFile, name, path, runtime, schedule }: FunctionResult): ManifestFunction => ({
   mainFile,
   name,
   path: resolve(path),
   runtime,
-  schedule: config.schedule,
+  schedule,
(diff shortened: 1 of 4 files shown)


netlify-team-account-1 (approved):
Loving these super small PRs, really easy to review ^^

We already have a test for `schedule`: https://github.com/netlify/zip-it-and-ship-it/blob/main/tests/main.js#L2241
What does the new test give us? (not blocking obvsly, the more tests the merrier)

eduardoboucas:
> Loving these super small PRs, really easy to review ^^
> 
> We already have a test for `schedule`: https://github.com/netlify/zip-it-and-ship-it/blob/main/tests/main.js#L2241 What does the new test give us? (not blocking obvsly, the more tests the merrier)

The other test was primarily built to assert the creation of a manifest file. I wanted to have a test specific for schedules, which will be useful once we add the in-code annotations. We can either remove the schedule assertions from the test you linked, or we can keep them because it's fine to have that assertion duplicated. What would you prefer?

eduardoboucas:
@netlify-team-account-1 the review got stale when I fixed the merge conflict - could you ✅  again, please?

minivan (approved):
LGTM

PR #684, opened 2021-09-28
feat: add support for in-source config declarations

eduardoboucas:
**- Summary**

⏳ 

**- Test plan**

Added new tests.

**- A picture of a cute animal (not mandatory but encouraged)**

⏳

PR #834, opened 2021-11-23
feat: add in-source config declaration support

eduardoboucas:
**- Summary**

Adds support for in-source config declarations with the `schedule` helper.

Currently works with esbuild only. We can add support for other bundlers in subsequent PRs.

It's sitting behind a new `parseISC` feature flag.

Supersedes https://github.com/netlify/zip-it-and-ship-it/pull/684.

**- Test plan**

Added new test.

thumb2-australian-wombat-cute-animals-summers-wildlife-australia


src/runtimes/node/bundlers/esbuild/plugin_in_source_config.ts
```diff
@@ -0,0 +1,21 @@
...
+const getISCImportFinderPlugin = ({ importerPaths }: { importerPaths: Set<string> }): Plugin => ({
+  name: 'find-isc-imports',
+  setup(build) {
+    const filter = new RegExp(`^${IN_SOURCE_CONFIG_MODULE}$`)
+
+    build.onResolve({ filter }, (args) => {
+      importerPaths.add(args.importer)
+
+      // eslint-disable-next-line unicorn/no-useless-undefined
+      return undefined
```
eduardoboucas, line 16: This is odd, but without this return I'm getting a TypeScript error:

```
Argument of type '(args: OnResolveArgs) => void' is not assignable to parameter of type '(args: OnResolveArgs) => OnResolveResult | Promise<OnResolveResult | null | undefined> | null | undefined'.
  Type 'void' is not assignable to type 'OnResolveResult | Promise<OnResolveResult | null | undefined> | null | undefined'.ts(2345)
```

netlify-team-account-1: I was also surprised by this at first, but found a good reasoning:

> It's not the same as `undefined` because a function returning `void` might, in javascript at runtime, return something other than `undefined`.
> ~ [Source](https://github.com/Microsoft/TypeScript/issues/30880#issuecomment-482547767)

eduardoboucas: What confuses me is that we _are_ returning `undefined` in this exact same esbuild plugin hook elsewhere in the codebase: https://github.com/netlify/zip-it-and-ship-it/blob/main/src/runtimes/node/bundlers/esbuild/plugin_native_modules.ts#L40. So why does it work in one instance but not the other?

src/runtimes/node/parser/imports.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (node.type !== 'VariableDeclaration') {
+    return []
+  }
+
+  const { declarations } = node
+  const requireDeclaration = declarations.find(
+    (declaration) => declaration.type === 'VariableDeclarator' && isRequire(declaration.init, importPath),
+  )
+
+  if (requireDeclaration === undefined || requireDeclaration.id.type !== 'ObjectPattern') {
```
netlify-team-account-1, line 31: Testing for `ObjectPattern` will mean we miss this:

```js
const nfFunctions = require("@netlify/functions")

exports.handler = nfFunctions.cron(...)
```

Probably out-of-scope for this PR, though.

src/runtimes/node/parser/exports.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import type { ISCExport } from '../in_source_config'
+
+import { isModuleExports } from './helpers'
+
+// Finds the main handler export in an AST.
+const getMainExport = (nodes: Statement[]) => {
+  let handlerExport: ISCExport[] = []
+
+  nodes.some((node) => {
```
netlify-team-account-1, line 11: could we use `nodes.find` instead?

src/runtimes/node/parser/helpers.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const object = expression.slice(0, -1)
+  const [property] = expression.slice(-1)
+
+  if (node.property.type !== 'Identifier' || node.property.name !== property) {
+    return false
+  }
+
+  if (object.length > 1) {
+    return isDotExpression(node.object, object)
```
netlify-team-account-1, line 25: see to read my comment 🤪

src/runtimes/node/parser/exports.ts
```diff
@@ -0,0 +1,85 @@
...
+    const { id, type } = childDeclaration
+
+    return type === 'VariableDeclarator' && id.type === 'Identifier' && id.name === 'handler'
+  })
+
+  if (handlerDeclaration?.init?.type !== 'CallExpression') {
+    return []
+  }
+
+  const exports = getExportsFromCallExpression(handlerDeclaration.init)
```
netlify-team-account-1, line 68: there can only ever be one export called `handler`. are we thinking in multiples here because to future-proof this for potential additional exports?

src/utils/format_result.ts
```diff
@@ -3,14 +3,18 @@ import { RuntimeName } from '../runtimes/runtime'
...
+}
 
 // Takes the result of zipping a function and formats it for output.
 const formatZipResult = (archive: FunctionArchive) => {
   const functionResult: FunctionResult = {
     ...archive,
+    inSourceConfig: undefined,
     runtime: archive.runtime.name,
-    schedule: archive?.config?.schedule,
+    schedule: archive.inSourceConfig?.schedule || archive?.config?.schedule,
```
netlify-team-account-1, line 17:
```suggestion
    schedule: archive.inSourceConfig?.schedule ?? archive?.config?.schedule,
```

src/utils/format_result.ts
```diff
@@ -3,14 +3,18 @@ import { RuntimeName } from '../runtimes/runtime'
...
+type FunctionResult = Omit<FunctionArchive, 'runtime'> & {
+  runtime: RuntimeName
+  schedule?: string
+}
 
 // Takes the result of zipping a function and formats it for output.
 const formatZipResult = (archive: FunctionArchive) => {
   const functionResult: FunctionResult = {
     ...archive,
+    inSourceConfig: undefined,
```
netlify-team-account-1, line 15: why not this?

```suggestion
    inSourceConfig: archive.inSourceConfig,
```

src/utils/format_result.ts
```diff
@@ -3,14 +3,18 @@ import { RuntimeName } from '../runtimes/runtime'
 
 import { removeUndefined } from './remove_undefined'
 
-type FunctionResult = Omit<FunctionArchive, 'runtime'> & { runtime: RuntimeName; schedule?: string }
+type FunctionResult = Omit<FunctionArchive, 'runtime'> & {
```
netlify-team-account-1, line 6: did something change here? why did prettier reformat this? 🤔

netlify-team-account-1 (approved):
💖  love this PR! left some minor comments.

src/runtimes/node/bundlers/esbuild/plugin_in_source_config.ts
netlify-team-account-1: Because the other doesn't return `void`, but `Promise<void>`.

src/runtimes/node/parser/imports.ts
eduardoboucas: Yeah, I think we can expand the patterns we catch in subsequent PRs, probably after the Beta release.

src/runtimes/node/parser/exports.ts
eduardoboucas: Yes, exactly. The overhead in complexity feels really small, and we prepare ourselves to accommodate different patterns.

src/runtimes/node/bundlers/esbuild/plugin_in_source_config.ts
eduardoboucas: Same deal if I make this one async.

Screenshot 2021-11-23 at 16 54 04

src/utils/format_result.ts
eduardoboucas: Because we want the individual properties generated by ISC (e.g. `schedules`) to be part of the result, not the `inSourceConfig` property itself.

src/utils/format_result.ts
eduardoboucas: No, but I had added a new property and then removed it, which I guess made Prettier change the formatting.

src/runtimes/node/parser/exports.ts
eduardoboucas: Sure. Done in https://github.com/netlify/zip-it-and-ship-it/pull/834/commits/1ed25fd7ca87bbeffe6f0edfa668f0d801912372.

src/utils/format_result.ts
netlify-team-account-1: But why do we set it to undefined then, instead of not including it at all?

netlify-team-account-1 (approved):

src/utils/format_result.ts
eduardoboucas: Because `inSourceConfig` is part of `archive` but we don't want it to be part of `functionResult`, so we must set it to `undefined` so that `removeUndefined` will get rid of it.

Issue #748, opened 2021-10-15
Upgrade minimal Node.js version to Node 12


ehmicky:
See background at https://github.com/netlify/team-dev/issues/34

We should upgrade the minimal Node.js version to Node `>=12.20.0` and make a major release.

This should only be done after Node 10 support in both Netlify CLI (https://github.com/netlify/cli/issues/3512) and Netlify Build (https://github.com/netlify/build/issues/3741) has been dropped first. 

This would not break sites because:
  - The buildbot is already running `@netlify/zip-it-and-ship-it` with Node 12. 
  - The CLI would not support Node 10 anymore.

ehmicky:
Done at https://github.com/netlify/zip-it-and-ship-it/pull/836

PR #835, opened 2021-11-23
chore: simplify `renovate.json5`

ehmicky:
Part of https://github.com/netlify/zip-it-and-ship-it/issues/748

This simplifies `renovate.json5` since those rules are already part of our shared Renovate configuration.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -9,14 +9,6 @@
   dependencyDashboard: true,
   automerge: false,
   packageRules: [
-    {
-      matchPackageNames: ['p-map'],
-      allowedVersions: '<5',
-    },
-    {
-      matchPackageNames: ['yargs'],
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -28,9 +20,5 @@
       packageNames: ['test'],
```
lukasholzer, line 20: shouldn't this be put in the top level config as well?

lukasholzer (approved):

eduardoboucas (approved):

renovate.json5
ehmicky: I am not sure. The comment says it is a fake dependency used in tests, so this might be specific to this repository.
@eduardoboucas Is this correct?

PR #836, opened 2021-11-23
chore!: drop support for Node 10

ehmicky:
Fixes https://github.com/netlify/zip-it-and-ship-it/issues/748

This drops Node 10 support.

This does not drop Node 8/10 support as a compilation target (`nodeVersion` option) since I believe this is separate from the Node.js version supported for consumers installing this module through npm, but please let me know if this is incorrect.

This can be reviewed, but can only be released once https://github.com/netlify/build/pull/3873 is released.

I will also fix the branch protection rule right before merging this.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -13,12 +13,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [10.18.0, '*']
+        node-version: [12.20.0, '*']
         exclude:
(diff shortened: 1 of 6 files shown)


lukasholzer (dismissed):

eduardoboucas (approved):
Looks great, thanks!

eduardoboucas (approved):

ehmicky:
I am realizing https://github.com/netlify/build/pull/3873 should be merged first. Could one of you please review it? Thanks!

lukasholzer (approved):

PR #847, opened 2021-11-30
feat: extend in-source configuration to all Node bundlers

eduardoboucas:
**- Summary**

Support for in-source configuration was added in https://github.com/netlify/zip-it-and-ship-it/pull/834 with support for esbuild only. The implementation relied on an esbuild plugin to find which files were importing `@netlify/functions`.

This PR simplifies that logic a bit by always parsing the function's entry file. This allows us to extend ISC support for all bundlers.

We may wish to revisit this in the future by leveraging the parsing round we're doing in each bundler to find out whether the library is being imported. We can't do this for ZISI though, so I think it's best to simplify this for the Beta release of Scheduled Functions and revisit later.

**- Test plan**

Adjusted existing tests.

Cute-Dog-Breeds


netlify-team-account-1 (approved):
LGTM!

PR #899, opened 2021-12-20
feat: add in-source-config detection to listFunctions

netlify-team-account-1:
**- Summary**

As discussed on Friday, this PR adds ISC to `listFunctions` (8537a801bd1c12f9b6cf311ca5098c0fdc011786...11e7c9294664cd7822cc64a8ef49f28cba2c39e3). It also adds a `listFunction` (singular) version (11e7c9294664cd7822cc64a8ef49f28cba2c39e3...402c013076a7d6859bf606455802d32c28585b3b).

**- Test plan**

Tests added.


netlify-team-account-1:
pre-released to https://www.npmjs.com/package/@netlify/zip-it-and-ship-it/v/5.4.0-isc-listfunctions to continue work on the CLI

src/runtimes/node/finder.ts
```diff
@@ -62,11 +65,13 @@ const getFunctionAtPath = async function (srcPath: string): Promise<SourceFile |
     return
   }
 
+  const iscValues = await findISCDeclarationsInPath(mainFile)
```
eduardoboucas, line 68: One thing to consider is that `listFunctions` is [called by Netlify Build](https://github.com/netlify/build/blob/088ff1bba18be6e49447ba2b250f1cb5f5d3af22/packages/build/src/plugins_core/functions/utils.js#L23) as part of the build process, in order to show a list of functions that will get bundled.

In those circumstances, getting access to the in-source config declarations is not needed, and the additional effort of parsing the main files will be wasted.

How do you feel about adding an additional parameter to `listFunctions`, disabled by default, which determines whether to extract ISC?

Also, I'm not sure that logic should be part of `getFunctionAtPath`, because finding whether a function exists and cracking it open to extract in-source config feel like two separate concerns. I'd prefer a similar approach to how we're [doing it now](https://github.com/netlify/zip-it-and-ship-it/blob/main/src/runtimes/node/index.ts#L120), where we augment the function object with ISC data (which in the case of `listFunctions` should happen conditionally).

src/runtimes/node/finder.ts
```diff
@@ -62,11 +65,13 @@ const getFunctionAtPath = async function (srcPath: string): Promise<SourceFile |
...
+
   const extension = extname(srcPath)
   const srcDir = stat.isDirectory() ? srcPath : dirname(srcPath)
   const name = basename(srcPath, extname(srcPath))
 
-  return { extension, filename, mainFile, name, srcDir, srcPath, stat }
+  return { extension, filename, mainFile, name, srcDir, srcPath, stat, inSourceConfig: iscValues }
```
eduardoboucas, line 74: [dust] Can we maintain alphabetical order of properties?

src/runtimes/runtime.ts
```diff
@@ -49,9 +57,18 @@ type ZipFunction = (
...
 
-export { FindFunctionsInPathsFunction, GetSrcFilesFunction, Runtime, RuntimeName, ZipFunction, ZipFunctionResult }
+export {
+  FindFunctionsInPathsFunction,
+  GetSrcFilesFunction,
+  Runtime,
+  RuntimeName,
+  ZipFunction,
+  ZipFunctionResult,
+  GetFunctionAtPathFunction,
```
eduardoboucas, line 73: [dust] Can we maintain alphabetical order of properties?

src/function.ts
```diff
@@ -20,6 +21,7 @@ interface SourceFile {
   srcDir: string
   srcPath: string
   stat: Stats
+  inSourceConfig?: ISCValues
```
eduardoboucas, line 24: [dust] Can we maintain alphabetical order of properties?

eduardoboucas (changes requested):

src/runtimes/go/index.ts
```diff
@@ -7,7 +7,7 @@ import { SourceFile } from '../../function'
 import { cachedLstat, cachedReaddir, FsCache } from '../../utils/fs'
 import { nonNullable } from '../../utils/non_nullable'
 import { detectBinaryRuntime } from '../detect_runtime'
-import { FindFunctionsInPathsFunction, Runtime, ZipFunction } from '../runtime'
+import { FindFunctionsInPathsFunction, GetFunctionAtPathFunction, Runtime, ZipFunction } from '../runtime'
```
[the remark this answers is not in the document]
netlify-team-account-1: What's your recommendation? Rename `GetFunctionAtPathFunction` to `FindFunctionInPathFunction`?
The idea behind *Get* was to convey that it doesn't do any complex *finding*, but only parses the one path it's been given. If you read that otherwise / don't think that distinction is worth the inconsistency, I'm happy to defer.

src/runtimes/index.ts
```diff
@@ -110,4 +112,28 @@ const getFunctionsFromPaths = async (
...
+ * Gets a list of functions found in a list of paths.
+ */
+const getFunctionFromPath = async (
+  path: string,
+  { config, featureFlags = defaultFlags }: { config?: Config; featureFlags?: FeatureFlags } = {},
+): Promise<FunctionSource | undefined> => {
+  const fsCache = makeFsCache()
+
+  for (const runtime of RUNTIMES) {
+    // eslint-disable-next-line no-await-in-loop
```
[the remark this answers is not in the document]
netlify-team-account-1: `getFunctionsFromPaths` implements this using `reduce`, but it's very hard to read. I'd prefer disabling the lint rule over the `reduce` implementation, really. How would you implement this (without parallelising it using `Promise.all`)?

src/runtimes/go/index.ts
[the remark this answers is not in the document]
eduardoboucas: But isn't the complex finding taking place in [`getFunctionAtPath`](https://github.com/netlify/zip-it-and-ship-it/blob/isc-on-listfunctions/src/runtimes/node/finder.ts#L54-L75)? That's where we're filtering out runtime-specific paths like `node_modules`, distinguishing files from directories, figuring out the various allowed filenames and extensions (via calling `getMainFile`), etc.

In [`findFunctionsInPaths`](https://github.com/netlify/zip-it-and-ship-it/blob/isc-on-listfunctions/src/runtimes/node/finder.ts#L29-L52) we're just taking the results from `getFunctionAtPath` and sorting them?

src/runtimes/node/finder.ts
netlify-team-account-1: implemented in 4048c42. is that what you had in mind?

src/runtimes/go/index.ts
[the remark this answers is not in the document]
netlify-team-account-1: I somehow skimmed over that bit. falls in line with me naming that parameter `mainFile` instead of `path`. will rename to `FindFunctionInPathFunction`.

src/main.ts
```diff
@@ -22,49 +24,95 @@ interface ListFunctionsOptions {
...
   featureFlags?: FeatureFlags
+  parseISC?: boolean
+}
+
+interface AugmentedFunctionSource extends FunctionSource {
+  inSourceConfig?: ISCValues
+}
+
+const augmentWithISC = async (func: FunctionSource): Promise<AugmentedFunctionSource> => {
+  if (func.runtime.name !== 'js') {
```
eduardoboucas, line 35:
```suggestion
  // ISC is currently only supported in JavaScript and TypeScript functions.
  if (func.runtime.name !== 'js') {
```

src/main.ts
```diff
@@ -22,49 +24,95 @@ interface ListFunctionsOptions {
   basePath?: string
   config?: Config
   featureFlags?: FeatureFlags
+  parseISC?: boolean
+}
+
+interface AugmentedFunctionSource extends FunctionSource {
```
eduardoboucas, line 30: [dust] Currently, this type only differs from `FunctionSource` in that it also includes ISC. Are you thinking it may be used in the future for other things? If we introduce new options to `listFunctions`, would that not be under a different conditional check than `parseISC`? And if so, do you think this should be called something like `FunctionSourceWithISC`?

src/main.ts
```diff
@@ -22,49 +24,95 @@ interface ListFunctionsOptions {
...
-  const functions = await getFunctionsFromPaths(paths, { featureFlags })
-  const listedFunctions = [...functions.values()].map(getListedFunction)
-  return listedFunctions
+  const functionsMap = await getFunctionsFromPaths(paths, { featureFlags, config })
+  const functions = [...functionsMap.values()]
+  const augmentedFunctions = parseISC ? await Promise.all(functions.map(augmentWithISC)) : functions
+  return augmentedFunctions.map(getListedFunction)
+}
+
+// List one Netlify Functions main entry file for a specific directory
```
eduardoboucas, line 61: I think this is a more accurate description?

```suggestion
// Finds a function at a specific path.
```

src/runtimes/node/finder.ts
eduardoboucas: You mean https://github.com/netlify/zip-it-and-ship-it/pull/899/commits/15c1cadfe33854a8bee6810cd66923f22f22a089? If so, yes! I left some minor comments below.

src/runtimes/index.ts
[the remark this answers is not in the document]
eduardoboucas: I think we should try to avoid disabling rules as much as possible, because that defeats the purpose of having them in the first place. Except for the ones in tests, I think all the cases where we're disabling a rule have a specific reason.

I personally would write it with a `reduce` in line with `getFunctionsFromPaths`, but if you think that's hard to read, perhaps we need to rethink the `no-await-in-loop` rule?

src/main.ts
netlify-team-account-1: I like it!

src/function.ts
netlify-team-account-1: outdated by https://github.com/netlify/zip-it-and-ship-it/pull/899/commits/15c1cadfe33854a8bee6810cd66923f22f22a089

src/runtimes/runtime.ts
netlify-team-account-1: outdated by https://github.com/netlify/zip-it-and-ship-it/pull/899/commits/15c1cadfe33854a8bee6810cd66923f22f22a089

src/main.ts
```diff
@@ -27,16 +28,30 @@ interface ListFunctionsOptions {
...
   const paths = await listFunctionsDirectories(srcFolders)
-  const functions = await getFunctionsFromPaths(paths, { featureFlags })
+  const functions = await getFunctionsFromPaths(paths, { featureFlags, config })
   const listedFunctions = [...functions.values()].map(getListedFunction)
   return listedFunctions
 }
 
+// List one Netlify Functions main entry file for a specific directory
+const listFunction = async function (
+  mainFile: string,
```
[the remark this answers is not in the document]
netlify-team-account-1: agreed! https://github.com/netlify/zip-it-and-ship-it/pull/899/commits/4048c42aa618c193d278f7f276544dc412ea0fd5

src/runtimes/go/index.ts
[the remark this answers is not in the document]
netlify-team-account-1: implemented in https://github.com/netlify/zip-it-and-ship-it/pull/899/commits/4048c42aa618c193d278f7f276544dc412ea0fd5

src/runtimes/index.ts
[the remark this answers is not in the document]
netlify-team-account-1: let's rethink the rule, and keep the exception for now.

src/main.ts
netlify-team-account-1: the `AugmentedFunctionSource` type is really the input type to `getListedFunction`. it's `FunctionSource` + whatever may need to be added to assemble the return value of `listFunctions` and `listFunction`. If there was a new flag, let's say `enablePartyMode`, then we can extend `AugmentedFunctionSource` to have an `partyModeResult?: Parteyyy` property, and `getListedFunction` could use that to assemble the return value.

eduardoboucas (approved):
🚀

PR #929, opened 2022-01-07
fix: use es2019 in tsconfig

XhmikosR:
**- Summary**

The supported Node.js version is >= 12 which means ES6 is outdated and produces more bloated code. You should make sure you use the same in all your projects :)

**- Test plan**

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,7 +11,7 @@
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "es6" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "target": "es2019" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
     // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
     // "jsx": "preserve",                                /* Specify what JSX code is generated. */
     // "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */


XhmikosR:
BTW it seems you are not including source maps in the npm package which is fine, but they are still generated and referenced.

eduardoboucas:
@netlify-team-account-1 tagging you for a review since you drove the TypeScript migration.

eduardoboucas:
> BTW it seems you are not including source maps in the npm package which is fine, but they are still generated and referenced.

It's useful to have source maps available during development, but we don't want to publish them as they will bloat the size of the package.

XhmikosR:
> > BTW it seems you are not including source maps in the npm package which is fine, but they are still generated and referenced.
> 
> It's useful to have source maps available during development, but we don't want to publish them as they will bloat the size of the package.

Yeah, I thought about it too, but it's probably better to use a different production environment then. It doesn't make a huge difference, I agree, but it's redundant to have the sourcemap URLs embedded in production since they are unused.

netlify-team-account-1:
> but it's redundant to have the sourcemap URLs embedded in production since they are unused.

to my knowledge, the source map references consists of a comment of ~80 characters or less at the end of each file. can you elaborate on why you think it's redundant? can this information also be found somewhere else?

XhmikosR:
> > but it's redundant to have the sourcemap URLs embedded in production since they are unused.
> 
> to my knowledge, the source map references consists of a comment of ~80 characters or less at the end of each file. can you elaborate on why you think it's redundant? can this information also be found somewhere else?

It's redundant because there are no sourcemap files in production. So, they are referenced, but not included in production. Hence why I suggested to use a different production config if you want to keep them for development :)

netlify-team-account-1 (approved):
Today I learned that ES2019 is fully supported from Node v12 on: https://node.green/#ES2019

tests are green,, so let's get this merged! 🚀

PR #930, opened 2022-01-07
fix: remove make-dir package and use `fs/promises`

XhmikosR:
**- Summary**

**- Test plan**

- CI is green

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -1,8 +1,8 @@
+import { promises as fs } from 'fs'
 import { resolve } from 'path'
 import { arch, platform } from 'process'
 
 import { FunctionResult } from './utils/format_result'
-import { writeFile } from './utils/fs'
 
 interface ManifestFunction {
   mainFile: string
@@ -33,7 +33,7 @@ const createManifest = async ({ functions, path }: { functions: FunctionResult[]
     version: MANIFEST_VERSION,
   }
 
-  await writeFile(path, JSON.stringify(payload))
+  await fs.writeFile(path, JSON.stringify(payload))
 }
 
 const formatFunctionForManifest = ({ mainFile, name, path, runtime, schedule }: FunctionResult): ManifestFunction => ({
--- a/src/utils/fs.ts
+++ b/src/utils/fs.ts
@@ -1,17 +1,7 @@
-import { lstat, readdir, readFile, stat, unlink, writeFile } from 'fs'
+import { promises as fs } from 'fs'
 import { dirname, format, join, parse, resolve } from 'path'
-import { promisify } from 'util'
-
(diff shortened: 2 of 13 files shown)


src/utils/fs.ts
```diff
@@ -1,9 +1,8 @@
 import { lstat, readdir, readFile, stat, unlink, writeFile } from 'fs'
```
XhmikosR, line 1: I'm unsure why the direct `fs/promises` imports are not used here in this file. Would be nice to get rid of `promisify` if the same can be done with `fs/promises`

XhmikosR:
This should be ready now assuming CI is green.

src/utils/fs.ts
```diff
@@ -88,29 +78,27 @@ const resolveFunctionsDirectories = (input: string | string[]) => {
...
   resolveFunctionsDirectories,
   safeUnlink,
-  pStat as stat,
-  pWriteFile as writeFile,
-  pReadFile as readFile,
   mkdirAndWriteFile,
 }
 export type { FsCache }
+
+export { stat, lstat, writeFile, readFile } from 'fs/promises'
```
XhmikosR, line 104: Not sure about this one. It doesn't make any sense to have these now.

XhmikosR:
All right, I need to fix the error on Node.js 12.20.0. It doesn't support `fs/promises`. Any suggestions how to tackle the issue?

XhmikosR:
How about this solution?

```js
import { promises as fs } from 'fs'

// foo

fs.writeFile()
```

src/utils/fs.ts
```diff
@@ -88,29 +78,25 @@ const resolveFunctionsDirectories = (input: string | string[]) => {
   return absoluteDirectories
 }
 
-const mkdirAndWriteFile: typeof pWriteFile = async (path, ...params) => {
+const mkdirAndWriteFile: typeof fs.writeFile = async (path, ...params) => {
```
XhmikosR, line 81: Not familiar with TS, I could use a confirmation

src/manifest.ts
```diff
@@ -1,8 +1,8 @@
+import { promises as fs } from 'fs'
```
XhmikosR, line 1: Maybe it'll be better and more consistent with your other projects to use `import { promises as fsPromises } from 'fs'`?

XhmikosR:
It seems here the supported versions differ from the ones in package.json: https://github.com/netlify/zip-it-and-ship-it/blob/10ac495f798acc62fe46c24dbcfaa197b5be889e/src/config.ts#L8

Not sure if they should be updated?

netlify-team-account-1 (approved):

PR #932, opened 2022-01-07
fix: remove unneeded ESLint suppressions

XhmikosR:
**- Summary**

Leftover from #926 I think

**- Test plan**

### DIFF
--- a/src/runtimes/index.ts
+++ b/src/runtimes/index.ts
@@ -122,7 +122,6 @@ const getFunctionFromPath = async (
   const fsCache = makeFsCache()
 
   for (const runtime of RUNTIMES) {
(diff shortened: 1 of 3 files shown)


erezrokah:
This should be coupled with https://github.com/netlify/zip-it-and-ship-it/pull/928 I think

XhmikosR:
Ah yeah, needs that update first. I had the new version locally. Feel free to cherry pick or land this after #928.

erezrokah:
Cherry picked into https://github.com/netlify/zip-it-and-ship-it/pull/928/commits/0e699aae536e98078d8f2bebe8335ebfea515658

PR #935, opened 2022-01-08
chore: minor refactoring

XhmikosR:
This is based on #930, so only the second patch is relevant.

Opening for discussion and I can split the changes you want later.

~~1. added `types` is package.json~~
~~2. added rimraf because the build dir wasn't being cleaned and could keep old files~~
3. a couple of minor refactorings which are subjective, so let me know if you want those
~~4. disabled source maps in TypeScript since the maps aren't included in the package; if they are supposed to be used in tests, maybe there should be a different production and dev config?~~

**- Summary**

**- Test plan**

### DIFF
--- a/src/utils/archive_size.ts
+++ b/src/utils/archive_size.ts
@@ -6,13 +6,11 @@ import type { FunctionArchive } from '../function'
 // Returns the input object with an additional `size` property containing the
 // size of the file at `path` when it is a ZIP archive.
 const addArchiveSize = async (result: FunctionArchive) => {
-  const { path } = result
-
-  if (extname(path) !== '.zip') {
+  if (extname(result.path) !== '.zip') {
     return result
   }
 
-  const { size } = await fs.stat(path)
+  const { size } = await fs.stat(result.path)
 
   return { ...result, size }
 }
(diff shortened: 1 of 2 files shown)


XhmikosR:
I removed the `sourcemap: false` change from tsconfig, but I still think you should look at disabling sourcemap for production since you don't ship them in the npm package but they are still referenced. Not a huge deal, just saves a few bytes and it's proper IMHO :)

src/runtimes/node/bundlers/esbuild/plugin_native_modules.ts
```diff
@@ -48,23 +48,20 @@ const getNativeModulesPlugin = (externalizedModules: NativeNodeModules): Plugin
...
-            if (externalizedModules[args.path] === undefined) {
-              // eslint-disable-next-line no-param-reassign
-              externalizedModules[args.path] = {}
-            }
-
-            // eslint-disable-next-line no-param-reassign
-            externalizedModules[args.path][modulePath] = packageJsonData.version
-
-            return { path: args.path, external: true }
+          if (isNative === false) {
```
XhmikosR, line 51: Not 100% sure it's fine, so the more eyes the better. :)

netlify-team-account-1:
Hey @XhmikosR! Thank you for your continued contributions to our projects, we really appreciate your help. We aim for PRs to fix a bug / introduce a new feature. For refactorings (like this PR), it's probably best to discuss them more holistically first, for example in an issue. Could you maybe open up an issue for this, so we can discuss potential changes to code style rules / linting behind this PR?

XhmikosR:
This was part of another branch I have, I don't mind closing this since it's not worth discussing such a minor refactoring. :)

PR #945, opened 2022-01-13
chore: update package.json fields and add rimraf

XhmikosR:
**- Summary**

1. added `types` is package.json
2. added rimraf because the build dir wasn't being cleaned and could keep old files

**- Test plan**

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,17 +3,17 @@
   "version": "5.4.0",
   "description": "Zip it and ship it",
   "main": "./dist/main.js",
+  "types": "./dist/main.d.ts",
   "bin": {
     "zip-it-and-ship-it": "dist/bin.js"
   },
   "files": [
-    "dist/**/*.js",
-    "dist/**/*.d.ts"
+    "dist/**/*.{js,ts}"
   ],
   "scripts": {
     "prepublishOnly": "npm ci && npm test",
     "prepack": "npm run build",
-    "build": "tsc",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,17 +3,17 @@
...
   "files": [
-    "dist/**/*.js",
-    "dist/**/*.d.ts"
+    "dist/**/*.{js,ts}"
   ],
   "scripts": {
     "prepublishOnly": "npm ci && npm test",
     "prepack": "npm run build",
-    "build": "tsc",
+    "build": "rimraf dist && tsc",
```
XhmikosR, line 16: This change is probably worth having it in all your TS projects to make sure the dist directory doesn't contain any leftovers (probably doesn't apply for automated publishes where the dist dir is clean already).

netlify-team-account-1:
Hi! can you elaborate on the issues this PR fixes? I think typescript types are already working at the moment, and as long as the `.d.ts` files are besides their `.js` files, there's no need to specify the `"types"` property in `package.json`.

package.json
netlify-team-account-1: We are publishing via CI, so pulling in a new dev dependency is unneccessary in this case.

XhmikosR: Cool, I'll revert it I guess :)

XhmikosR:
No problem, types is automatically added by `npm init -y` and I prefer to be explicit personally, but no worries :)

PR #951, opened 2022-01-16
fix(deps): test my node-precinct fork

XhmikosR:
Opening this to demonstrate the size on disk reduction. Also fixes one npm install warning about `flatten`.

Size of `@netlify/zip-it-and-ship-it`'s `node_modules` folder for production dependencies:

```bat
rem Windows command
git clean -dfx && npm install --production --ignore-scripts && for /r node_modules /d %F in (.) do @dir /b "%F" | findstr "^" >nul || rmdir %~fF
```

```
Before:  126 MB  (132.340.985 bytes)
After:  74,5 MB   (78.160.053 bytes)
Diff:   51.5 MB
```

I've opened upstream PRs to help with this:

1. https://github.com/pahen/detective-typescript/pulls
2. https://github.com/dependents/node-precinct/pulls

When we have new detective-typescript and and node-precinct versions, you should be able to update your packages. The root cause is the multiple typescript versions which are not deduped.

The Node.js 12.20 error is because I use GitHub repos.

---

**- Summary**

**- Test plan**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,14 +31,14 @@
         "p-map": "^4.0.0",
         "path-exists": "^4.0.0",
         "pkg-dir": "^5.0.0",
-        "precinct": "^8.3.1",
+        "precinct": "github:XhmikosR/node-precinct#deps-ts",
         "read-package-json-fast": "^2.0.2",
         "require-package-name": "^2.0.1",
         "resolve": "^2.0.0-next.1",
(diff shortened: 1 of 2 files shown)


XhmikosR:
@erezrokah maybe you could contact the 2 package maintainers above in order to get things moving?

XhmikosR:
Closing since there should be a major version soon upstream.

PR #955, opened 2022-01-18
feat: add `zipGo` config property

eduardoboucas:
#### Summary

Go functions must be zipped before we upload them to AWS. This PR adds a `zipGo` configuration option, which determines whether the output of a Go function is a binary or a ZIP containing it.

We used to have a similar configuration property, but it was removed in #372 because it wasn't being used anywhere. We shouldn't need to use it in buildbot because when it finds a Go binary, it [takes care of wrapping it in a ZIP archive](https://github.com/netlify/open-api/blob/master/go/porcelain/deploy.go#L741-L757). This is why this flow currently works in buildbot and fails in the CLI.

Once this config property is in place, we should conditionally set it based on whether we are bundling in buildbot or in the CLI.

Part of https://github.com/netlify/cli/issues/1147.

istockphoto-1172290687-612x612

### DIFF
--- a/tests/fixtures/go-simple/test.go
+++ b/tests/fixtures/go-simple/test.go
@@ -1,7 +1,0 @@
-package main
-
-import "fmt"
-
-func main() {
-	fmt.Println("test")
-}
--- a/src/config.ts
+++ b/src/config.ts
@@ -19,6 +19,7 @@ interface FunctionConfig {
   processDynamicNodeImports?: boolean
   rustTargetDirectory?: string
   schedule?: string
+  zipGo?: boolean
 }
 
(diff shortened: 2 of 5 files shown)


tests/fixtures/go-simple/test.go
```diff
@@ -1,7 +0,0 @@
-package main
```
netlify-team-account-1, line 1: I don't quite get why this file is removed. was that an accident?

netlify-team-account-1: makes sense, thank you! didn't see the binary in there.

netlify-team-account-1 (approved):

PR #957, opened 2022-01-19
feat: zip Go functions from source when `zipGo` is enabled

eduardoboucas:
#### Summary

This is a follow up to #955. When `zipGo` is enabled, we must zip the binary even if it was created from source by us, which is currently not handled.


karagulamos (approved):
🚀

PR #966, opened 2022-01-27
fix: pin `@babel/parser` to `7.16.8`

ehmicky:
Some builds are experiencing a big slowdown when we upgrade `@babel/parser` from `7.16.8` to `7.16.10` (note: there is no `7.16.9` release).

The only PR introduced by that release is https://github.com/babel/babel/pull/14130, so it appears to have some issues. It could be one of the following bugs, but we are not sure yet: https://github.com/babel/babel/issues/14182, https://github.com/babel/babel/issues/14175

We are certain that the problem comes from this specific package and this specific version. We also know the problem comes from `@babel/parser` as a direct dependency of `@netlify/zip-it-and-ship-it`, as opposed to as an indirect dependency through `precinct` and `node-source-walker`.

This PR pins the version of `@babel/parser` until we figure out why this version is creating a problem on our side.

---

      passes our tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "5.5.1",
       "license": "MIT",
       "dependencies": {
-        "@babel/parser": "^7.15.7",
+        "@babel/parser": "7.16.8",
         "@netlify/esbuild": "^0.13.6",
         "@vercel/nft": "^0.17.0",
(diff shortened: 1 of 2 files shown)


eduardoboucas (approved):
❤️

Issue #749, opened 2021-10-15
Use pure ES modules


ehmicky:
See background at https://github.com/netlify/team-dev/issues/36

Once https://github.com/netlify/zip-it-and-ship-it/issues/748 is done and released, we should use pure ES modules and make a major release.

This is more than just switching from CommonJS to `import`/`export`. See [this list](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) for other changes which might be involved. This should be broken in many PRs, as much as possible, to lower the risk. Also, non-breaking changes (such as adding file extensions in imports, or loading JSON files differently) should be done before the breaking changes (such as using `import`/`export` statements).

This should only be done after pure ES modules are supported in both Netlify CLI (https://github.com/netlify/cli/issues/3514) and Netlify Build (https://github.com/netlify/build/issues/3742).

ehmicky:
I have opened several PRs to tackle most of the problems related to migrating to pure ES modules:
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/970
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/971
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/972
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/973
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/974
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/975
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/976
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/977
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/978
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/979
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/980
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/981
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/982
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/983
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/984
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/985
  - [ ] https://github.com/netlify/zip-it-and-ship-it/pull/986

Two problems are still pending:
[...]

PR #974, opened 2022-01-28
fix: use `.js` and `index.js` in imports

ehmicky:

Pure ES modules require file extensions to be specified in file imports. In TypeScript, this is done by using `.js`, i.e. the path to the transpiled file, not to the source file (`.ts`) (which might not seem intuitive).

Also, pure ES modules require `index.js` imports to be explicit.

This PR changes import file paths to match this.

---

      passes our tests.


.eslintrc.js
```diff
@@ -2,11 +2,18 @@ const { overrides } = require('@netlify/eslint-config-node')
 
 module.exports = {
   extends: '@netlify/eslint-config-node',
+  rules: {
+    'import/extensions': ['error', 'ignorePackages'],
```
ehmicky, line 6: Some new ESLint rules are needed with the new import patterns.

erezrokah (dismissed):

erezrokah (approved):

eduardoboucas (approved):

PR #977, opened 2022-01-28
chore: use ES modules in benchmarks

ehmicky:

Depends on #975

This migrates the benchmarks to using pure ES modules.

---

      passes our tests.

### DIFF
--- a/benchmarks/package.json
+++ b/benchmarks/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "module"
+}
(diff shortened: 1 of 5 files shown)


benchmarks/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
```
ehmicky, line 2: Temporary workaround since the top-level `package.json` does not use pure ES modules.
To remove once this is not the case anymore.

erezrokah (dismissed):

erezrokah (approved):

PR #982, opened 2022-01-28
fix: fix JSON imports for ES modules

ehmicky:

JSON imports with ES modules support in Node.js is still experimental. Therefore, to migrate to ES modules, we need to replace it with doing manual `fs.readFile()` + `JSON.parse()`.

---

      passes our tests.


.eslintrc.js
```diff
@@ -2,6 +2,11 @@ const { overrides } = require('@netlify/eslint-config-node')
 
 module.exports = {
   extends: '@netlify/eslint-config-node',
+  rules: {
+    // This rule enforces using Buffers with `JSON.parse()`. However, TypeScript
+    // does not recognize yet that `JSON.parse()` accepts Buffers as argument.
```
ehmicky, line 7: Due to the above, we pass `utf8` when in TypeScript, but not when in JavaScript.

erezrokah (approved):

erezrokah (approved):

PR #980, opened 2022-01-28
fix: use dynamic `import()` in tests

ehmicky:
Part of https://github.com/netlify/zip-it-and-ship-it/issues/749

This replaces `require()` to dynamic `import()` in tests, as part of the migration to ES modules.

---

For us to review and ship your PR efficiently, please perform the following steps:

- [x] Open a [bug/issue](https://github.com/netlify/zip-it-and-ship-it/issues/new/choose) before writing your code
      🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re
      fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [x] Update or add tests (if any source code was changed or added) 🧪
- [x] Update or add documentation (if features were changed or added) 📝
- [x] Make sure the status checks below are successful ✅

PR #1004, opened 2022-02-07
feat: generate pure ESM functions for NFT

eduardoboucas:
#### Summary

As of last month, AWS Lambda [supports native ESM functions](https://aws.amazon.com/blogs/compute/using-node-js-es-modules-and-top-level-await-in-aws-lambda/). Currently, we're unable to take advantage of this functionality since zip-it-and-ship-it always generates CJS functions.

This PR starts changing that by generating pure ESM functions when all of the following conditions are met:

- The bundler is NFT
- The Node.js version (supplied via the `nodeVersion` configuration property) is greater than or equal to Node 14
- The `zisi_pure_esm` feature flag is enabled

In any other scenario, zip-it-and-ship-it continues to transpile ESM and generate CJS functions.

### DIFF
--- a/src/runtimes/node/bundlers/esbuild/index.ts
+++ b/src/runtimes/node/bundlers/esbuild/index.ts
@@ -122,6 +122,7 @@ const bundle: BundleFunction = async ({
     bundlerWarnings,
     inputs,
     mainFile: normalizedMainFile,
+    moduleFormat: 'cjs',
     nativeNodeModules,
     nodeModulesWithDynamicImports,
     srcFiles: [...supportingSrcFiles, ...bundlePaths.keys()],
--- a/src/runtimes/node/bundlers/index.ts
+++ b/src/runtimes/node/bundlers/index.ts
@@ -1,13 +1,17 @@
 import type { Message } from '@netlify/esbuild'
 
-import type { NodeBundlerName } from '..'
 import { FunctionConfig } from '../../../config'
 import { FeatureFlag, FeatureFlags } from '../../../feature_flags'
 import { FunctionSource } from '../../../function'
+import { detectEsModule } from '../utils/detect_es_module'
+import { ModuleFormat } from '../utils/module_format'
+import { NodeVersionString } from '../utils/node_version'
 
 import esbuildBundler from './esbuild'
 import nftBundler from './nft'
 import zisiBundler from './zisi'
+
+export type NodeBundlerName = 'esbuild' | 'esbuild_zisi' | 'nft' | 'zisi'
 
 // TODO: Create a generic warning type
 type BundlerWarning = Message
@@ -49,6 +53,7 @@ type BundleFunction = (
   cleanupFunction?: CleanupFunction
   inputs: string[]
   mainFile: string
+  moduleFormat: ModuleFormat
   nativeNodeModules?: NativeNodeModules
   nodeModulesWithDynamicImports?: string[]
   srcFiles: string[]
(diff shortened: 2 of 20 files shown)


src/runtimes/node/bundlers/esbuild/index.ts
```diff
@@ -122,6 +122,7 @@ const bundle: BundleFunction = async ({
     bundlerWarnings,
     inputs,
     mainFile: normalizedMainFile,
+    moduleFormat: 'cjs',
```
eduardoboucas, line 125: To be addressed in a follow-up PR.

src/runtimes/node/bundlers/index.ts
```diff
@@ -86,5 +90,37 @@ const getBundler = (name: NodeBundlerName): NodeBundler => {
   }
 }
 
-export { getBundler }
+// We use ZISI as the default bundler, except for certain extensions, for which
+// esbuild is the only option.
+const getDefaultBundler = async ({
```
eduardoboucas, line 95: This function was moved from `src/runtimes/node/index.ts`. The code was not changed.

src/runtimes/node/utils/normalize_path.ts
```diff
@@ -0,0 +1,24 @@
+import { normalize, sep } from 'path'
+
+import unixify from 'unixify'
+
+// `adm-zip` and `require()` expect Unix paths.
+// We remove the common path prefix.
+// With files on different Windows drives, we remove the drive letter.
+const normalizeFilePath = function ({
```
eduardoboucas, line 8: Moved unchanged from `src/runtimes/node/utils/zip.ts`.

tests/main.js
```diff
@@ -506,6 +543,70 @@ testMany(
...
+    t.true(func1.handler())
+    t.true(func2.handler())
+
+    const functionsAreESM = await Promise.all(
+      functionPaths.map((functionPath) => detectEsModule({ mainFile: functionPath })),
+    )
+
+    t.true(functionsAreESM.every(Boolean))
+  },
+)
```
ehmicky, line 608: Nice tests!

tests/main.js
```diff
@@ -506,6 +543,70 @@ testMany(
...
+    })
+    const { files, tmpDir } = await zipFixture(t, fixtureName, {
+      length,
+      opts,
+    })
+
+    await unzipFiles(files, (path) => `${path}/../${basename(path)}_out`)
+
+    const functionPaths = [join(tmpDir, 'func1.zip_out', 'func1.js'), join(tmpDir, 'func2.zip_out', 'func2.js')]
+    const func1 = await import(functionPaths[0])
```
ehmicky, line 564: `pathToFileURL()` must be used, like in https://github.com/netlify/zip-it-and-ship-it/pull/980/files#diff-91254b3a8e85bba2e8bad93724729fcd3531241f5b57196acc1e49db3e856a49R113

This is because `import()` argument is a URI, not a file path (unlike `require()`).
On Windows, absolute file paths like `C:\...` are not valid URLs. Also, URIs are percent-encoded, so this would also fail for file paths with characters which would normally be percent-encoded.

tests/main.js
```diff
@@ -506,6 +543,70 @@ testMany(
...
```
ehmicky, line 596: Same comment as above.

ehmicky (changes requested):
Looks good! Except a small issue with some tests on Windows.

tests/main.js
eduardoboucas: Thank you! Done in https://github.com/netlify/zip-it-and-ship-it/pull/1004/commits/7156614214c1350c0174fce855119d17289f562b.

src/runtimes/node/bundlers/index.ts
```diff
@@ -86,5 +90,37 @@ const getBundler = (name: NodeBundlerName): NodeBundler => {
...
+  mainFile,
+  featureFlags,
+}: {
+  extension: string
+  mainFile: string
+  featureFlags: FeatureFlags
+}): Promise<NodeBundlerName> => {
+  const { defaultEsModulesToEsbuild, traceWithNft } = featureFlags
+
+  if (['.mjs', '.ts'].includes(extension)) {
```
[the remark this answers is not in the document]
eduardoboucas: Definitely!

ehmicky (approved):

PR #1018, opened 2022-02-14
feat: add pure ESM support to esbuild and default bundlers

eduardoboucas:
#### Summary

This is a follow-up to #1004, adding support for pure ESM functions to:

- The default bundler, using NFT to trace dependencies when an ESM file with a `.js` extension is found (https://github.com/netlify/zip-it-and-ship-it/commit/a468380cc1f3d80a1f17f8b416dcec1cbaf5d26f)
- esbuild (https://github.com/netlify/zip-it-and-ship-it/commit/8e0c72584180f90d01bf77d96705b89193e00684)


src/feature_flags.ts
```diff
@@ -3,9 +3,9 @@ import { env } from 'process'
 export const defaultFlags: Record<string, boolean> = {
   buildGoSource: Boolean(env.NETLIFY_EXPERIMENTAL_BUILD_GO_SOURCE),
   buildRustSource: Boolean(env.NETLIFY_EXPERIMENTAL_BUILD_RUST_SOURCE),
-  defaultEsModulesToEsbuild: Boolean(env.NETLIFY_EXPERIMENTAL_DEFAULT_ES_MODULES_TO_ESBUILD),
   parseWithEsbuild: false,
   traceWithNft: false,
+  zisi_detect_esm: false,
```
eduardoboucas, line 8: This determines whether we'll look at a `.js` file and figure out whether it's CJS or ESM.

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -84,11 +87,21 @@ export const bundleJsFile = async function ({
   // URLs, not paths, so even on Windows they should use forward slashes.
   const sourceRoot = targetDirectory.replace(/\\/g, '/')
 
+  // Configuring the output format of esbuild. The `includedFiles` array we get
```
eduardoboucas, line 90: We were not including a `package.json` when bundling with esbuild. This becomes a problem with ESM functions, since an ESM file and a lack of a `package.json` with `{"type": "module"}` would lead to a CJS function with ESM syntax.

Skn0tt:
added a minor refactoring with TypeScript syntax. was hard to capture in a GitHub suggestion, so committed instead - let me know if you disagree :) https://github.com/netlify/zip-it-and-ship-it/pull/1018/commits/0e95b0f5c8b7893b31e45779d66df09c386609a3

src/runtimes/node/bundlers/esbuild/bundler_target.ts
```diff
@@ -10,18 +13,41 @@ const versionMap = {
...
-  return DEFAULT_VERSION
+  return versionMap[`${DEFAULT_NODE_VERSION}.x`]
+}
+
+const getModuleFormat = async (
+  srcDir: string,
+  featureFlags: FeatureFlags,
+  configVersion?: string,
+): Promise<{ includedFiles: string[]; moduleFormat: ModuleFormat }> => {
+  const packageJsonFile = await getClosestPackageJson(srcDir)
```
Skn0tt, line 31: I'm not sure why we need to use `getClosestPackageJson` in this case. `srcDir` would be smth like `projectDir/netlify/functions/my-function.ts`, right? The difference between `getClosesPackageJson` and `getPackageJson` is that one has an optimisation to not read further than the nearest `node_modules` folder, which wouldn't be the case here. What am I missing?

eduardoboucas:
> added a minor refactoring with TypeScript syntax. was hard to capture in a GitHub suggestion, so committed instead - let me know if you disagree :) [0e95b0f](https://github.com/netlify/zip-it-and-ship-it/commit/0e95b0f5c8b7893b31e45779d66df09c386609a3)

I initially had that and was torn on which one was more readable. If you prefer that approach, I'm more than happy to use it!

src/runtimes/node/bundlers/esbuild/bundler_target.ts
eduardoboucas: `getClosestPackageJson` returns the contents of the `package.json` _and_ its path, whereas `getPackageJson` returns just the contents. I agree that we should consolidate these, but I'd prefer to do that as a separate PR to reduce the amount of changes. What do you think?

ehmicky (approved):

src/runtimes/node/bundlers/esbuild/bundler_target.ts
Skn0tt: oh, somehow missed this reply. agree! I'll work on that.

PR #1030, opened 2022-02-18
feat: add support to per-function configuration files

eduardoboucas:
#### Summary

Adds support to per-function configuration files. These files are only taken into account if they are inside a directory that is part of the new `configFileDirectories` configuration property.

Take the following project structure as an example:

```
.netlify/
  functions-internal/
    internal-function-1/
      internal-function-1.json
      internal-function-1.js
    internal-function-2.json
    internal-function-2.ts
netlify/
  functions/
    user-function-1.ts
```

If `configFileDirectories` contains `.netlify/functions-internal`, then `internal-function-1` and `internal-function-2` will read configuration properties from `.netlify/functions-internal/internal-function-1/internal-function-1.json` and `.netlify/functions-internal/internal-function-2.json`, respectively.

Even if there was a `netlify/functions/user-function-1.json` file, it would be ignored if `netlify/functions` was not part of `configFileDirectories`.

Any configuration properties loaded from a per-function configuration file take precedence over the global `config` object.

where-to-see-sloths-in-costa-rica


danez:
@eduardoboucas What was the idea behind restricting this to certain directories instead of allowing it for every function? We did not want to expose this config support to user-generated functions?

eduardoboucas:
> What was the idea behind restricting this to certain directories instead of allowing it for every function? We did not want to expose this config support to user-generated functions?

Yes, exactly! Once we get to ISC, I think we'll want to allow all types of functions to be configured that way, but until then I don't think it makes sense for user-generated functions to use this functionality. Asking people to write a per-function JSON feels incredibly unergonomic, and they'd be better off using `netlify.toml` instead.

> Also do you think that zisi is the right place for reading the config?

Note that zip-it-and-ship-it is the only package capable of resolving a functions config. Imagine this:

```toml
[functions."next_*"]
node_bundler = "nft"
```

`@netlify/config` doesn't know how to unpack this into a list of functions, because it doesn't know how to tell whether a file is a function. This is why zip-it-and-ship-it is currently responsible for handling functions configuration.

Personally, I don't think that passing to zip-it-and-ship-it a configuration property that it doesn't use or understand is that big of a problem, but I'm happy to discuss any alternative options you have in mind.

danez:
Other question: The Proposal only specifies `nodeBundler` and `included_files` as valid configuration options coming from the deployment config. The current implementation does not limit the config though, that means [all config options](https://github.com/netlify/zip-it-and-ship-it/blob/main/src/config.ts#L8-L20) from zisi are currently available to be set in the config.

Should we limit/restrict this?

I personally don't see any downside of this, but I probably am missing something.

Should we add validation of the config? We can also do that later, once we introduce non zisi config stuff. At least then it would make sense IMHO.

@jackiewmacharia @eduardoboucas

eduardoboucas:
I also don't see any obvious downsides of allowing all configuration properties. I think the proposal mentioned those two as examples of properties we have an identified use case for in the deploy configuration API.

src/config.ts
```diff
@@ -1,11 +1,15 @@
...
 import { FunctionSource } from './function.js'
 import type { NodeBundlerName } from './runtimes/node/bundlers/types.js'
 import type { NodeVersionString } from './runtimes/node/index.js'
 import { minimatch } from './utils/matching.js'
 
-export interface FunctionConfig {
+interface FunctionConfig {
   externalNodeModules?: string[]
   includedFiles?: string[]
   includedFilesBasePath?: string
```
jackiewmacharia, line 15: We need one more field here: `nodeBundler?: string` to allow for bundler customisation.

src/config.ts
```diff
@@ -56,3 +98,23 @@ export const getConfigForFunction = ({
...
+
+const getFromFile = async (func: FunctionWithoutConfig): Promise<FunctionConfig> => {
+  const filename = `${basename(func.mainFile, extname(func.mainFile))}.json`
+  const configFilePath = join(dirname(func.mainFile), filename)
+
+  try {
+    const data = await fs.readFile(configFilePath, 'utf8')
+    const configFile = JSON.parse(data) as FunctionConfigFile
+
+    if (configFile.version === 1) {
```
jackiewmacharia, line 110: Curious as to how we will use the version field. Is it for us to know what version of functions bundling a framework could be running or can it also be used by a framework to track which config set was used with which release?

jackiewmacharia (commented):
Looks good overall, left a few comments.

src/config.ts
jackiewmacharia: My bad, `nodeBundler` format needs to take a path to the main file as well as the format for file configuration. i.e `nodeBundler?: NodeBundlerConfig`.

I'm also now thinking that this should come with work on allowing of skipping bundling altogether: https://github.com/netlify/pillar-runtime/issues/201

```ts
interface NodeBundlerConfig {
    mainFile: string,
    moduleFormat: string
}
```

----------

example json
```json
{
  "nodeBundler": {
    "mainFile": "/some/path/to/a/function.js",
    "moduleFormat": "esm"
  },
  "includedFiles": ["some/path/**/*.js", "!other/path/**/*.js"]
}
```

jackiewmacharia: We may also not need to specify mainFile here since we are already handling one function at a time at this point, we may be able to get away with: `nodeBundler?: NodeBundlerName`.

eduardoboucas: I think you still need a `mainFile` field, because we need to know which file is the entry point to the Lambda.

src/config.ts
eduardoboucas: I always like to version these files, because it makes it easy to introduce breaking changes in their format if we ever need to. If we do introduce new versions, consumers would check this field first and treat the rest of the file accordingly.

I would say it's very unlike that we'll ever need to introduce a breaking change in this file, but the price of adding this field is very low so I figured it was worth doing. It's totally fine to remove it if you don't see a need for it, though.

src/config.ts
danez: But we are having the convention here that `function.js` has the config file `function.json` ins the same folder. So wouldn't function.js always be the main file?

eduardoboucas: `function.js` is a source file, whereas `mainFile` is the distribution/compiled file that lives in the ZIP archive. Sometimes those two files will be the same, but not always. For example, when you're dealing with a TypeScript function, the source file will be something like `function.ts`, but your `mainFile` has to be a JavaScript because the Lambda runtime doesn't understand TypeScript.

You can see an example of that [here](https://github.com/netlify/zip-it-and-ship-it/blob/main/src/runtimes/node/bundlers/esbuild/index.ts#L112)

danez: Also for the nideBundler:  I think it makes sense to make it more clear to "users" what the config actually does if it is just that:

```json
"nodeBundler": {
  "mainFile": "/some/path/to/a/function.js",
  "moduleFormat": "esm"
},
```
I would not know what this actually does.

I think it would be more clear if we would for all bundlers do something like this:

```ts
type NodeBundlerName = 'esbuild' | 'nft' | 'zisi';
interface NodeBundler<N extends NodeBundlerName | 'none', O = never> {
  name: N
  options: O
}

interface UnbundledOptions {
  "mainFile":string
  "moduleFormat": "esm" | 'commonjs'
}

interface Config {
  nodeBundler: NodeBundlerName | NodeBundler<NodeBundlerName> | NodeBundler<'none', UnbundledOptions>
  ...
}
```

Allowing direct names as a convenience for all current bundlers, as they do not have options (yet?). `nodeBundler: 'esbuild'`. But it could also be written like `nodeBundler: { name: 'esbuild' }`. The exception is the `none` bundler which has some mandatory options, so no shortcut is allowed.
This way by looking at the config or its types it is always clear which bundler is configured.

danez: > `function.js` is a source file, whereas `mainFile` is the distribution/compiled file that lives in the ZIP archive. Sometimes those two files will be the same, but not always. For example, when you're dealing with a TypeScript function, the source file will be something like `function.ts`, but your `mainFile` has to be a JavaScript because the Lambda runtime doesn't understand TypeScript.

I thought if bundling is disabled, we would expect `functions.js` to be the already pre-bundled version. In this case we do not need the source of the function or do we?

eduardoboucas: Hmm good point, maybe we don't need it.

src/config.ts
jackiewmacharia: It makes sense, I think it's better to keep it and not need it than not have it but need it later.

src/config.ts
danez: I will do this in a new PR. I started in this PR, but this actually touches a lot of points, so it might be easier to do it separately.

So in this PR lets go with the config we support currently.

danez:
@eduardoboucas I cannot add you as reviewer, because you are the creator. would you mind looking over basically your own code once more, and comment if you notice something?

danez (dismissed):

src/config.ts
```diff
@@ -19,16 +23,54 @@ export interface FunctionConfig {
...
+  configFileDirectories?: string[]
+  func: FunctionWithoutConfig
+  featureFlags: FeatureFlags
+}): Promise<FunctionConfig> => {
+  const fromConfig = getFromMainConfig({ config, func })
+
+  // We try to read from a function config file if the function directory is
+  // inside one of `configFileDirectories`.
+  const shouldReadConfigFile =
+    featureFlags.project_deploy_configuration_api_use_per_function_configuration_files &&
```
eduardoboucas, line 52: oof, this is a pretty long name for a feature flag! 😅

src/config.ts
```diff
@@ -19,16 +23,54 @@ export interface FunctionConfig {
...
+    configFileDirectories?.some((directory) => func.srcDir.startsWith(directory))
```
eduardoboucas, line 53: Is there the potential for the `startsWith` to catch false positives? For example:

- `configFileDirectories`: [`"foo/bar"`]
- `func.srcDir`: `"foo/bar-and-something-else"`

tests/main.js
```diff
@@ -2758,3 +2780,109 @@ test('listFunctionsFiles does not include wrong arch functions and warns', async
 
   console.warn.restore()
 })
+
+testMany(
```
eduardoboucas, line 2784: Should we assert that any configuration properties passed via the existing `config` object are kept even if there's a function-specific configuration file that doesn't overwrite it?

eduardoboucas: Could be nice to also assert that the property from the function-specific config takes priority.

eduardoboucas:
> would you mind looking over basically your own code once more, and comment if you notice something?

Done!

tests/main.js
danez: Added the tests

src/config.ts
danez: I changed the logic and use the package `is-path-inside` which ensures this works correctly. Also adjusted the test to cover for this. See new commits.

tests/main.js
```diff
@@ -2758,3 +2780,154 @@ test('listFunctionsFiles does not include wrong arch functions and warns', async
...
+
+    t.true(await pathExists(`${tmpDir}/internal-function.zip_out/blog/one.md`))
+    t.true(await pathExists(`${tmpDir}/root-function.zip_out/blog/one.md`))
+    t.false(await pathExists(`${tmpDir}/user-function.zip_out/blog/one.md`))
+    t.false(await pathExists(`${tmpDir}/not-internal.zip_out/blog/one.md`))
+  },
+)
+
+testMany(
+  'Keeps config for functions passed to zisi, but overwrites with config from json config',
```
eduardoboucas, line 2838: Mega nit! 🤓 

```suggestion
  'Keeps config for functions passed to ZISI, but overwrites with config from JSON config',
```

eduardoboucas:
LGTM!

danez:
Okay I consider this done. I'm not going to merge this yet, but will start to build the skip bundling based on this PR and then merge both together.

danez (approved):

PR #1053, opened 2022-03-07
chore: update delta-action

eduardoboucas:
⏳


danez:
I think this can be closed, as the change is already released in delta-action 3.0.1

eduardoboucas:
> I think this can be closed, as the change is already released in delta-action 3.0.1

Sounds good!

PR #1069, opened 2022-04-21
fix: adapt to new node:test module in node 18

Skn0tt:

#### Summary

Fixes tests on main. Read my PR to precinct upstream for more info: https://github.com/dependents/node-precinct/pull/108

Until precinct is fixed upstream, this PR fixes it on our end.

---

      passes our tests.

### DIFF
--- a/src/runtimes/node/bundlers/zisi/src_files.ts
+++ b/src/runtimes/node/bundlers/zisi/src_files.ts
@@ -93,16 +93,18 @@ const getDependencies = async function ({
 const paperwork = async (path: string) => {
   const modules = await precinct.paperwork(path, { includeCore: true })
   return modules.filter((moduleName) => {
+    if (moduleName.startsWith('node:')) {
+      return false
+    }
+
     // only require("node:test") refers to the
     // builtin, require("test") doesn't
     if (moduleName === 'test') {
       return true
     }
 
-    const normalisedModuleName = moduleName.startsWith('node:') ? moduleName.slice('node:'.length) : moduleName
-
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const isNativeModule = normalisedModuleName in (process as any).binding('natives')
+    const isNativeModule = moduleName in (process as any).binding('natives')
 
     return !isNativeModule
   })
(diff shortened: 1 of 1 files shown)


src/runtimes/node/bundlers/zisi/src_files.ts
```diff
@@ -87,6 +88,19 @@ const getDependencies = async function ({
   }
 }
 
+const paperwork = async (path: string) => {
+  const modules = await precinct.paperwork(path, { includeCore: true })
```
eduardoboucas, line 92: If I understand correctly, by using `includeCore: true` instead of `includeCore: false`, we're taking over the job of filtering the core modules, which you're doing with `isBuiltinModule`. I see the precinct seems to use [a different detection mechanism](https://github.com/dependents/node-precinct/blob/253d4b05c1d26ac662b76390046bc48d4814c5b2/index.js#L21), though.

Is there a risk this will cause regressions, with some modules suddenly being evaluated differently?

src/runtimes/node/bundlers/zisi/src_files.ts
```diff
@@ -87,6 +88,19 @@ const getDependencies = async function ({
   }
 }
 
+const paperwork = async (path: string) => {
```
eduardoboucas, line 91: Can we add a comment explaining that we're temporarily adding a step between us and precinct because of a specific issue?

src/runtimes/node/bundlers/zisi/src_files.ts
Skn0tt: that's a good point - we should stick to process.binding if possible. isBuiltinModule is using a [similar mechanism](https://github.com/sindresorhus/builtin-modules/blob/main/index.js) under the hood, but there could be subtle differences ...

src/runtimes/node/bundlers/zisi/src_files.ts
Skn0tt: done in https://github.com/netlify/zip-it-and-ship-it/pull/1069/commits/c75179fea6142e65f0facd3174625829d5a66408

src/runtimes/node/bundlers/zisi/src_files.ts
Skn0tt: done in https://github.com/netlify/zip-it-and-ship-it/pull/1069/commits/c75179fea6142e65f0facd3174625829d5a66408

eduardoboucas (approved):
🚀 🌕

PR #1083, opened 2022-05-09
feat: update precinct

Skn0tt:
This is a follow-up to #1069, where I commited a fix that's now [been merged](https://github.com/dependents/node-precinct/pull/108) into upstream precinct. This allows us to remove our fix in ZISI.


eduardoboucas (approved):
Nice one! I've updated the PR title to include `feat:` instead of `chore:` because otherwise this will not trigger a release.

PR #1084, opened 2022-05-09
chore: add "cypress passed" comment to integration test

Skn0tt:
When looking at https://github.com/netlify/zip-it-and-ship-it/pull/1080#issuecomment-1120830764, I was wondering "hm, does this mean I should take a look at this? 🤔". To make it obvious that this includes some e2e tests, I added the "cypress tests passed" line to the comment.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -88,5 +88,7 @@ jobs:
           comment: |
             ## Integration Test: netlify/netlify-plugin-nextjs
 
+            ✅ Cypress tests passed
+
             - **Deploy URL**: ${{ env.deploy_url }}
             - **Deploy logs**: ${{ env.deploy_log_url }}


Skn0tt:
btw, now that the integration test includes some cypress tests, what do you think about running the integration test on every PR instead of only on releases @eduardoboucas?

eduardoboucas:
> btw, now that the integration test includes some cypress tests, what do you think about running the integration test on every PR instead of only on releases @eduardoboucas?

Good question! The reason we're doing it on release PRs only is that it creates actual deployments on Netlify, which might be a bit overkill on all of the Renovate PRs, for example. It feels to me that doing it on releases is a way to guarantee that we don't ship a regression (which you've improved a lot by adding the Cypress step), without adding too much bloat to the day-to-day workflow.

What do you think?

Skn0tt:
Sounds good 👍 Just wanted to double-check.

eduardoboucas (approved):

PR #1086, opened 2022-05-17
fix(nft): correctly annotate transpilation errors in nft bundler

danez:

#### Summary

Add customErrorInfo to errors happening during transpilation of esm with esbuild in the nft bundler.
I had to loop the function name through to the actual `esbuild` call, to generate the error location properly.

---

      passes our tests.

### DIFF
--- a/src/runtimes/node/bundlers/nft/es_modules.ts
+++ b/src/runtimes/node/bundlers/nft/es_modules.ts
@@ -56,6 +56,7 @@ export const processESM = async ({
   fsCache,
   mainFile,
   reasons,
+  name,
 }: {
   basePath: string | undefined
   config: FunctionConfig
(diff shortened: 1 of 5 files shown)


eduardoboucas (approved):
Looks great! I left a super minor comment that we usually wouldn't care much about, but since you're new to the codebase I figured it's useful to align on those things.

Also, we tend to use the [feedback ladder](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/) in PR reviews, hence the `[dust]` prefix.

eduardoboucas (approved):
🚀 

Ah, a final comment about conventions: you're rightly using the `fix:` prefix to signal a patch update, but there's no need to use the `(nft)` namespace. We typically only do that for dependency updates ([example](https://github.com/netlify/zip-it-and-ship-it/pull/1081)), which are handled automatically by robots. 🤖

PR #1098, opened 2022-05-30
feat: include resolved included_files in response of `zipFunction`

danez:
#### Summary

Refs netlify/cli#4454

Return the resolved paths of `included_files` from the config. This will allow the cli to watch these files and rebuild the function on demand.

H30cfc3adb41641e497c1ce2f446b091aV


tests/main.js
```diff
@@ -381,6 +381,24 @@ testMany(
...
+    config: {
+      '*': {
+        nodeBundler: undefined,
+        includedFiles: ['included/*.js'],
+      },
+    },
+  })
+
+  t.true(Array.isArray(result.includedFiles))
+  t.true(/node-module-next-image[/\\]included[/\\]abc\.js/.test(result.includedFiles[0]))
```
eduardoboucas, line 399: Are you using a regular expression to catch Unix and Windows paths? If so, we've been using `unixify` to this effect:

```suggestion
  t.is(unixify(result.includedFiles[0]), 'node-module-next-image/included/abc.js')
```

src/runtimes/node/index.ts
```diff
@@ -108,6 +109,10 @@ const zipFunction: ZipFunction = async function ({
     bundlerWarnings,
     config,
     inputs,
+    includedFiles: await getFilteredPathsOfIncludedFiles(
```
eduardoboucas, line 112: Am I right in thinking that each bundler is already calling `getPathsOfIncludedFiles` and we're now adding a second call, at the runtime level? Would it make sense to have each bundler return this information to the runtime?

danez: The problem is, that for the esbuild and zisi bundler `getPathsOfIncludedFiles` is called inside `getSrcFiles`. In order to return included_files from the bundler without an additional call, I would need to change the return type of `getSrcFiles`. So instead of an array of paths return `{srcFiles:[], included_paths:[]}`. Do you think it is worth changing this?

eduardoboucas: Ah, I see! I'm okay with shipping as is for now and maybe create an issue to revisit later (or not revisit at all, if we don't think it's worth doing). Happy to defer to you.

tests/main.js
```diff
@@ -381,6 +381,24 @@ testMany(
   },
 )
 
+testMany('Includes includedFiles in the response of zipFunction', ['bundler_default'], async (options, t) => {
```
eduardoboucas, line 384: Do we want to include more bundlers than just `bundler_default`?

src/runtimes/node/index.ts
danez: It was less work than anticipated. I thought the impact of changing this would have been bigger. So now `getSrcFiles` returns an object.

tests/main.js
danez: I switched this around. It is still a regex, as I cannot match the whole path as it is environment dependent. But I use unixify now.

eduardoboucas (dismissed):
Great work! I wonder if there's anything we could/should do to make the terminology clearer, because we have `srcFiles` (which represents files identified by the bundler) and `includedFiles` (which represents files identified by the user), and that might not be obvious by the naming alone?

What do you think? Can we rename something to make it clearer? Or maybe some comments to help guide people in the right direction?

eduardoboucas:
Oh, I should've clarified that my questions above are non-blocking. I think it's fine to ship this as is, just wondering if there's anything we should do in a follow up issue/PR.

src/runtimes/node/bundlers/esbuild/src_files.ts
```diff
@@ -15,11 +15,12 @@ export const getSrcFiles: GetSrcFilesFunction = async ({ config, mainFile, plugi
...
   })
-  const includedPaths = filterExcludedPaths([...dependencyPaths, ...includedFilePaths], excludedPaths)
+  const srcFiles = filterExcludedPaths(dependencyPaths, excludedPaths)
+  const includedPaths = filterExcludedPaths(includedFilePaths, excludedPaths)
 
   return {
-    srcFiles: [...includedPaths, mainFile],
-    includedFiles: filterExcludedPaths(includedFilePaths, excludedPaths),
+    srcFiles: [...srcFiles, ...includedPaths, mainFile],
+    includedFiles: includedPaths,
```
danez, line 23: I just made some minor adjustments to evrey bundler, so that we do not iterate over includedFilePaths twice.

eduardoboucas: Nice! That also makes things a bit clearer, IMO.

eduardoboucas (approved):

danez:
Created #1099

PR #1105, opened 2022-06-07
fix: detect a wider range of scheduled functions

danez:
scheduled functions which are reassigned or not directly exported are now detected

This PR extends the functionality of the detection for the in-source-config of scheduled functions.

2 cases were not supported previously:

1. The schedule is not directly supplied as string to the `schedule()` helper but assigned to a variable `schedule(SCHEDULE, ...)`
2. The schedule helper is not directly exported (`export const handler = schedule()`) but instead assigned to a variable which is exported `export { handler }`

To solve both of these cases I opted to collect all variable bindings and their current value (only in the root scope). This is done by iterating through all nodes and collecting all declarations of variables and all reassigns. This method is memoized so that multiple calls (at the moment maximum 2) won't result in duplicated work.

Now, when the cron pattern in the schedule helper (1) or the export (2) is encountered as an identifier instead of the expected value, the code now can resolve the identifier to its value using the binding-map.

### DIFF
--- a/src/runtimes/node/parser/bindings.ts
+++ b/src/runtimes/node/parser/bindings.ts
@@ -20,7 +20,7 @@ const getBindingsFromNode = function (node: Statement, bindings: Bindings): void
     node.expression.type === 'AssignmentExpression' &&
     node.expression.left.type === 'Identifier'
   ) {
-    // The variable was reassigned, so lets store the new value
+    // The variable was reassigned, so let's store the new value
     bindings.set(node.expression.left.name, node.expression.right)
   } else if (node.type === 'ExportNamedDeclaration' && node.declaration?.type === 'VariableDeclaration') {
     // A `export const|let ...` creates a binding that can later be referenced again
--- a/src/runtimes/node/parser/bindings.ts
+++ b/src/runtimes/node/parser/bindings.ts
@@ -1,0 +1,55 @@
+import type { Expression, Statement, VariableDeclaration } from '@babel/types'
+
+type Bindings = Map<string, Expression>
+
+const getBindingFromVariableDeclaration = function (node: VariableDeclaration, bindings: Bindings): void {
+  node.declarations.forEach((declaration) => {
+    if (declaration.id.type === 'Identifier' && declaration.init) {
+      bindings.set(declaration.id.name, declaration.init)
+    }
+  })
+}
+
+// eslint-disable-next-line complexity
+const getBindingsFromNode = function (node: Statement, bindings: Bindings): void {
+  if (node.type === 'VariableDeclaration') {
+    // A variable was created, so create it and store the potential value
+    getBindingFromVariableDeclaration(node, bindings)
+  } else if (
+    node.type === 'ExpressionStatement' &&
+    node.expression.type === 'AssignmentExpression' &&
+    node.expression.left.type === 'Identifier'
+  ) {
+    // The variable was reassigned, so lets store the new value
+    bindings.set(node.expression.left.name, node.expression.right)
+  } else if (node.type === 'ExportNamedDeclaration' && node.declaration?.type === 'VariableDeclaration') {
+    // A `export const|let ...` creates a binding that can later be referenced again
+    getBindingFromVariableDeclaration(node.declaration, bindings)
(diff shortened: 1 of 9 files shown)


src/runtimes/node/parser/bindings.ts
```diff
@@ -0,0 +1,55 @@
...
+const getBindingsFromNode = function (node: Statement, bindings: Bindings): void {
+  if (node.type === 'VariableDeclaration') {
+    // A variable was created, so create it and store the potential value
+    getBindingFromVariableDeclaration(node, bindings)
+  } else if (
+    node.type === 'ExpressionStatement' &&
+    node.expression.type === 'AssignmentExpression' &&
+    node.expression.left.type === 'Identifier'
+  ) {
+    // The variable was reassigned, so lets store the new value
```
eduardoboucas, line 23:
```suggestion
    // The variable was reassigned, so let's store the new value
```

eduardoboucas, line 23: We're processing expressions in declaration order, so I'm assuming there's no risk of assigning the wrong binding when there are multiple reassignments?

src/runtimes/node/parser/bindings.ts
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  return bindings
+}
+
+export type BindingMethod = () => Bindings
+
+export const createBindingsMethod = function (nodes: Statement[]): BindingMethod {
+  // memoize the result for these nodes
+  let result: Bindings | null = null
```
eduardoboucas, line 48: [dust] Could we rely on `undefined` instead and simplify the expression? Or do you feel that using `null` is more semantically correct?

```suggestion
  let result: Bindings
```

src/runtimes/node/parser/bindings.ts
```diff
@@ -0,0 +1,55 @@
...
+  return () => {
+    if (!result) {
+      result = getAllBindings(nodes)
+    }
```
eduardoboucas, line 52: [dust] I hope you don't mind the small, stylistic nits. I'll try to point you in the direction we've been using in this codebase, but you're totally free to ignore.

```suggestion
    }

```

eduardoboucas (commented):
Looks really good. I left some comments and questions, mostly related to the fact that traversing ASTs is tricky and we should go the extra mile to make sure stuff is easy to understand by our future selves!

src/runtimes/node/parser/exports.ts
```diff
@@ -66,16 +76,36 @@ const getMainExportFromESM = (node: Statement) => {
...
+
+const getExportsFromBindings = (
+  specifiers: (ExportSpecifier | ExportDefaultSpecifier | ExportNamespaceSpecifier)[],
+  getAllBindings: BindingMethod,
+) => {
+  const specifier = specifiers.find(
+    ({ type, exported }) =>
+      type === 'ExportSpecifier' &&
+      ((exported.type === 'Identifier' && exported.name === 'handler') ||
+        (exported.type === 'StringLiteral' && exported.value === 'handler')),
```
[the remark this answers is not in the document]
danez: I will extract this.

It covers the case `export { a as "handler" }`

src/runtimes/node/parser/bindings.ts
eduardoboucas: Yes, absolutely! That would be great. ❤️

src/runtimes/node/parser/exports.ts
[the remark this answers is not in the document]
eduardoboucas: Ah, makes sense.

src/runtimes/node/parser/bindings.ts
```diff
@@ -0,0 +1,56 @@
...
+  })
+
+  return bindings
+}
+
+export type BindingMethod = () => Bindings
+
+export const createBindingsMethod = function (nodes: Statement[]): BindingMethod {
+  // memoize the result for these nodes
+  let result: Bindings
```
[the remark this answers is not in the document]
danez: will tackle these with the eslint rule.

PR #1107, opened 2022-06-09
chore: enforce newline before return and after variable declarations

danez:
#### Summary

As discussed in #1105, this new eslint rule enforces newlines before returns and after variable delcarations. I did not enable it for the tests, as that would change a lot.

Might be worth upstreaming this to the netlify eslint config?
The rule is already upstream but has no config https://github.com/netlify/eslint-config-node/blob/main/.eslintrc.cjs#L114 and according to the docs 

> This rule does nothing if no configurations are provided.

https://eslint.org/docs/rules/padding-line-between-statements


eduardoboucas:
> Might be worth upstreaming this to the netlify eslint config?

Definitely! I would suggest opening a PR and adding an action item to the next developer tooling guild sync.

eduardoboucas (approved):
😍😍😍

PR #1119, opened 2022-06-21
fix: handle invalid `files` entry in package.json

danez:
#### Summary

Fixes netlify/cli#4612

Add more defensive code so that invalid `files` properties don't make everything explode. See https://github.com/math-utils/hamming-distance/pull/4


eduardoboucas (approved):
🚀

PR #1121, opened 2022-06-23
fix: always build for amd64 when building go functions

danez:
I noticed that we build for whatever architecture the current host has. Which when for example building locally on M1 MacBook probably creates Linux/arm64 binary which would be wrong, but even if not I feel saver by defining it explicitly.

Ref: https://github.com/netlify/pillar-runtime/issues/231


eduardoboucas (approved):
Maybe it'd be nice to make this configurable in the future, in case we start using a different arch in Lambda (like ARM).

This is great for now, though! ✨

PR #1122, opened 2022-06-23
chore: report unused eslint disable statements

danez:
I just ran into a case where I added a disable comment for eslint but then fixed the code, but nearly forgot to remove it again now that the eslint issue was fixed. So I figured it might be a good idea to check for unused disable satements when running eslint.


eduardoboucas (approved):
Nice!

PR #1124, opened 2022-06-24
feat: detect invalid binaries and show warning

danez:
#### Summary

This now detects binaries not built for Linux/amd64 and prints a warning in this case. 

Fixes netlify/pillar-runtime/issues#231

baboon copy


src/runtimes/detect_runtime.ts
```diff
@@ -20,15 +22,27 @@ export const detectBinaryRuntime = async function ({
     // We're using the Type Assertion because the `cachedReadFile` abstraction
     // loses part of the return type information. We can safely say it's a
     // Buffer in this case because we're not specifying an encoding.
-    const binaryType = detect(buffer as Buffer)
+    const binaryInfo = detect(buffer as Buffer)
+
+    if (!isValidFunctionBinary(binaryInfo)) {
+      console.warn(`
```
eduardoboucas, line 28: zip-it-and-ship-it can be invoked as a binary, which [prints a JSON string to stdout](https://github.com/netlify/zip-it-and-ship-it/blob/main/src/bin.ts#L20). For this reason, we don't typically print any additional messages because that'll technically make the output incorrect. To get around this, we've been adding properties to the output and then reading them from Netlify Build and/or CLI and show warning/error messages accordingly.

This is something we should revisit, because I don't think we're actually using the binary version anywhere, but I still wanted to mention it.

eduardoboucas: https://github.com/netlify/cli/blob/2c58d10f77eb4c86daeb267925d5d98bbc867b70/src/utils/deploy/hash-fns.js#L113-L115 and https://github.com/netlify/cli/blob/89452281c9cfb1aff19991a93ed2638d0bc03734/src/utils/deploy/deploy-site.js#L101-L112 are an example of this.

eduardoboucas: Ah, I see. Agreed, let's not overcomplicate things! I think it's fine to ship this as is, and then look into removing the binary separately.

danez: I changed the logic slightly to simply not print the warning in cli mode.

src/bin.ts
```diff
@@ -7,13 +7,19 @@ import { hideBin } from 'yargs/helpers'
...
+}
+
 // CLI entry point
 const runCli = async function () {
   // @ts-expect-error TODO: `destFolder` and  `srcFolder` are not being passed
   // back from `parseArgs()`.
   const { destFolder, srcFolder, ...options } = parseArgs()
 
   try {
+    global.ZISI_CLI = true
```
mraerino, line 22: does this need to be a global?

danez: Not necessarily, but the alternative would be to pass the flag through the complete function hierarchy. And as the binary of zisi seems it might be deprecated and removed at some point, this was by far the more straightforward solution.

mraerino (approved):

PR #1134, opened 2022-07-05
fix: add more error info on missing functions folders

danez:

#### Summary

For https://github.com/netlify/pod-compute/issues/108

Throws underlying errors if the error is not a FileNotFound Error.

### DIFF
--- a/src/utils/fs.ts
+++ b/src/utils/fs.ts
@@ -48,11 +48,14 @@ export const listFunctionsDirectories = async function (srcFolders: string[]) {
   const validDirectories = filenamesByDirectory
     .map((result) => {
       if (result.status === 'rejected') {
-        if (result.reason instanceof Error && (result.reason as NodeJS.ErrnoException).code !== 'ENOENT') {
-          throw result.reason
+        // If the error is about `ENOENT` (FileNotFound) then we only throw later if this happens
+        // for all directories.
+        if (result.reason instanceof Error && (result.reason as NodeJS.ErrnoException).code === 'ENOENT') {
+          return null
         }
 
-        return null
+        // In any other error case besides `ENOENT` throw immediately
+        throw result.reason
       }
 
       return result.value
--- a/src/utils/fs.ts
+++ b/src/utils/fs.ts
@@ -41,34 +41,36 @@ export const safeUnlink = async (path: string) => {
 // filenames within those directories, if at least one of the directories
 // exists. If not, an error is thrown.
 export const listFunctionsDirectories = async function (srcFolders: string[]) {
-  const filenamesByDirectory = await Promise.all(
-    srcFolders.map(async (srcFolder) => {
-      try {
-        const filenames = await listFunctionsDirectory(srcFolder)
+  const filenamesByDirectory = await Promise.allSettled(
+    srcFolders.map((srcFolder) => listFunctionsDirectory(srcFolder)),
+  )
+  const errorMessages: string[] = []
+  const validDirectories = filenamesByDirectory
+    .map((result) => {
+      if (result.status === 'rejected') {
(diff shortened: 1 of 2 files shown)


src/utils/fs.ts
```diff
@@ -41,34 +41,34 @@ export const safeUnlink = async (path: string) => {
...
+${errorMessages.join('\n')}`)
   }
 
   return validDirectories.flat()
 }
 
-export const listFunctionsDirectory = async function (srcFolder: string) {
-  try {
-    const filenames = await fs.readdir(srcFolder)
+const listFunctionsDirectory = async function (srcFolder: string) {
```
danez, line 68: I removed the export as it is nowhere used outside this file, and also is not a Public API.

src/utils/fs.ts
```diff
@@ -41,34 +41,34 @@ export const safeUnlink = async (path: string) => {
...
+  const filenames = await fs.readdir(srcFolder)
 
-    return filenames.map((name) => join(srcFolder, name))
-  } catch {
-    throw new Error(`Functions folder does not exist: ${srcFolder}`)
-  }
```
danez, line 71: As we do not export it we do not need to catch anymore, because the error will be matched in the function above.

src/utils/fs.ts
```diff
@@ -41,34 +41,34 @@ export const safeUnlink = async (path: string) => {
...
-  const validDirectories = filenamesByDirectory.filter(nonNullable)
+
+      return result.value
+    })
+    .filter(nonNullable)
 
   if (validDirectories.length === 0) {
-    throw new Error(`Functions folder does not exist: ${srcFolders.join(', ')}`)
+    throw new Error(`Functions folders do not exist: ${srcFolders.join(', ')}
+${errorMessages.join('\n')}`)
```
eduardoboucas, line 62: I understand that you're doing this to capture additional information about the error for debugging purposes, but I worry this will add noise and confusion to an error message that was short and to the point. Any thoughts on how we could access this information without adding noise to the actual message we present to users?

danez: But the error message might also not be accurate. If for whatever reason there is an IO error, we still show this message, even though It might be a different error. (I think this is what actually happened in https://github.com/netlify/pod-compute/issues/108, because in the build logs it first prints the functions it found in these folders and then an error that the functions folders do not exist.)

But to your point: We could introduce a logLevel, that is by default different in build and cli.

danez: Other solution would be to check if the error is `fileNotFound` for our nice error and otherwise let the error bubble.

danez: Done

src/utils/fs.ts
```diff
@@ -41,34 +41,36 @@ export const safeUnlink = async (path: string) => {
...
-      try {
-        const filenames = await listFunctionsDirectory(srcFolder)
+  const filenamesByDirectory = await Promise.allSettled(
+    srcFolders.map((srcFolder) => listFunctionsDirectory(srcFolder)),
+  )
+  const errorMessages: string[] = []
+  const validDirectories = filenamesByDirectory
+    .map((result) => {
+      if (result.status === 'rejected') {
+        if (result.reason instanceof Error && (result.reason as NodeJS.ErrnoException).code !== 'ENOENT') {
```
danez, line 51: For some reason Error does not contain `code`, so this is the workaround.

eduardoboucas, line 51: Not a big deal, but I think this would be a bit easier to understand if we inverted the logic to treat `ENOENT` as the special case — i.e. if the error is `ENOENT` we return `null`, otherwise we throw. Might also be good to add a comment explaining the intent.

eduardoboucas (approved):

PR #1173, opened 2022-08-08
feat: new bundler that allows to skip bundling

danez:
#### Summary

This ads a new bundler that does not do anything. It is to allow skipping the bundling.

This branch is targeting PR #1030 because it is based on the Deployment config branch. Needs to be retargeted before merging.

I added the new bundler wherever it was possible in the tests.

hippo-1


src/runtimes/node/bundlers/types.ts
```diff
@@ -5,7 +5,25 @@ import type { FeatureFlag, FeatureFlags } from '../../../feature_flags.js'
...
-export type NodeBundlerName = 'esbuild' | 'esbuild_zisi' | 'nft' | 'zisi'
+export const enum NodeBundlerType {
+  ESBUILD = 'esbuild',
+  ESBUILD_ZISI = 'esbuild_zisi',
+  NFT = 'nft',
+  ZISI = 'zisi',
+  SKIP = 'skip',
+}
+
+export interface NoOptionsConfig<
```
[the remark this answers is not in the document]
danez: Something like:
```ts
interface FunctionConfig<Bundler extends NodeBundlerType = NodeBundlerType> {
  nodeBundler?: Bundler,
  nodeBundlerOptions: Bundler extends NodeBundlerType.SKIP ? SkipConfig: never 
}
```

Problem here is that nodeBundlerOptions would be either mandatory or not mandatory for all cases.
This could be worked around with:

```ts
interface BaseFunctionConfig {
...
}

type FCWithoutOptions = BaseFunctionConfig & { nodeBundler: Exclude<NodeBundlerType, NodeBundlerType.SKIP> }
type FCWithSkipOptions = BaseFunctionConfig & { nodeBundler: NodeBundlerType.SKIP, nodeBundlerOptions: SkipOptions }

type FunctionConfig = FCWithoutOptions | FCWithSkipOptions
```

I personally feel this is similar complicated, just in a different location. And it is less extensible in case one of the other Bundlers starts to have options.

What I did now is completely move the logic for the nodeBundler field into the bundlers/types.ts file, so it is in one place and also tried to simplify it a little bit.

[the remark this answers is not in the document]
eduardoboucas: > Problem here is that nodeBundlerOptions would be either mandatory or not mandatory for all cases.

What if it's never mandatory? The worst thing that could happen is that someone sets `nodeBundler: "custom"` and doesn't provide a `nodeBundlerOptions` object so we don't include any files in the bundle? Given that this will be used for very specific use cases and advanced integrations, I think that would be an okay tradeoff if it means simplifying the implementation.

I can also see an argument for having rigid types, so I'm happy to defer to you.

src/runtimes/node/bundlers/none/index.ts
```diff
@@ -0,0 +1,109 @@
...
+import { BundleFunction, GetSrcFilesFunction, NodeBundlerType } from '../types.js'
+
+/**
+ * This bundler is a simple no-op bundler, that does no bundling at all.
+ * It returns the detected moduleFormat and the mainFile + includedFiles from the config.
+ */
+
+/**
+ * Mimics the logic from Node.js, as functions with this bundler will be executed as is in Node.js.
+ */
```
eduardoboucas, line 19: I think I understand what you mean with this comment, but I find the formulation a bit confusing. Here's a suggestion with a link to the Node.js docs, but feel free to discard if you don't think it helps.

```suggestion
// Uses the same algorithm as the ESM_FILE_FORMAT resolver from Node.js to get
// the right module format for a function based on its main file.
// See https://nodejs.org/api/esm.html#resolver-algorithm-specification.
```

src/runtimes/node/bundlers/none/index.ts
```diff
@@ -0,0 +1,109 @@
...
+    srcPath,
+    stat,
+  })
+  const dirnames = srcFiles.map((filePath) => normalize(dirname(filePath)))
+  const moduleFormat = await getModuleFormat(mainFile)
+  const nodeSupport = getNodeSupportMatrix(config.nodeVersion)
+
+  if (moduleFormat === ModuleFormat.ESM && !nodeSupport.esm) {
+    throw new FunctionBundlingUserError(
+      `Entrypoint file is an ESM module, but the Node.js version in the config (${config.nodeVersion}) does not support ESM. ESM is supported as of version 14 of Node.js.`,
```
eduardoboucas, line 88: I think we should either use "ES Module" or "ESM", not "ESM module".

```suggestion
      `Function file is an ES module, which the Node.js version specified in the config (${config.nodeVersion}) does not support. ES modules are supported as of version 14 of Node.js.`,
```

tests/fixtures/config-files-invalid-version/my-function-1/my-function-1.js
```diff
@@ -1 +1 @@
-export const handler = () => true
+module.exports.handler = () => true
```
eduardoboucas, line 1: Why the change in these fixtures?

danez: Right I changed them back.

eduardoboucas (approved):

PR #1179, opened 2022-08-16
feat: convert Runtime, NodeBundler and ModuleFormat to enums

danez:

#### Summary

This is split out from #1173 

This contains only the change to use enums for Runtime, Bundler, and ModuleFormat.

To be very semver correct, I marked this as breaking, because if someone uses this with typescript the update will be breaking for the types.


eduardoboucas (approved):
Nice! ✨

PR #1187, opened 2022-08-22
feat: add support for layers

eduardoboucas:
#### Summary

Introduces a new `layers` configuration property at the function level which specifies the names of Lambda Layers to interact with at runtime. On every layer, zip-it-and-ship-it will call a `getHandler` method to wrap the underlying handler with new functionality. This PR adds a test that shows a concrete example of that flow with 2 layers.


src/runtimes/node/utils/entry_file.ts
```diff
@@ -8,11 +9,21 @@ export interface EntryFile {
   filename: string
 }
 
-const getEntryFileContents = (mainPath: string, moduleFormat: string) => {
+const getEntryFileContents = (mainPath: string, moduleFormat: string, layers: string[]) => {
   const importPath = `.${mainPath.startsWith('/') ? mainPath : `/${mainPath}`}`
 
   if (moduleFormat === ModuleFormat.COMMONJS) {
-    return `module.exports = require('${importPath}')`
+    if (layers.length === 0) {
```
eduardoboucas, line 16: We could get rid of this `if` and just use the code below for everything, since `getLayersBootstrap()` will return an empty array if there are no layers. Still, in the spirit of reducing the blast radius, I want to keep the existing code path unmodified. We can have a feature flag in Netlify Build that decides whether to populate the `layers` config property.

src/runtimes/node/utils/entry_file.ts
```diff
@@ -8,11 +9,21 @@ export interface EntryFile {
   filename: string
 }
 
-const getEntryFileContents = (mainPath: string, moduleFormat: string) => {
+const getEntryFileContents = (mainPath: string, moduleFormat: string, layers: string[]) => {
   const importPath = `.${mainPath.startsWith('/') ? mainPath : `/${mainPath}`}`
 
   if (moduleFormat === ModuleFormat.COMMONJS) {
```
eduardoboucas, line 15: We'll need to implement a similar bootstrap code for ESM functions. We're not generating any of those at the moment, so we can do that in a follow up PR.

src/runtimes/node/utils/zip.ts
```diff
@@ -116,7 +120,7 @@ const createZipArchive = async function ({
   const userNamespace = hasEntryFileConflict ? DEFAULT_USER_SUBDIRECTORY : ''
 
   if (needsEntryFile) {
```
eduardoboucas, line 122: There's a caveat with the current PR in that we won't generate the layer code if `needsEntryFile` is `false`. We'll need to figure that out separately.

danez:
Is this for plugins? Or for users? I don't see the use case for this yet.

tests/main.js
```diff
@@ -2815,3 +2815,62 @@ testMany('None bundler emits esm with default nodeVersion', ['bundler_none'], as
 
   t.is(originalFile, bundledFile)
 })
+
+// The paths for the layer files are Unix only. Since we'll actually only use
+// layers in production (which is always Linux), we skip this test on Windows.
+if (platform !== 'win32') {
```
danez, line 2821: So we don't want to/can use this in `netlify dev`?

eduardoboucas: Hmm, good point. We probably will. I'll make it work on Windows. 👍🏻

eduardoboucas: Done!

danez (approved):

PR #1190, opened 2022-08-26
feat: output cjs extension for bundles and entry-files

danez:
### Summary

Fixes https://github.com/netlify/pod-compute/issues/149

This is behind a FF.

With FF enabled:
* esbuild outputs the bundle with cjs extension
* entry-file will be generated with `cjs` extension

Independent of FF we now check for file collisions with all possible entry-files `js, cjs, mjs` and not only with `js`. This is right now not as important as `js` takes precedence over `mjs and cjs` in AWS lambda, but once the FF is on this is critical as we generate the entry file with `cjs` extension but a user-non-main-file could be `func-name.js` and AWS would actually pick this one up instead of our entry-file.

Once we move to having a unique entry file like `__netlifyHandler.(c|m)js` we can clean this all up a lot.

For which functions will there be a hash change (with ff on):

- All functions that use the `*.cjs` file extension for the source function file

### DIFF
--- a/src/runtimes/node/utils/entry_file.ts
+++ b/src/runtimes/node/utils/entry_file.ts
@@ -20,7 +20,7 @@ const getEntryFileContents = (mainPath: string, moduleFormat: string) => {
   return `export { handler } from '${importPath}'`
 }
 
-// They are also in the order that lambda will try to find the entry point
+// They are also in the order that AWS Lambda will try to find the entry point
 const POSSIBLE_LAMBDA_ENTRY_EXTENSIONS = ['.js', '.mjs', '.cjs']
 
 export const isEntryFile = (
--- a/src/runtimes/node/utils/entry_file.ts
+++ b/src/runtimes/node/utils/entry_file.ts
@@ -1,18 +1,16 @@
-import { ModuleFormat } from './module_format.js'
+import { basename, extname, resolve } from 'path'
+
+import type { FeatureFlags } from '../../../feature_flags.js'
+
+import { getFileExtensionForFormat, ModuleFormat } from './module_format.js'
 import { normalizeFilePath } from './normalize_path.js'
 
-export const getEntryFile = ({
-  commonPrefix,
-  mainFile,
-  moduleFormat,
-  userNamespace,
-}: {
-  commonPrefix: string
-  mainFile: string
-  moduleFormat: ModuleFormat
-  userNamespace: string
-}) => {
-  const mainPath = normalizeFilePath({ commonPrefix, path: mainFile, userNamespace })
+export interface EntryFile {
+  contents: string
+  filename: string
+}
+
+const getEntryFileContents = (mainPath: string, moduleFormat: string) => {
   const importPath = `.${mainPath.startsWith('/') ? mainPath : `/${mainPath}`}`
 
   if (moduleFormat === ModuleFormat.COMMONJS) {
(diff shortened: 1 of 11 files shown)


src/runtimes/node/utils/entry_file.ts
```diff
@@ -12,28 +14,72 @@ const getEntryFileContents = (mainPath: string, moduleFormat: string) => {
   const importPath = `.${mainPath.startsWith('/') ? mainPath : `/${mainPath}`}`
 
   if (moduleFormat === ModuleFormat.COMMONJS) {
-    return `module.exports = require('${importPath}')`
+    return `module.exports=require('${importPath}')`
```
eduardoboucas, line 17: Why are we changing this? Note that this will start creating functions with a different SHA for the same input, which means creating new Lambdas unnecessarily.

src/runtimes/node/utils/entry_file.ts
```diff
@@ -12,28 +14,72 @@ const getEntryFileContents = (mainPath: string, moduleFormat: string) => {
...
   }
 
-  return `export { handler } from '${importPath}'`
+  return `export{handler}from'${importPath}'`
 }
 
+// They are also in the order that lambda will try to find the entry point
+const POSSIBLE_LAMBDA_ENTRY_EXTENSIONS = ['.js', '.mjs', '.cjs']
+
+export const isEntryFile = (
```
eduardoboucas, line 26: It's not obvious to me why we need a function that checks whether a file is the entry file, since theoretically we're in control of specifying what the entry file is? Could you add a comment to make the intent more explicit?

src/runtimes/node/utils/entry_file.ts
```diff
@@ -12,28 +14,72 @@ const getEntryFileContents = (mainPath: string, moduleFormat: string) => {
...
+  },
+) =>
+  POSSIBLE_LAMBDA_ENTRY_EXTENSIONS.some((extension) => {
+    const entryFilename = getEntryFileName({ extension, filename })
+    const entryFilePath = resolve(basePath, entryFilename)
+
+    return srcFiles.some((srcFile) => srcFile === entryFilePath && srcFile !== mainFile)
+  })
+
+const getEntryFileName = ({ extension, filename }: { extension: string; filename: string }) =>
```
eduardoboucas, line 62: I think a comment here would also help.

src/runtimes/node/utils/entry_file.ts
danez: Reverted, I though might be nice to save some bytes and because it is generated there is no need for codestyle, but I haven't thought about the hash implications. Good catch

src/runtimes/node/utils/entry_file.ts
danez: Added a comment.

src/runtimes/node/utils/entry_file.ts
```diff
@@ -1,23 +1,88 @@
...
+  }
+
+  return `export { handler } from '${importPath}'`
+}
+
+// They are in the order that AWS Lambda will try to find the entry point
+const POSSIBLE_LAMBDA_ENTRY_EXTENSIONS = [ModuleFileExtension.JS, ModuleFileExtension.MJS, ModuleFileExtension.CJS]
+
+// checks if the file is considered a entry-file in AWS Lambda
+export const isNamedLikeEntryFile = (
```
eduardoboucas, line 27: Any reason to be using a combination of normal arguments and named parameters here?

tests/main.test.ts
```diff
@@ -1023,6 +1023,32 @@ describe('zip-it-and-ship-it', () => {
     },
   )
 
+  // TODO can be merged with the above once the FF is on `zisi_output_cjs_extension`
+  testMany(
```
eduardoboucas, line 1027: No need to use `testMany` if you're just testing a single bundler.

tests/main.test.ts
```diff
@@ -2445,4 +2471,81 @@ describe('zip-it-and-ship-it', () => {
     expect(result.stderr).not.toContain('Dynamic require of "path" is not supported')
     expect(result).not.toBeInstanceOf(Error)
   })
+
+  testMany(
+    'bundlers emit entry file with cjs extension when FF on',
```
eduardoboucas, line 2476: Nit for consistency.

```suggestion
    'Emits entry file with .cjs extension when `zisi_output_cjs_extension` flag is on',
```

tests/main.test.ts
```diff
@@ -2445,4 +2471,81 @@ describe('zip-it-and-ship-it', () => {
...
+
+      if (options.getCurrentBundlerName() === 'esbuild') {
+        // nft does not create an entry here because the main file is already an entrypoint
+        const bundledFile1 = await importFunctionFile(join(tmpDir, 'func1.cjs'))
+        expect(bundledFile1.handler()).toBe(true)
+      }
+    },
+  )
+
+  testMany('bundlers keep cjs file extension', ['bundler_default', 'bundler_nft'], async (options) => {
```
eduardoboucas, line 2502: Nit for consistency.

```suggestion
  testMany('Keeps .cjs extension', ['bundler_default', 'bundler_nft'], async (options) => {
```

tests/main.test.ts
```diff
@@ -2445,4 +2471,81 @@ describe('zip-it-and-ship-it', () => {
...
+    const bundledFile2 = await importFunctionFile(join(tmpDir, 'func2.cjs'))
+    const bundledFile3 = await importFunctionFile(join(tmpDir, 'index.cjs'))
+
+    expect(bundledFile1.handler()).toBe(true)
+    expect(bundledFile2.handler()).toBe(true)
+    expect(bundledFile3.handler()).toBe(true)
+  })
+
+  testMany(
+    'bundlers do not create cjs entry file if entry with js extension is already present',
```
eduardoboucas, line 2524: Nit for consistency.

```suggestion
    'Does not create .cjs entry file if entry with .js extension is already present',
```

eduardoboucas (approved):
LGTM! Left a few non-blocking comments.

src/runtimes/node/utils/entry_file.ts
danez: My logic was that using a named param makes it more readable. So `file` is part of the message `is_namedLikeEntryFile(file, additional_options)`

Same with `conflictsWithEntryFile`, the title relates to the named param  `_conflictsWithEntryFile(srcFiles, additional_options)`

If it is mangled into the options then I would be wondering which of the params is actually `namedLikeEntryFile`.

tests/main.test.ts
danez: Because this potentially will be merged again with the test above, and then be `testMany` again I would like to leave it as that.

src/runtimes/node/utils/entry_file.ts
eduardoboucas: Sounds good!

eduardoboucas (approved):

PR #1214, opened 2022-09-29
feat: zisi system logging

jackiewmacharia:

#### Summary

This PR adds system logging to ZISI

Related to https://github.com/netlify/pod-compute/issues/196 and https://github.com/netlify/pod-compute/issues/246

Depends on https://github.com/netlify/build/pull/4547

---

      passes our tests.


src/zip.ts
```diff
@@ -89,6 +96,8 @@ export const zipFunctions = async function (
         stat: func.stat,
       })
 
+      zippedFunctionOutput.push({ name: func.name, config: func.config, featureFlags: functionFlags })
```
eduardoboucas, line 99: Is there a reason to buffer the logs as opposed to emitting one per function?

eduardoboucas: I think emitting separate logs would make the ZISI code a bit simpler, as you don't need to buffer the messages, and would protect us a bit from super-long log lines. It's true that these can get intertwined with other logs, but I personally don't think that's a huge deal. You can always adjust your Humio query to filter them out.

jackiewmacharia: Done. This is what an individual log looks like now:

```
Function details: {
  "name": "hello-world",
  "config": {
    "includedFilesBasePath": "/Users/jackie/src/playground/edge-functions-examples",
    "zipGo": true
  },
  "featureFlags": {
    "buildRustSource": false,
    "parseWithEsbuild": false,
    "traceWithNft": false,
    "zisi_pure_esm": false,
    "zisi_pure_esm_mjs": false,
    "project_deploy_configuration_api_use_per_function_configuration_files": true
  }
}
```

jackiewmacharia: Is there anything more we can do with ZISI logging that could help with how long bundling takes?

tests/unit/utils/logger.js
```diff
@@ -0,0 +1,63 @@
+// adapted from https://github.com/netlify/edge-bundler/blob/main/node/logger.test.ts
+
+const test = require('ava')
```
danez, line 3: This now needs to be converted to `vitest`. See any of the other tests.

package.json
```diff
@@ -76,6 +76,7 @@
     "p-map": "^4.0.0",
     "path-exists": "^4.0.0",
     "precinct": "^9.0.1",
+    "pretty-ms": "^7.0.0",
```
danez, line 79: You can also use v8 now

danez (changes requested):

package.json
jackiewmacharia: v8 doesn't work with ZISI right now because it's not a module and converting it to one at this time would require a host of lots of other changes.

jackiewmacharia: Sorry! Just seen the ESM work, will move to v8, thank you!

tests/unit/utils/logger.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  // no-op
+}
+
+afterEach(() => {
+  // Restoring global `console.log`.
+  console.log = consoleLog
+})
+
+test('Prints user logs to stdout', () => {
+  const mockConsoleLog = stub()
```
danez, line 18:
```suggestion
  const mockConsoleLog = vi.fn()
```

tests/unit/utils/logger.test.ts
```diff
@@ -0,0 +1,61 @@
+import { stub } from 'sinon'
+import { afterEach, test, expect } from 'vitest'
```
danez, line 2:
```suggestion
import { afterEach, test, expect, vi } from 'vitest'
```

Then we can again remove sinon.

tests/unit/utils/logger.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  const mockConsoleLog = stub()
+  console.log = mockConsoleLog
+
+  const logger1 = getLogger(noopLogger, true)
+  const logger2 = getLogger(noopLogger, false)
+
+  logger1.user('Hello with `debug: true`')
+  logger2.user('Hello with `debug: false`')
+
+  expect(mockConsoleLog.callCount).toBe(2)
```
danez, line 27:
```suggestion
  expect(mockConsoleLog).toHaveBeenCalledTimes(2)
```

tests/unit/utils/logger.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  expect(mockConsoleLog.firstCall.firstArg).toBe('Hello with `debug: true`')
+  expect(mockConsoleLog.secondCall.firstArg).toBe('Hello with `debug: false`')
```
danez, line 29:
```suggestion
  expect(mockConsoleLog).toHaveBeenNthCalledWith(1, 'Hello with `debug: true`')
  expect(mockConsoleLog).toHaveBeenNthCalledWith(2, 'Hello with `debug: false`')
```

danez (changes requested):
We do not need sinon anymore, we can use vitest directly.

src/zip.ts
```diff
@@ -167,6 +192,17 @@ export const zipFunction = async function (
...
     stat: stats,
   })
+  const durationNs = endTimer(startIntervalTime)
+  const durationMs = roundTimerToMillisecs(durationNs)
+  const logObject = {
+    name,
+    config,
+    featureFlags: functionFlags,
+    // e.g 45m 48s 373ms
+    duration: prettyMilliseconds(durationMs, { formatSubMilliseconds: true }),
```
eduardoboucas, line 202: I'm curious about your reasoning for using a human-readable value here. Since system logs are for our own usage only, it feels like a machine-friendly value (like the value in milliseconds or nanoseconds) would be more useful, because we could more easily extract that into a variable in Humio using regex and manipulate it in different ways.

jackiewmacharia: I got [this comment](https://github.com/netlify/pod-compute/issues/246#issue-1420815403) mixed up and read it as though we still needed it human friendly. I've changed it to use durationMs.

src/zip.ts
```diff
@@ -88,6 +96,16 @@ export const zipFunctions = async function (
         srcPath: func.srcPath,
         stat: func.stat,
       })
+      const durationNs = endTimer(startIntervalTime)
+      const durationMs = roundTimerToMillisecs(durationNs)
+      const logObject = {
+        name: func.name,
+        config: func.config,
+        featureFlags: functionFlags,
+        durationMs,
```
eduardoboucas, line 105: Is there any reason for us not to use nanoseconds here? Feels a bit strange that we're reaching for the high-resolution time for the extra precision and then losing it when emitting the log.

jackiewmacharia: I'm thinking during an investigation, what would be faster for us to compute/make sense of at a glance? i.e would it be  easier to see 600000ms as 10min or 600000000000ns as 10min? Initial issue [request](https://github.com/netlify/pod-compute/issues/246#issue-1420815403) was also for milliseconds. 

The conversion from nanoseconds is because the [process.hrtime](https://nodejs.org/api/process.html#processhrtimetime) operation gives us time in nanoseconds.

eduardoboucas: We're talking about the duration of bundling individual functions here, not about the entire bundling step. So 10 minutes is really not the order of magnitude we should be dealing with. Small functions should be processed really quickly, and we might actually be losing precision by converting to milliseconds.

jackiewmacharia: Okay, PR updated.

eduardoboucas (approved):
🚀

danez (approved):

PR #1221, opened 2022-10-14
feat: Make additional function attributes available via functions manifest

jackiewmacharia:

#### Summary

Issue: https://github.com/netlify/pod-compute/issues/192
Makes displayName, nodeBundler and isInternalFunction available to the API

---

      passes our tests.


src/runtimes/constants.ts
```diff
@@ -0,0 +1 @@
+export const FUNCTIONS_INTERNAL_DIR = '.netlify/internal-functions'
```
eduardoboucas, line 1: There's something about this that doesn't feel right to me. When we added support for the internal functions directory, we made the decision to keep that abstraction outside of zip-it-and-ship-it, such that [it only receives a list of directories](https://github.com/netlify/zip-it-and-ship-it/blob/main/src/zip.ts#L50) without knowing anything about them (i.e. which ones contain user-generated or framework-generated functions).

With this change, we're making zip-it-and-ship-it aware of the concept of internal functions, but only partially. On the one hand, we're giving it a list of directories without conveying the type, and on the other hand we're expecting zip-it-and-ship-it to know what the internal directory is.

In my mind, it would make more sense to add a parameter specifically for the internal functions directory. If found, it's added to the list of directories to be traversed and it will also be used to populate the `isInternalFunction` property.

Does that make sense?

src/utils/check_is_internal_function.ts
```diff
@@ -0,0 +1,8 @@
+import { FUNCTIONS_INTERNAL_DIR } from '../runtimes/constants.js'
+
+export const checkIsInternalFunction = (srcDir = '') => {
+  const BACKSLASH_REGEXP = /\\/g
+
+  // Backslashes need to be converted for Windows.
+  return srcDir?.replace(BACKSLASH_REGEXP, '/').includes(FUNCTIONS_INTERNAL_DIR)
```
danez, line 7: Can you please use `unixify()` instead of the custom regex.
Also we should use `is-path-inside` to check if the srcDir is inside `FUNCTIONS_INTERNAL_DIR`, as a path called `.netlify/internal-functions-not/` will happily match even though it is not our internal folder. `is-path-inside` takes care of that.

tests/main.test.ts
```diff
@@ -1718,6 +1737,32 @@ describe('zip-it-and-ship-it', () => {
...
+    vi.mocked(shellUtils.runCommand).mockImplementation(async (...args) => {
+      await writeFile(args[1][2], '')
+
+      return {} as any
+    })
+
+    const fixtureName = 'go-source-internal'
+    const { files } = await zipFixture(fixtureName, {
+      length: 2,
+      fixtureDir: join(FIXTURES_DIR, FUNCTIONS_INTERNAL_DIR),
```
danez, line 1750: can we please move the fixtures around and instead of having `tests/fixtures/.netlify/internal-functions/node_internal` do `tests/fixtures/node_internal/.netlify/internal-functions/`

danez (changes requested):

src/runtimes/node/bundlers/esbuild/bundler_target.ts
```diff
@@ -17,7 +17,7 @@ const versionMap = {
   '18.x': 'node18',
 } as const
 
-type VersionValues = (typeof versionMap)[keyof typeof versionMap]
+type VersionValues = typeof versionMap[keyof typeof versionMap]
```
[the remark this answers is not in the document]
khendrikse: yup! I was already on it :)

tests/main.test.ts
khendrikse: ✅ !

src/runtimes/constants.ts
khendrikse: Yeah it makes sense :) I'll start with Daniels suggestions, and then implement this!

src/utils/check_is_internal_function.ts
khendrikse: In the current setup, `is-path-inside` doesn't seem to work because `FUNCTIONS_INTERNAL_DIR` is not a full path. I changed the regex into unixify, and I'll figure out the path inside logic once I pick up Eduardo's comment, because I think then I'll have a more thorough logic that will have a full path for the internal directory :)

README.md
```diff
@@ -150,6 +150,10 @@ The following properties are accepted:
   The `[name]` placeholder will be replaced by the name of the function, allowing you to use it to construct the path to
   the target directory.
 
+- `displayName`
```
eduardoboucas, line 153: As you rightly pointed out, this property is called `name` in edge functions, and it would be great if we could standardise them somehow.

I know that we already have a `name` property in zip-it-and-ship-it, but it only exists in the output object, not the input. So we could potentially accept `name` as the input and then present it as something else in the output object (like `displayName`).

Alternatively, we could even repurpose the `name` property in the output object and start populating it with the value of the `name` _input_ property (and default to the current value if there's no `name` supplied). I think we only ever use that for presentation purposes anyway, which would come with the benefit of starting to show the display name in the CLI and Netlify Build when listing functions.

What do you think? (I totally understand if you don't want to make that part of this PR, just wanted to offer these suggestions for you to consider.)

README.md
```diff
@@ -202,6 +213,14 @@ properties.
 
   The size of the generated archive, in bytes.
 
+- `isInternalFunction` `boolean`
```
eduardoboucas, line 216: Could this be just `internal`?

tests/zip_function.test.ts
```diff
@@ -138,4 +138,34 @@ describe('zipFunction', () => {
...
+    ['bundler_default', 'bundler_esbuild', 'bundler_nft'],
+    async (options) => {
+      const { path: tmpDir } = await getTmpDir({ prefix: 'zip-it-test' })
+      const basePath = join(FIXTURES_DIR, 'node-internal', '.netlify/internal-functions')
+      const opts = merge(options, {
+        internalFunctionsFolder: basePath,
+      })
+      const result = (await zipFunction(`${basePath}/function-1.js`, tmpDir, opts))!
+
+      expect(result.isInternalFunction).toBeTruthy()
```
eduardoboucas, line 153: `toBeTruthy` will accept any value that isn't falsy (🥁). It's not guaranteeing that `result.isInternalFunction` is `true`, which I think we should be guaranteeing here.

README.md
khendrikse: I'd then maybe change it to `internalFunction` just because I like it to point explicitly to _what_ it is that is internal :) I think `internal` might just be a bit too implicit, and anything to help understand what it refers to might be nice.

tests/zip_function.test.ts
khendrikse: Makes sense! I'll change it :)

README.md
eduardoboucas: But all of the properties in the output object are about a specific function, so rather than `path` and `runtime` you'd have `functionPath`, `functionRuntime`, etc.

README.md
khendrikse: I like the idea of standardising the properties 🤔. Maybe we should start with the input `name` instead of `displayName`. But still have the output be `displayName` as you said. I'd like to really make sure we know what the usecase is for the output `name` property right now. I could investigate that in a later moment?

README.md
khendrikse: That's true 🤔 yeah I'll change it!

README.md
eduardoboucas: Sounds good!

README.md
khendrikse: How about `isInternal`. I know it looks like a small difference, but then it at least still refers to itself somehow? Giving it a bit more context?

eduardoboucas: Go for it.

README.md
```diff
@@ -172,11 +176,18 @@ JSON-formatted string with the following properties:
 
 #### `parallelLimit`
 
-- _Type_: `number`\
+- _Type_: `number`
 - _Default value_: `5`
 
 Maximum number of functions to bundle at the same time.
 
+#### `internalFunctionsFolder`
```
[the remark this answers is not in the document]
khendrikse: Yes I love this! Thank you, it didn't feel 100% right to me anyway :D

tests/main.test.ts
```diff
@@ -1899,6 +1950,44 @@ describe('zip-it-and-ship-it', () => {
     )
   })
 
+  test('Builds Rust functions from an internal-functions dir with a configured name', async () => {
+    vi.mocked(shellUtils.runCommand).mockImplementation(async (...args) => {
+      // eslint-disable-next-line unicorn/no-useless-undefined
+      const [rootCommand, , { env: environment = undefined } = {}] = args
```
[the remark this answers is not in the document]
khendrikse: ✅

src/runtimes/node/utils/module.ts
```diff
@@ -1,11 +1,10 @@
...
-
-// Windows path normalization
-const BACKSLASH_REGEXP = /\\/g
+import unixify from 'unixify'
 
 // When doing require("moduleName/file/path"), only keep `moduleName`
 export const getModuleName = function (dependency: string): string {
-  const dependencyA = dependency.replace(BACKSLASH_REGEXP, '/')
+  // Windows path normalization
+  const dependencyA = unixify(dependency)
```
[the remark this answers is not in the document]
khendrikse: makes sense!

eduardoboucas (approved):
🚀

PR #1243, opened 2022-10-31
chore: fix calculation of average in benchmark

danez:
#### Summary

See inline comments

---

      passes our tests.

### DIFF
--- a/benchmarks/helpers/main.js
+++ b/benchmarks/helpers/main.js
@@ -1,10 +1,20 @@
+/* eslint-disable max-statements */
+import assert from 'assert'
 import { performance, PerformanceObserver } from 'perf_hooks'
 import { fileURLToPath } from 'url'
 
 export const FIXTURES_DIR = fileURLToPath(new URL('../fixtures', import.meta.url))
 
+const sleep = (ms) =>
+  new Promise((resolve) => {
+    setTimeout(resolve, ms)
+  })
+
 export const timeFunction = async (func, runs = 1) => {
   const finishedRuns = new Map()
+
+  performance.clearMarks()
+  performance.clearMeasures()
 
   const observer = new PerformanceObserver((list) => {
     const [entry] = list.getEntries()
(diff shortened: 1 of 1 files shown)


benchmarks/helpers/main.js
```diff
@@ -1,11 +1,21 @@
...
+const sleep = (ms) =>
+  new Promise((resolve) => {
+    setTimeout(resolve, ms)
+  })
+
 export const timeFunction = async (func, runs = 1) => {
   const finishedRuns = new Map()
 
+  performance.clearMarks()
+  performance.clearMeasures()
```
danez, line 17: Not strictly necessary but if we run multiple benchmarks after each other, this is critical. Anyway it doesn't hurt.

benchmarks/helpers/main.js
```diff
@@ -14,17 +24,23 @@ export const timeFunction = async (func, runs = 1) => {
 
   observer.observe({ entryTypes: ['measure'] })
 
-  await Promise.all(
-    Array.from({ length: runs }).map(async (_, index) => {
-      performance.mark(`run-${index}-start`)
+  for (let index = 0; index < runs; index++) {
```
danez, line 27: Use for loop so that the runs are not running in parallel.

benchmarks/helpers/main.js
```diff
@@ -14,17 +24,23 @@ export const timeFunction = async (func, runs = 1) => {
...
 
-      await func(index)
+    performance.measure(`run-${index}`, `run-${index}-start`)
+  }
 
-      performance.measure(`run-${index}`, `run-${index}-start`)
-    }),
-  )
+  // wait for PerformanceObserver to gather all data
+  await sleep(100)
```
danez, line 36: As the PerformanceObserver is triggered async, we need to wait till it was triggered.
Right now out of the 3 runs we usually only calculate the average with the numbers of 2 runs but divide it by 3.

The assert a little down is also there to ensure we captured all runs.

khendrikse (approved):

PR #1244, opened 2022-11-01
feat: improve performance by reworking caching and supply cache to NFT

danez:
#### Summary

Based on #1243 

Ref: https://github.com/netlify/pod-compute/issues/252

There is also another PR that I created in NFT, as NFT does not correctly cache, which will bring another small performance boost. vercel/nft#320

---

      passes our tests.

### DIFF
--- a/src/runtimes/detect_runtime.ts
+++ b/src/runtimes/detect_runtime.ts
@@ -1,8 +1,6 @@
-import type { Buffer } from 'buffer'
+import { readFile } from 'fs/promises'
 
 import { detect, Runtime as BinaryRuntime, Arch, Platform, BinaryInfo } from '@netlify/binary-info'
-
-import { cachedReadFile, FsCache } from '../utils/fs.js'
 
 import { RuntimeType } from './runtime.js'
 
@@ -21,20 +19,10 @@ The binary needs to be built for Linux/Amd64, but it was built for ${Platform[bi
 }
 
 // Try to guess the runtime by inspecting the binary file.
-export const detectBinaryRuntime = async function ({
-  fsCache,
-  path,
-}: {
-  fsCache: FsCache
-  path: string
-}): Promise<RuntimeType | undefined> {
+export const detectBinaryRuntime = async function ({ path }: { path: string }): Promise<RuntimeType | undefined> {
   try {
-    const buffer = await cachedReadFile(fsCache, path)
-
-    // We're using the Type Assertion because the `cachedReadFile` abstraction
-    // loses part of the return type information. We can safely say it's a
-    // Buffer in this case because we're not specifying an encoding.
-    const binaryInfo = detect(buffer as Buffer)
+    const fileContents = await readFile(path)
+    const binaryInfo = detect(fileContents)
 
     if (!isValidFunctionBinary(binaryInfo)) {
       return warnIncompatibleBinary(path, binaryInfo)
--- a/src/runtimes/node/bundlers/esbuild/plugin_native_modules.ts
+++ b/src/runtimes/node/bundlers/esbuild/plugin_native_modules.ts
@@ -1,10 +1,9 @@
 import path from 'path'
 
 import type { Plugin } from '@netlify/esbuild'
-import readPackageJson from 'read-package-json-fast'
 
 import { isNativeModule } from '../../utils/detect_native_module.js'
-import { PackageJson } from '../../utils/package_json.js'
+import { PackageJson, readPackageJson } from '../../utils/package_json.js'
 import type { NativeNodeModules } from '../types.js'
 
 type NativeModuleCacheEntry = [boolean | undefined, PackageJson]
--- a/src/runtimes/node/bundlers/nft/es_modules.ts
+++ b/src/runtimes/node/bundlers/nft/es_modules.ts
@@ -4,15 +4,16 @@ import { NodeFileTraceReasons } from '@vercel/nft'
 
 import type { FunctionConfig } from '../../../../config.js'
 import { FeatureFlags } from '../../../../feature_flags.js'
-import { cachedReadFile, FsCache } from '../../../../utils/fs.js'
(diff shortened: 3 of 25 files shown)


src/runtimes/detect_runtime.ts
```diff
@@ -21,20 +19,10 @@ The binary needs to be built for Linux/Amd64, but it was built for ${Platform[bi
...
-}): Promise<RuntimeType | undefined> {
+export const detectBinaryRuntime = async function ({ path }: { path: string }): Promise<RuntimeType | undefined> {
   try {
-    const buffer = await cachedReadFile(fsCache, path)
-
-    // We're using the Type Assertion because the `cachedReadFile` abstraction
-    // loses part of the return type information. We can safely say it's a
-    // Buffer in this case because we're not specifying an encoding.
-    const binaryInfo = detect(buffer as Buffer)
+    const fileContents = await readFile(path)
```
danez, line 24: I removed the cache here, as this is reading a binary file and we only do that once for a binary we find in the function folder.

src/runtimes/node/bundlers/esbuild/plugin_native_modules.ts
```diff
@@ -1,10 +1,9 @@
 import path from 'path'
 
 import type { Plugin } from '@netlify/esbuild'
-import readPackageJson from 'read-package-json-fast'
 
 import { isNativeModule } from '../../utils/detect_native_module.js'
-import { PackageJson } from '../../utils/package_json.js'
+import { PackageJson, readPackageJson } from '../../utils/package_json.js'
```
danez, line 6: `read-package-json-fast` does `JSON.parse(readFile())` plus some additional validation and checks for binary folders. So our implementation should actually be faster without those checks. :)

src/utils/cache.ts
```diff
@@ -0,0 +1,35 @@
...
+  // file content
+  fileCache: FileCache
+  lstatCache: LstatCache
+  readDirCache: ReaddirCache
+  // NFT cache, which should not be used in zisi and only supplied to NFT
+  // this cache shares the file cache with zisi
+  nftCache: Partial<NFTCache>
+}
+
+export const createNewCache = (): RuntimeCache => {
```
eduardoboucas, line 24: I find it curious that you have a method called `createNew<something>` that returns an object with internal state. Any reason not to use a class and do `new Cache()`?

src/runtimes/node/bundlers/nft/es_modules.ts
```diff
@@ -148,21 +149,21 @@ const shouldTranspile = (
...
+  cache: RuntimeCache
   config: FunctionConfig
   esmPaths: Set<string>
-  fsCache: FsCache
   reasons: NodeFileTraceReasons
   name: string
 }) => {
-  const cache: Map<string, boolean> = new Map()
-  const pathsToTranspile = [...esmPaths].filter((path) => shouldTranspile(path, cache, esmPaths, reasons))
+  const shouldCompileCache: Map<string, boolean> = new Map()
```
eduardoboucas, line 165: I feel that this comment would help contextualise this variable in relation to the `cache` argument.

```suggestion
  // Used for memoizing the check for whether a path should be transpiled.
  const shouldCompileCache: Map<string, boolean> = new Map()
```

src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -75,30 +78,33 @@ const ignoreFunction = (path: string) => {
...
   pluginsModulesPath?: string
   name: string
 }) {
-  const fsCache: FsCache = {}
   const {
     fileList: dependencyPaths,
     esmFileList,
     reasons,
   } = await nodeFileTrace([mainFile], {
+    fileIOConcurrency: 2048,
```
eduardoboucas, line 101: Can we add a comment explaining the reasoning behind this magic number?

src/utils/cache.ts
```diff
@@ -0,0 +1,35 @@
...
+
+interface NFTCache {
+  fileCache: FileCache
+  statCache: Map<string, unknown>
+  symlinkCache: Map<string, unknown>
+  analysisCache: Map<string, unknown>
+}
+
+export interface RuntimeCache {
+  // file content
```
eduardoboucas, line 15: I found this comment a bit confusing. I think this would make it clearer?

```suggestion
  // IO operations
```

src/utils/cache.ts
```diff
@@ -0,0 +1,35 @@
...
+  fileCache: FileCache
+  lstatCache: LstatCache
+  readDirCache: ReaddirCache
+  // NFT cache, which should not be used in zisi and only supplied to NFT
+  // this cache shares the file cache with zisi
+  nftCache: Partial<NFTCache>
```
eduardoboucas, line 21: Why the use of `Partial`?

src/utils/cache.ts
danez: Huh, I didn't even think about this. Yes that makes sense. Done

src/utils/cache.ts
danez: I cleared that up

src/utils/cache.ts
danez: I changed the type to only include the props that are important for us.

eduardoboucas (approved):
LGTM! This change is significant enough to warrant a feature flag, but I'm not sure how easy it would be to add one, since the changes have pretty deep ramifications.

Have you thought about that?

danez:
> LGTM! This change is significant enough to warrant a feature flag, but I'm not sure how easy it would be to add one, since the changes have pretty deep ramifications.
> 
> Have you thought about that?

Yes I forgot about this. I put the nft cache and the higher file IO limit behind FF. The rest was more or less a refactor and should not change much.

eduardoboucas (approved):

eduardoboucas (approved):

PR #1258, opened 2022-11-04
fix: correctly detect conflicting function names consistently

danez:
#### Summary

I was investigating why the tests fail sometimes. The tests ensure that when multiple files for the same function exist we always use the same order in how we detect them.

My first thought was it has to do with async stuff in the tests, just to figure out during my perf change I did change `locate-path` to not preserve order, which triggers this flaky behavior.

This reverts this change, plus some additional fixes for the tests, I run into on the way.

---

      passes our tests.

### DIFF
--- a/tests/helpers/main.ts
+++ b/tests/helpers/main.ts
@@ -47,6 +47,8 @@ export const zipFixture = async function (
   const bundlerString = getBundlerNameFromConfig(config) || 'default'
   const { path: tmpDir } = await getTmpDir({
     prefix: `zip-it-test-bundler-${bundlerString}`,
+    // Cleanup the folder on process exit even if there are still files in them
+    unsafeCleanup: true,
   })
 
   if (env.ZISI_KEEP_TEMP_DIRS !== undefined) {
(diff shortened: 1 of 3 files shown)


tests/helpers/main.ts
```diff
@@ -47,6 +47,8 @@ export const zipFixture = async function (
   const bundlerString = getBundlerNameFromConfig(config) || 'default'
   const { path: tmpDir } = await getTmpDir({
     prefix: `zip-it-test-bundler-${bundlerString}`,
+    // Cleanup the folder on process exit even if there are still files in them
+    unsafeCleanup: true,
```
danez, line 51: We never clean the temp folders, because there are still files left in them. With this change we do force remove the tmp folders on process exit.

tests/helpers/test_many.ts
```diff
@@ -73,7 +73,7 @@ export const makeTestMany = <M extends string>(
   const testFns = ['fails', 'only', 'concurrent', 'skip', 'todo']
 
   testFns.forEach((fn) => {
-    testBundlers[fn] = ((...args) => testBundlers(...args, test[fn])) as TestMany<M>
+    testBundlers[fn] = ((...args) => testBundlers(...args, testAPI[fn])) as TestMany<M>
```
danez, line 76: This does not really mattter, but it is cleaner to use the testAPI that is supplied to the `makeTestMany()` which is `test` instead of using `test` directly.

jackiewmacharia (approved):

PR #1272, opened 2022-11-28
chore: unzip in tests in series and add missing types for tmp

danez:
#### Summary

Sometimes the tests in zisi fail on windows, this happens because the unzip in the tests is running in parallel for all functions. As they all extract to the same directory, permission errors happen when two functions contain the same files, which usually is [the package.json](https://github.com/netlify/zip-it-and-ship-it/blob/main/tests/fixtures/package.json) file.

Also added `@types/tmp`. We are not using tmp directly, but `tmp-promise` uses types from `@types/tmp`

---

      passes our tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,6 +53,7 @@
         "@types/normalize-path": "^3.0.0",
         "@types/resolve": "^1.20.2",
         "@types/semver": "^7.3.8",
+        "@types/tmp": "^0.2.3",
         "@types/unixify": "^1.0.0",
         "@types/yargs": "^17.0.4",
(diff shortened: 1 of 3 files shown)


eduardoboucas (approved):
LGTM. I'm curious about how this will impact test times?

danez:
I thought about this too, the alternative would be to extract every function in a separate directory. Totally doable but needs more changes.

Looking at the main branch most commits I looked at take between 58s to 1m20s for the tests with coverage and so does this PR (2 runs: 58s & 56s). Not very scientific, but I think good enough. So unless we were really lucky in this PR and got the fastest github action runner ever it should be neglectable I think.

PR #1291, opened 2022-12-14
fix: mark errors from babel-parser as user error

danez:
#### Summary

Fix https://github.com/netlify/pod-compute/issues/308


src/runtimes/node/bundlers/zisi/src_files.ts
```diff
@@ -111,9 +110,7 @@ const getFileDependencies = async function ({
   state.localFiles.add(path)
 
   const basedir = dirname(path)
-  const dependencies = featureFlags.parseWithEsbuild
-    ? await listImports({ functionName, path })
-    : await precinct.paperwork(path, { includeCore: false })
```
danez, line 116: This was moved to list_imports and there surrounded with a try catch.

tests/main.test.ts
```diff
@@ -1384,13 +1384,14 @@ describe('zip-it-and-ship-it', () => {
   })
 
   testMany(
-    'Adds `type: "functionsBundling"` to user errors when parsing with esbuild',
-    ['bundler_esbuild'],
+    'Adds `type: "functionsBundling"` to user errors when parsing with esbuild or zisi',
+    ['bundler_esbuild', 'bundler_esbuild_zisi', 'bundler_default'],
```
eduardoboucas, line 1388: What happens with NFT?

eduardoboucas (approved):

tests/main.test.ts
danez: Not sure exactly, but I know it ignores missing files and probably also files that have a syntax error while finding imports. So it does not throw at buildtime.

PR #1308, opened 2023-01-11
Test: add a unit test for addArchiveSize

khendrikse:

#### Summary

Just a small PR to add more unit tests. There is an integration test that covers this function, but I think it would be nice to also have something simple that could help point out where the problem is, if the integration test were to fail.

---

      passes our tests.


tests/unit/utils/archive_size.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import { describe, expect, test } from 'vitest'
+
+import { FunctionArchive } from '../../../src/function.js'
+import { addArchiveSize } from '../../../src/utils/archive_size.js'
+import { FIXTURES_DIR } from '../../helpers/main.js'
+
+describe('addArchiveSize', () => {
+  test('adds the archive size of the file at `path` if it is a ZIP archive', async () => {
+    const functionArchive = {
+      path: resolve(join(FIXTURES_DIR, 'archive-size', 'normal.zip')),
```
eduardoboucas, line 12: I don't think you need `join` here.

```suggestion
      path: resolve(FIXTURES_DIR, 'archive-size', 'normal.zip'),
```

tests/unit/utils/archive_size.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    } as FunctionArchive
+
+    const result = await addArchiveSize(functionArchive)
+
+    expect(result.size).toBe(1098)
+  })
+
+  test('does not add the archive size of the file at `path` if it is not a ZIP archive', async () => {
+    const functionArchive = {
+      path: resolve(join(FIXTURES_DIR, 'archive-size', 'normal.js')),
```
eduardoboucas, line 22: I don't think you need `join` here.

```suggestion
      path: resolve(FIXTURES_DIR, 'archive-size', 'normal.js'),
```

package.json
```diff
@@ -16,6 +16,7 @@
     "prepublishOnly": "npm ci && npm test",
     "prepack": "npm run build",
     "build": "tsc",
+    "build-watch": "tsc -w",
```
eduardoboucas, line 19: nit: I think this would be more aligned with the other scripts.

```suggestion
    "build:dev": "tsc -w",
```

eduardoboucas (dismissed):
Awesome! Thanks for doing this. ✨

eduardoboucas (approved):
✨ 

Small thing: we typically use the `chore:` prefix for this type of PR. Just a convention with no actual impact, since it's not meant to trigger a release.

PR #1310, opened 2023-01-12
fix: generate entry file on naming conflicts

danez:
#### Summary

When a naming conflict occurs we need to create a entry file even if there is already a similar named file, but it will be moved to a subfolder.

Fixes the nuxt3 problem

when a functions has this layout

```
netlify/functions/server/
                         server.js <-- detected as main file
                         server.mjs
```

previously we would leave everything as is, as `server.js` is the main file and has the same name as the entry file so nothing to do.

With #1190 we now consider `server.mjs` also as a potential entry file and detect that it can potentially conflict with the entry file, so we move everything from that function into a subfolder.

```
netlify/functions/server/src/
                             server.js
                             server.mjs
```

when we do this we now need to generate an entry file again, even though we would detect that `server.js` would satisfy as entry-point, but due to everything being moved into a subfolder we need to.

```
netlify/functions/server/src/
                             server.js
                             server.mjs
                         server.js
```

And the chain in this scenario is `./server.js -> ./src/server.js -> ./src/server.mjs`

This is all not optimal, and we should prioritize changing the AWS entry point to something unique like `__netlifyLambdaEntry` and disallow a function to be named like that. Then all of this code can be removed.


eduardoboucas (commented):
Do you mind adding a couple of examples to the PR description so that it's easier to understand what happens exactly in the cases we're targeting?

danez:
> Do you mind adding a couple of examples to the PR description so that it's easier to understand what happens exactly in the cases we're targeting?

Done

eduardoboucas (approved):

eduardoboucas:
> This is all not optimal, and we should prioritize changing the AWS entry point to something unique like `__netlifyLambdaEntry` and disallow a function to be named like that. Then all of this code can be removed.

Agreed. Do we have an issue for tracking that work? If not, do you mind creating one?

danez:
> Agreed. Do we have an issue for tracking that work? If not, do you mind creating one?

It is part of the notion doc about all the improvements for zip-it-and-ship-it.

PR #1328, opened 2023-01-30
feat: disallow unique entry file name

danez:
#### Summary

Ref https://github.com/netlify/pod-compute/issues/349

This disallows functions to be named `___netlify-lambda-entry` and also disallows any additional files/directories in the bundled function to contain `___netlify-lambda-entry`.

This is behind a FF. I also thought about silently ignoring function files named like this, but I think that would be more surprising.

The release plan is to roll this feature out and monitor how often this error happens and if it happens, disable FF and investigate and maybe pick another name. But I suspect it not being used at all. The expected case is that no error is thrown at all when activating the FF.

---

      passes our tests.


src/feature_flags.ts
```diff
@@ -23,6 +23,9 @@ export const defaultFlags: Record<string, boolean> = {
 
   // Output CJS file extension
   zisi_output_cjs_extension: false,
+
+  // Output CJS file extension
```
eduardoboucas, line 27: This needs to change?

src/runtimes/index.ts
```diff
@@ -4,9 +4,11 @@ import { Config, getConfigForFunction, FunctionWithoutConfig } from '../config.j
 import { defaultFlags, FeatureFlags } from '../feature_flags.js'
 import { FunctionSource } from '../function.js'
 import type { RuntimeCache } from '../utils/cache.js'
+import { FunctionBundlingUserError } from '../utils/error.js'
 
 import goRuntime from './go/index.js'
 import jsRuntime from './node/index.js'
+import { UNIQUE_ENTRY_FILE_NAME } from './node/utils/entry_file.js'
```
eduardoboucas, line 11: nit: Can this be called just `ENTRY_FILE_NAME`? Not sure what "unique" is adding here.

src/runtimes/index.ts
```diff
@@ -47,15 +49,27 @@ const findFunctionsInRuntime = async function ({
...
-      ...func,
-      extension: extname(func.mainFile),
-      filename: basename(func.srcPath),
-      runtime,
-    },
-  ])
+  const augmentedFunctions: FunctionTupleWithoutConfig[] = functions.map((func) => {
+    if (featureFlags.zisi_disallow_new_entry_name && func.name === UNIQUE_ENTRY_FILE_NAME) {
+      throw new FunctionBundlingUserError(
+        `The name '${UNIQUE_ENTRY_FILE_NAME}' is a reserved name and it is not allowed to be used as a function name.`,
```
eduardoboucas, line 55: nit: I think this reads a bit easier.

```suggestion
        `'${UNIQUE_ENTRY_FILE_NAME}' is a reserved name and cannot be used as a function name.`,
```

src/runtimes/node/utils/entry_file.ts
```diff
@@ -46,14 +50,32 @@ export const conflictsWithEntryFile = (
...
     basePath: string
+    extension: string
+    featureFlags: FeatureFlags
     filename: string
     mainFile: string
   },
-) => srcFiles.some((srcFile) => isNamedLikeEntryFile(srcFile, { basePath, filename }) && srcFile !== mainFile)
+) => {
+  srcFiles.forEach((srcFile) => {
+    if (featureFlags.zisi_disallow_new_entry_name && srcFile.includes(UNIQUE_ENTRY_FILE_NAME)) {
```
eduardoboucas, line 66: Can we do the feature flag check on the outside and avoid iterating over `srcFiles` again if the flag is not empty?

Also, `srcFiles` can get quite large. Would it make sense to merge this with the `.some` so that we iterate over the array only once?

src/feature_flags.ts
danez: ooppps :)

src/runtimes/node/utils/entry_file.ts
danez: Okay I merged them.

src/runtimes/node/utils/entry_file.ts
```diff
@@ -46,14 +50,38 @@ export const conflictsWithEntryFile = (
...
+    featureFlags: FeatureFlags
     filename: string
     mainFile: string
   },
-) => srcFiles.some((srcFile) => isNamedLikeEntryFile(srcFile, { basePath, filename }) && srcFile !== mainFile)
+) => {
+  let hasConflict = false
+
+  srcFiles.forEach((srcFile) => {
+    if (featureFlags.zisi_disallow_new_entry_name && srcFile.includes(ENTRY_FILE_NAME)) {
```
eduardoboucas, line 68: Why are you using `includes` here?

danez: To catch dirs as well as files that contain the name. It would also catch names that only include `___netlify-lambda-entry` as part of it, but it is simpler, the alternative would be:

```js
basename(srcFile, extname(srcFile)) === ENTRY_FILE_NAME || srcFile.includes(`/${ENTRY_FILE_NAME}/`)
```

eduardoboucas: I see. I find it a bit disconcerting that we're checking so loosely for that string anywhere in the file or directory names, even if that's a very unique string.

But more generally, I worry if we need to do this type of check at all. We're the ones adding the entry file, so we know exactly what is its name and extension. Can we check, at the point of inserting it, whether there's already a file in the zip with that name? At that point, it becomes a simple equality check, which should also be more performant?

danez: The problem is that it is not just one file, but multiple that we are forbidding. When AWS searches for the entrypoint this is what it looks for (tested just now):

1. /___netlify-lambda-entry.js
2. /___netlify-lambda-entry.mjs
3. /___netlify-lambda-entry.cjs
4. /___netlify-lambda-entry/index.js
5. /___netlify-lambda-entry/index.mjs
6. /___netlify-lambda-entry/index.cjs

So any user file named like this should not be allowed. Checking when adding the entry file is a little too late for now, because this PR is meant to check if `___netlify-lambda-entry` is already used. And as we currently do not always generate an entry file, we would not always check for these files.

Alternatively we could have a Set with all those relative paths and check if the user files is included?

danez: Once we generate the file (lets say `/___netlify-lambda-entry.mjs`) this would change, as the only thing we need to forbid is `/___netlify-lambda-entry.js` (because it has higher prior than its mjs equvivalent) and `/___netlify-lambda-entry.mjs`

eduardoboucas: Ah, I see!

> Alternatively we could have a Set with all those relative paths and check if the user files is included?

That makes sense to me. If you're okay with that, I would suggest this route.

danez: Worse, some combinations make Lambda crash I just noticed:
Crashes:
Screenshot 2023-01-31 at 15 27 08

Rename the folder:
Screenshot 2023-01-31 at 15 27 28


I will look into the Set.

danez: Disallowing `/___netlify-lambda-entry/` and `___netlify-lambda-entry.(c|m)?js` would look like this:
I am not sure I like this, and also does a lot more per iteration.

```diff
@@ -1,4 +1,4 @@
-import { basename, extname, resolve } from 'path'
+import { basename, extname, relative, resolve } from 'path'
 
 import type { FeatureFlags } from '../../../feature_flags.js'
 import { FunctionBundlingUserError } from '../../../utils/error.js'
@@ -24,6 +24,12 @@
   return `export { handler } from '${importPath}'`
 }
 
+const LAMBDA_ENTRY_FILES = new Set([
+  ENTRY_FILE_NAME + ModuleFileExtension.JS,
+  ENTRY_FILE_NAME + ModuleFileExtension.MJS,
+  ENTRY_FILE_NAME + ModuleFileExtension.CJS,
+])
+
 // They are in the order that AWS Lambda will try to find the entry point
 const POSSIBLE_LAMBDA_ENTRY_EXTENSIONS = [ModuleFileExtension.JS, ModuleFileExtension.MJS, ModuleFileExtension.CJS]
 
@@ -65,7 +71,12 @@
   let hasConflict = false
 
   srcFiles.forEach((srcFile) => {
-    if (featureFlags.zisi_disallow_new_entry_name && srcFile.includes(ENTRY_FILE_NAME)) {
+    const relativeSrcFile = relative(basePath, srcFile)
+
+    if (
+      featureFlags.zisi_disallow_new_entry_name &&
+      (relativeSrcFile.startsWith(`${ENTRY_FILE_NAME}/`) || LAMBDA_ENTRY_FILES.has(relativeSrcFile))
+    ) {
       throw new FunctionBundlingUserError(
         `'${ENTRY_FILE_NAME}' is a reserved word and cannot be used as a file or directory name.`,
         {

```

eduardoboucas: Agreed.

eduardoboucas (approved):
🚀

PR #1341, opened 2023-02-08
feat: add generator as an option to config and save it in the manifest and remove isInternal

khendrikse:

#### Summary

In this PR we do the following:

* Remove the [recently introduced](https://github.com/netlify/zip-it-and-ship-it/pull/1221) `isInternal` and `internalSrcFolder` configuration options.
* Add a `generator` configuration option.

For more information about the **why**, feel free to check this issue: https://github.com/netlify/pod-compute/issues/350

---

      passes our tests.

image of cute rabbit


README.md
```diff
@@ -152,8 +152,18 @@ The following properties are accepted:
...
+  - _Default value_: undefined
+
   A name to use when displaying the function in the Netlify UI. Populates the `displayName` property in the functions manifest for the specified function.
 
+- `generator`
+
+  - _Type_: `string`
+  - _Default value_: undefined
+
+  A field to use if the function has been autogenerated by a plugin or integration. A recommended format is `@netlify/fake-plugin@1.0.0`, where adding the version is highly appreciated.
```
danez, line 165: I wonder about this scenario:

A user has a website with an integration at version 1.0.0. The user updates the integration to 1.0.1, and the following build will build the functions, but the functions are identical to the one previously, so we do not upload them. The version in this case will still be 1.0.0, even though 1.0.1 is installed and deploys have happened with this version.

We should check if this is the case. Not a big issue, just something to keep in mind that the version on the function might not be equal the version that is installed.

danez (approved):
🚀

README.md
khendrikse: Good point, I'll start looking into this!

PR #1367, opened 2023-03-06
feat: add new runtimeVersion to manifest

danez:
#### Summary

The option `nodeVersion` now works differently and allows different styles of how the Node.js version is specified.
If it cannot be coerced to a Node.js version, the `undefined` is added to the manifest, leaving the choice of runtime to origin. We could also change this to have zisi add the default to the manifest, but I didn't do it because I feel its nicer to have the default version only in one place.

How it is supposed to work is that `@netlify/build` sets `nodeVersion` to `AWS_LAMBDA_JS_VERSION || buildUserNodeVersion` and based on this input we set a new field `runtimeVersion` in the manifest, or not if it is not a valid runtime version.
open-api reads the manifest and sends `runtimeVersion || runtime`.  In the origin we then detect if it is a runtime shortcut 'js' or a full runtime image name.

Ref: netlify/pod-compute#33
Ref: netlify/pod-compute#64

This would also allow the runtime being set by the deployment config API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ The following properties are accepted:
   - _Type_: `string`\
   - _Default value_: `16`
 
-  The version of Node.js to use as the compilation target and reported execution runtime. Any valid and supported
-  Node.js version is accepted. Examples:
+  The version of Node.js to use as the compilation target for bundlers. This is also used to determine the runtime
+  reported in the manifest. Any valid and supported Node.js version is accepted. Examples:
 
   - `14.x`
   - `16.1.0`
--- a/README.md
+++ b/README.md
@@ -128,16 +128,15 @@ The following properties are accepted:
 - `nodeVersion`
 
   - _Type_: `string`\
-  - _Default value_: `16.x`
+  - _Default value_: `16`
 
-  The version of Node.js to use as the compilation target. Possible values:
+  The version of Node.js to use as the compilation target and reported execution runtime. Any valid and supported
+  Node.js version is accepted. Examples:
 
-  - `8.x` (or `nodejs8.x`)
-  - `10.x` (or `nodejs10.x`)
-  - `12.x` (or `nodejs12.x`)
-  - `14.x` (or `nodejs14.x`)
-  - `16.x` (or `nodejs16.x`)
-  - `18.x` (or `nodejs18.x`)
+  - `14.x`
+  - `16.1.0`
+  - `v18`
+  - `nodejs18.x` (for backwards compatibility)
 
 - `rustTargetDirectory`
 
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -9,7 +9,7 @@ interface ManifestFunction {
   name: string
   path: string
   runtime: string
-  runtimeName?: string
+  runtimeVersion?: string
   schedule?: string
   displayName?: string
   bundler?: string
@@ -48,7 +48,7 @@ const formatFunctionForManifest = ({
   name,
   path,
   runtime,
-  runtimeName,
+  runtimeVersion,
   schedule,
 }: FunctionResult): ManifestFunction => ({
   bundler,
@@ -56,7 +56,7 @@ const formatFunctionForManifest = ({
   generator,
   mainFile,
   name,
-  runtimeName,
+  runtimeVersion,
   path: resolve(path),
   runtime,
   schedule,
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -9,6 +9,7 @@ interface ManifestFunction {
   name: string
   path: string
   runtime: string
+  runtimeName?: string
   schedule?: string
   displayName?: string
   bundler?: string
@@ -40,21 +41,23 @@ export const createManifest = async ({ functions, path }: { functions: FunctionR
 }
 
 const formatFunctionForManifest = ({
+  bundler,
+  displayName,
+  generator,
   mainFile,
   name,
   path,
   runtime,
+  runtimeName,
   schedule,
-  displayName,
-  bundler,
-  generator,
 }: FunctionResult): ManifestFunction => ({
+  bundler,
+  displayName,
+  generator,
   mainFile,
   name,
+  runtimeName,
   path: resolve(path),
   runtime,
   schedule,
-  displayName,
-  bundler,
-  generator,
 })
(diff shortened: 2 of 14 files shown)


src/manifest.ts
```diff
@@ -40,21 +41,23 @@ export const createManifest = async ({ functions, path }: { functions: FunctionR
 }
 
 const formatFunctionForManifest = ({
```
khendrikse, line 43: Nit: I'm not sure what this function is really doing still, the only thing I see is that it resolves the path :/

khendrikse (dismissed):

README.md
```diff
@@ -155,14 +154,16 @@ The following properties are accepted:
   - _Type_: `string`
   - _Default value_: undefined
 
-  A name to use when displaying the function in the Netlify UI. Populates the `displayName` property in the functions manifest for the specified function.
+  A name to use when displaying the function in the Netlify UI. Populates the `displayName` property in the functions
```
eduardoboucas, line 157: What's up with these formatting changes? Makes the diff a bit hard to parse.

src/runtimes/node/bundlers/esbuild/bundler_target.ts
```diff
@@ -35,20 +27,22 @@ const getModuleFormat = async (
   extension: string,
   configVersion?: string,
 ): Promise<{ includedFiles: string[]; moduleFormat: ModuleFormat }> => {
-  if (extension === ModuleFileExtension.MJS && featureFlags.zisi_pure_esm_mjs) {
+  if (featureFlags.zisi_pure_esm_mjs && extension === ModuleFileExtension.MJS) {
```
eduardoboucas, line 30: nit: Should this be moved to the `if` block below?

eduardoboucas (commented):
Can you add a note outlining the breaking changes?

src/manifest.ts
```diff
@@ -40,21 +41,23 @@ export const createManifest = async ({ functions, path }: { functions: FunctionR
...
+  bundler,
+  displayName,
+  generator,
   mainFile,
   name,
   path,
   runtime,
+  runtimeID,
```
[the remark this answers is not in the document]
danez: Yes this was also annoying me.
`runtime` is `js`, `rs`, `go`
Where `runtimeID` now is the actual runtime`nodejs18.x`

Imho, `runtime` should probably called `language`, but I didn't want to change an existing field. Other solution would be instead of `runtimeID` add a field called `nodeVersion` where we simply add the node version that zisi received and use it in bitballoon/FO to decide on the runtime. Although that would not allow the deployment config API to set a runtime, but maybe if they set a node-version would be okay?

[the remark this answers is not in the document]
biruwon: I am waiting on this for reviewing the rest of the PRs, as this field it's going to be change it makes the other PRs obsolete or in needs of a change. My two cents: I need to always check in code what these 2 are.

`language` and `runtimeID` seems good for me, as it kind of match the description for [AWS lambda runtimes](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html).

By [this comment](https://github.com/netlify/pod-compute/issues/64#issuecomment-1490060429), looks like we are going to make a PR in every repository touching runtimes, so maybe the effort of renaming them is not that big?

src/runtimes/node/bundlers/esbuild/bundler_target.ts
danez: These are two different FF.

eduardoboucas: Ah, my bad!

danez:
> Can you add a note outlining the breaking changes?

No breaking changes anymore. the `nodeVersion` field now supports a wider range of values, but still supports what was supported before.

vitest.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
   test: {
     setupFiles: ['./tests/helpers/vitest_setup.ts'],
     include: ['tests/**/*.test.ts'],
-    testTimeout: 30_000,
+    testTimeout: 60_000,
```
danez, line 8: Some windows test unrelated to this commit were timing out.

src/manifest.ts
danez: I guess yeah resolving the path and limiting what we output. We might not return everything from `FunctionResult`, but haven't checked.

src/manifest.ts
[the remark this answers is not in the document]
danez: I went for using one field for either language or the runtime name.

README.md
```diff
@@ -128,16 +128,15 @@ The following properties are accepted:
 - `nodeVersion`
 
   - _Type_: `string`\
-  - _Default value_: `16.x`
+  - _Default value_: `16`
 
-  The version of Node.js to use as the compilation target. Possible values:
+  The version of Node.js to use as the compilation target and reported execution runtime. Any valid and supported
```
eduardoboucas, line 133: Curious what "reported" means in this context?

src/runtimes/runtime.ts
```diff
@@ -39,14 +39,15 @@ export interface ZipFunctionResult {
...
+  generator?: string
   inputs?: string[]
   includedFiles?: string[]
   inSourceConfig?: ISCValues
   nativeNodeModules?: object
   nodeModulesWithDynamicImports?: string[]
   path: string
-  displayName?: string
-  generator?: string
+  runtimeName?: string
```
eduardoboucas, line 50: My understanding is that this now contains the version (e.g. `nodejs18.x`), so it's technically not just a runtime _name_ but also version?

eduardoboucas (approved):

khendrikse (commented):
I'm not seeing specifically weird stuff, and would approve, but I see Eduardo is also looking at this, and I think his review could be more valuable right now ❤️

README.md
danez: changed

src/runtimes/runtime.ts
danez: Any idea what I should name it?

src/manifest.ts
```diff
@@ -9,6 +9,7 @@ interface ManifestFunction {
   name: string
   path: string
   runtime: string
+  runtimeName?: string
```
eduardoboucas, line 12: As discussed on Slack, we should probably rename this to `runtimeVersion`?

eduardoboucas (changes requested):

PR #1382, opened 2023-04-12
chore: specify nodeBundler property type so the string "esbuild" can …

bijela-gora:
…be assigned to it

🎉 Thanks for submitting a pull request! 🎉

#### Summary

Fixes #1381

---

For us to review and ship your PR efficiently, please perform the following steps:

- [x] Open a [bug/issue](https://github.com/netlify/zip-it-and-ship-it/issues/new/choose) before writing your code 🧑‍💻.
      This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing
      a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**
image
Source of the image: https://external-content.duckduckgo.com/iu/?u=http%3A%2F%2Fimages5.fanpop.com%2Fimage%2Fphotos%2F30300000%2FCute-Animals-3-cute-animals-30317875-500-332.jpg&f=1&nofb=1&ipt=2d30c03a9fdbaa511839a6be5c293cf16140a7b43865607f3427e963b2b5ef6d&ipo=images

PR #1392, opened 2023-04-19
feat!: refactor enums and export types

### LINKED ISSUE #1381: NodeBundlerType as enum
Hello,

- I searched for duplicates in issues
  - `is:issue is:open NodeBundlerType`
  - `is:issue is:open enum`

What I found is not a real bug but a point where DX might be improved for some developers. Let me describe it.

I'm using `zip-it-and-ship-it` JS API in a script. To make the script type-safe I added TypeScript to a project. After this I found TSC reporting an error:
Screenshot from 2023-04-12 23-28-08

Then I tried to use enum member instead of string:
image

Only after I changed `isolatedModules` to `true` the error gone.

The comment `// @ts-ignore` helpful also:
image

It would be a little better if [the example from README.md](https://github.com/netlify/zip-it-and-ship-it#esbuild) worked as it is in the context of TypeScript also.

**- Please mention your node.js, and operating system version.**
Node.js v18.14.0
Fedora Linux 37 (Workstation Edition)

danez:
`ZipFunctionOptions`, `ZipFunctionsOption`, `Config`, and `FunctionConfig` are now exported because they are needed in build.

Inspired by this Video https://www.youtube.com/watch?v=jjMbPt_H3RQ  and by the reported issue in #1382 this is an attempt to fix the issue by removing enums altogether and instead use const objects.

This allows strings to be supplied in the config while also keeping the grouped constants for the different usecases.

BREAKING CHANGE: The exported typescript types for `NodeBundlerType`, `RuntimeType`, and
`ModuleFormat` are not enums anymore but unions of strings. Use the newly exported
`NODE_BUNDLER`, `RUNTIME`, and `MODULE_FORMAT` for constants. Also the types
`NodeBundlerType` and `RuntimeType` were renamed to `NodeBundlerName` and `RuntimeName`
respectively.

Fixes #1381

---

      passes our tests.


src/archive.ts
```diff
@@ -6,10 +6,16 @@ import { promisify } from 'util'
 import archiver, { Archiver } from 'archiver'
 import endOfStream from 'end-of-stream'
 
+import { ObjectValues } from './types/utils.js'
+
 export { Archiver as ZipArchive } from 'archiver'
 
-// TODO make enum
-export type ArchiveFormat = 'none' | 'zip'
+export const ARCHIVE_FORMAT = {
```
lukasholzer, line 13: you could use an `const enum ARCHIVE_FORMAT` as well

danez: But then we cannot export the type for external use. `const enum` does not exist at runtime.

PR #1394, opened 2023-04-20
fix: add feature flag for the inheritance of the Node.js version from build

danez:
#### Summary

I just want to make sure, we can turn this feature off completely. a featureFlag in build would actually not be able to 100% disable it.

Ref: https://github.com/netlify/pod-compute/issues/64

---

      passes our tests.


src/feature_flags.ts
```diff
@@ -23,10 +26,9 @@ export const defaultFlags: Record<string, boolean> = {
 
   // Do not allow ___netlify-entry-point as function or file name
   zisi_disallow_new_entry_name: false,
-}
+} as const
 
-export type FeatureFlag = keyof typeof defaultFlags
-export type FeatureFlags = Record<FeatureFlag, boolean>
+export type FeatureFlags = Partial<Record<keyof typeof defaultFlags, boolean>>
```
danez, line 31: The type changes make it so that we have autocompletion and validation for all feature flags

eduardoboucas (commented):
When you say "build", you're talking about the deployment configuration API, right? If so, I think it would be important to reflect that in the naming of things, otherwise it gets a bit confusing.

khendrikse (dismissed):

danez:
> When you say "build", you're talking about the deployment configuration API, right? If so, I think it would be important to reflect that in the naming of things, otherwise it gets a bit confusing.

No i mean `@nelify/build`. It sets `nodeVersion` through the node js api.

khendrikse (approved):

PR #1395, opened 2023-04-21
feat: add private support for v2 API

eduardoboucas:
#### Summary

Adds support for the v2 API of serverless functions, behind a feature flag.

It works by injecting a `___netlify-bootstrap.js` file at the root of the Lambda with the contents of the `@netlify/serverless-functions-api` module, and using its `getLambdaHandler` wrapper function as the handler function.

Part of https://github.com/netlify/pod-compute/issues/475.


eduardoboucas:
> This is mostly testing it everything still works right?

What do you mean by that? 🤔

PR #1407, opened 2023-05-04
feat: add support for new entry file

danez:

#### Summary

Ref netlify/pod-compute#349

---

      passes our tests.


eduardoboucas:
> I adapted the current test, but haven't added any new tests. The reason is that currently in the tests we extract all bundled functions from a fixture into one directory. This wasn't a good idea up until now if two function zips contain the same file, but now with every function containing the same entry file, we pretty much always overwrite files when extracting.

@danez Maybe I'm missing something, but doesn't [the second parameter of `unzipFiles`](https://github.com/netlify/zip-it-and-ship-it/blob/main/tests/helpers/main.ts#L85) solve this by letting you configure where you extract things to?

danez:
> @danez Maybe I'm missing something, but doesn't [the second parameter of `unzipFiles`](https://github.com/netlify/zip-it-and-ship-it/blob/main/tests/helpers/main.ts#L85) solve this by letting you configure where you extract things to?

True, right now there isn't enough information that this method gets, but that could be easily added (like the name of the function). But also knowing that we had problems in the past on windows when running unzipping in parallel, because of multiple files being unzipped to the same location.

The main reason I would want to fundamentally change this by default is because I think it has a lot of potential to make false positives. So a test might work, because one of the functions has a package.json, that after extracting all files has also an effect on the other extracted functions. Making sure they always stay separated would eliminate this forever.

Issue #1408, opened 2023-05-04
Fix tests to unzip multiple functions into separate folders


danez:
Currently we unzip all functions of a fixture into one folder. If the functions include the same files, we overwrite one or the other.
This gets worse now that we want to have a uniquely named entry file. Now every function contains the same file and for example we cannot test if all the functions contain the unique entry file.

The solution would be to unzip the functions to individual folders and adapt all tests to work with the new unzip location.

PR #1420, opened 2023-05-10
chore: unzip every function into their own folder in tests

danez:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Fixes: #1408 

This unzips all functions into separate folders in the tests and all tests have been modified to accommodate this change.

Some other improvements have been made see inline comments.

---

For us to review and ship your PR efficiently, please perform the following steps:

- [ ] Open a [bug/issue](https://github.com/netlify/zip-it-and-ship-it/issues/new/choose) before writing your code 🧑‍💻.
      This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing
      a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [ ] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #1423, opened 2023-05-11
feat: remove usage of del and end-of-stream package

danez:
#### Summary

Since #1408 there were some randomly tests failing. This happened because I created `v2api.test.ts` which had the same cleanup mechanism than `main.test.ts`, by removing all temp directories that start with `zip-it-test-bundler-`. So these two tests interfered with each other and removed temp directories that were still in use.

While figuring out what the issue was is stumbled upon some other improvements, see inline comments.


src/archive.ts
```diff
@@ -53,6 +49,12 @@ export const addZipContent = function (archive: Archiver, content: Buffer | stri
...
 // End zipping files
 export const endZip = async function (archive: Archiver, output: Writable): Promise<void> {
-  archive.finalize()
-  await pEndOfStream(output)
+  const result = new Promise<void>((resolve, reject) => {
+    output.on('error', (error) => reject(error))
+    output.on('finish', () => resolve())
+  })
+
+  await archive.finalize()
```
danez, line 57: We did not await this before, probably not a biggy as we anyway wait till the file is written, but better wait for both.

src/runtimes/node/utils/zip.ts
```diff
@@ -75,7 +74,7 @@ const createDirectory = async function ({
   const functionFolder = join(destFolder, basename(filename, extension))
 
   // Deleting the functions directory in case it exists before creating it.
-  await deleteFiles(functionFolder, { force: true })
+  await fs.rm(functionFolder, { recursive: true, force: true })
```
danez, line 77: no need for the node package.

tests/main.test.ts
```diff
@@ -63,12 +61,6 @@ declare module 'vitest' {
 }
 
 describe('zip-it-and-ship-it', () => {
-  afterAll(async () => {
-    if (env.ZISI_KEEP_TEMP_DIRS === undefined) {
-      await deleteAsync(`${tmpdir()}/zip-it-test-bundler-*`, { force: true })
-    }
-  })
```
danez, line 70: This was the main culprit for the failing tests.

tests/main.test.ts
```diff
@@ -552,19 +544,40 @@ describe('zip-it-and-ship-it', () => {
...
+
+    afterAll(async () => {
+      await rm(fixtureTmpDir, { recursive: true, force: true })
     })
 
-    const basePath = `${fixtureTmpDir}/many-dependencies`
-    await cpy('many-dependencies/**', basePath, { cwd: FIXTURES_DIR })
-    await execa('npm', ['install', '--no-package-lock', '--no-audit'], {
-      cwd: basePath,
+    testMany('Works with many dependencies', [...allBundleConfigs], async (options) => {
```
danez, line 565: This test was for each bundler copying the fixture to a temp directory and running npm install. This was completely unnecessary as `zipNode()` does not actually modify the fixture in any way.

So we do this once before the test and run all tests against this prepared fixture.

Afterwards we also cleanup the folder, which wasn't done before.

tests/helpers/main.ts
```diff
@@ -49,19 +70,21 @@ export const zipNode = async function (
...
-    // Cleanup the folder on process exit even if there are still files in them
-    unsafeCleanup: true,
+    keep: ZISI_KEEP_TEMP_DIRS,
   })
 
-  if (env.ZISI_KEEP_TEMP_DIRS !== undefined) {
+  if (ZISI_KEEP_TEMP_DIRS) {
     console.log(tmpDir)
+  } else {
+    cleanupDirectories.push(tmpDir)
```
danez, line 87: Instead of removing all temporary directories that start with `zip-it-test-bundler-`, we now collect all directories and remove them after all tests.

tests/helpers/main.ts
```diff
@@ -49,19 +65,24 @@ export const zipNode = async function (
...
-    // Cleanup the folder on process exit even if there are still files in them
+    // Cleanup the folder even if there are still files in them
     unsafeCleanup: true,
+    keep: ZISI_KEEP_TEMP_DIRS,
   })
 
-  if (env.ZISI_KEEP_TEMP_DIRS !== undefined) {
+  if (ZISI_KEEP_TEMP_DIRS) {
     console.log(tmpDir)
+  } else if (!isCI) {
```
danez, line 83: Do not do the cleanup on CI, windows is very sensitive to fs access on github actions, and this increases the runtime 3x.

src/runtimes/node/utils/zip.ts
```diff
@@ -75,11 +75,11 @@ const createDirectory = async function ({
   const functionFolder = join(destFolder, basename(filename, extension))
 
   // Deleting the functions directory in case it exists before creating it.
-  await deleteFiles(functionFolder, { force: true })
-  await fs.mkdir(functionFolder, { recursive: true })
+  await rm(functionFolder, { recursive: true, force: true, maxRetries: 3 })
```
danez, line 78: use rm instead of `del` package. retries on common windows errors like EPERM, EBUSY, etc. with linear backoff with 100ms more each time.

tests/helpers/sha.ts
```diff
@@ -9,7 +9,11 @@ export const computeSha1 = async function (filePath: string): Promise<string> {
   const fileStream = createReadStream(filePath)
   const hashStream = createHash('sha1')
   hashStream.setEncoding('hex')
+
   const sha1Checksum = await getStream(fileStream.pipe(hashStream))
 
+  fileStream.destroy()
+  hashStream.destroy()
```
danez, line 16: ensure the streams are destroyed so that windows fs handles are released.

tests/helpers/main.ts
```diff
@@ -97,7 +118,6 @@ export const unzipFiles = async function (files: FunctionResult[]): Promise<Test
     Object.keys(files).map(async (key) => {
       const { path, name } = files[key]
       const dest = join(dirname(path), name)
-      await expect(path).toPathExist()
```
danez, line 100: this was added in #1420, but windows gets really slow because of this.

tests/helpers/main.ts
```diff
@@ -16,6 +17,8 @@ import { ZipFunctionsOptions } from '../../src/zip.js'
 export const FIXTURES_DIR = fileURLToPath(new URL('../fixtures', import.meta.url))
 export const BINARY_PATH = fileURLToPath(new URL('../../dist/bin.js', import.meta.url))
 
+const ZISI_KEEP_TEMP_DIRS = env.ZISI_KEEP_TEMP_DIRS !== undefined
```
eduardoboucas, line 20: nit: This is not a constant so I would name it something like `keepTempDirs`.

tests/helpers/main.ts
```diff
@@ -33,6 +36,17 @@ interface ZipNodeReturn extends ZipReturn {
   files: TestFunctionResult[]
 }
 
+let cleanupDirectories: string[] = []
```
eduardoboucas, line 39: I know you mention this later, but I think having the comment at the top would be useful.

```suggestion
// Accumulator of temporary directories that need to be cleaned up afterwards.
let cleanupDirectories: string[] = []
```

eduardoboucas, line 39: nit: You mention this in an inline comment, but having this in an actual comment in the code would be useful.

```suggestion
// Accumulator of temporary directories that need to be cleaned up afterwards.
let cleanupDirectories: string[] = []
```

eduardoboucas (dismissed):
Nice work! We should make this a `feat:` or `fix:` though, since you're modifying actual runtime functionality, so we should trigger a release and monitor the rollout.

eduardoboucas (approved):

PR #1429, opened 2023-05-12
feat: detect V2 API using static analysis

eduardoboucas:
#### Summary

Part of https://github.com/netlify/pod-compute/issues/476.

### DIFF
--- a/src/runtimes/node/in_source_config/index.ts
+++ b/src/runtimes/node/in_source_config/index.ts
@@ -5,15 +5,16 @@ import { FunctionBundlingUserError } from '../../../utils/error.js'
 import { nonNullable } from '../../../utils/non_nullable.js'
 import { RUNTIME } from '../../runtime.js'
 import { createBindingsMethod } from '../parser/bindings.js'
-import { getMainExport } from '../parser/exports.js'
+import { getExports } from '../parser/exports.js'
 import { getImports } from '../parser/imports.js'
-import { safelyParseFile } from '../parser/index.js'
+import { parseSource, safelyReadSource } from '../parser/index.js'
 
 import { parse as parseSchedule } from './properties/schedule.js'
 
 export const IN_SOURCE_CONFIG_MODULE = '@netlify/functions'
 
 export type ISCValues = {
+  apiVersion?: number
   invocationMode?: InvocationMode
   schedule?: string
 }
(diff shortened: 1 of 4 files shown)


src/runtimes/node/in_source_config/index.ts
```diff
@@ -37,22 +39,43 @@ const validateScheduleFunction = (functionFound: boolean, scheduleFound: boolean
...
+
   let scheduledFunctionFound = false
   let scheduleFound = false
 
   const getAllBindings = createBindingsMethod(ast.body)
-  const mainExports = getMainExport(ast.body, getAllBindings)
-  const iscExports = mainExports
+  const { defaultExport, handlerExports } = getExports(ast.body, getAllBindings)
+
+  if (featureFlags.zisi_functions_api_v2 && handlerExports.length === 0 && defaultExport !== undefined) {
```
eduardoboucas, line 72: Now the feature flag is used to control whether the detection mechanism kicks in. We should be able to roll it out safely to everyone and remove the flag entirely as a quick follow-up.

src/runtimes/node/parser/index.ts
```diff
@@ -76,6 +76,7 @@ export const parseExpression = ({
   resolveDir: string
 }) => {
   const { program } = parse(rawExpression, {
+    plugins: ['typescript'],
```
eduardoboucas, line 79: Because we're now parsing the source before bundling/transpiling.

src/runtimes/node/index.ts
```diff
@@ -67,6 +59,17 @@ const zipFunction: ZipFunction = async function ({
     return { config, path: destPath }
   }
 
+  const inSourceConfig = await findISCDeclarationsInPath(mainFile, name, featureFlags)
```
eduardoboucas, line 62: I'd like to call out this change, since we're now parsing ISC before bundling. This means we'll start parsing ESM, CJS and TS rather than the bundled/transpiled CJS. I don't anticipate any problems with this, but it's not something I can easily wrap with a feature flag, so something to keep an eye on.

package.json
```diff
@@ -55,6 +55,7 @@
   },
   "dependencies": {
     "@babel/parser": "7.16.8",
+    "@babel/plugin-syntax-typescript": "^7.21.4",
```
danez, line 58: There is no need to install this. This plugin is for babel itself, but we are only using the parser, which has no concept of plugin packages. For the parser plugins are simply strings that enable features.

src/runtimes/node/parser/exports.ts
```diff
@@ -1,35 +1,58 @@
...
 
-// Finds the main handler export in an AST.
-export const getMainExport = (nodes: Statement[], getAllBindings: BindingMethod) => {
-  let handlerExport: ISCExport[] = []
+// Finds and returns the following types of exports in an AST:
+// 1. Named `handler` function exports
+// 2. Default function export
+// 3. Named `config` object export
+export const getExports = (nodes: Statement[], getAllBindings: BindingMethod) => {
+  const handlerExports: ISCExport[] = []
```
Skn0tt, line 20: could you give an example for a file that has multiple exports named `handler`?

danez (changes requested):

PR #1138, opened 2022-07-08
feat: update default nodejs version to 16

biruwon:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

As commented by @Skn0tt, we are updating the default nodejs runtime to version 16 but was missing here. The version will be updated on Monday, so we can release it after that. Let me know if I am doing something wrong as it is my first PR here 😄 

---

For us to review and ship your PR efficiently, please perform the following steps:

- [ ] Open a [bug/issue](https://github.com/netlify/zip-it-and-ship-it/issues/new/choose) before writing your code 🧑‍💻.
      This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing
      a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [ ] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #1444, opened 2023-05-29
feat: update node version to 18

biruwon:

#### Summary

_This is a copy of the [past PR](https://github.com/netlify/zip-it-and-ship-it/pull/1138/files)_

Fixes the update for node js version to 18

I found another usage of node 16 [in here](https://github.com/netlify/zip-it-and-ship-it/blob/c8e1d4d64182ac4a71d0cf25cecdc1326b77c112/tsconfig.json#L6), but updating it breaks the tests. Does anyone knows why?

---

      passes our tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ The following properties are accepted:
 - `nodeVersion`
 
   - _Type_: `string`\
-  - _Default value_: `16`
+  - _Default value_: `18`
 
   The version of Node.js to use as the compilation target for bundlers. This is also used to determine the runtime
   reported in the manifest. Any valid and supported Node.js version is accepted. Examples:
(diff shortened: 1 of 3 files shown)


eduardoboucas:
> I found another usage of node 16 [in here](https://github.com/netlify/zip-it-and-ship-it/blob/c8e1d4d64182ac4a71d0cf25cecdc1326b77c112/tsconfig.json#L6), but updating it breaks the tests. Does anyone knows why?

That controls which version of Node we "compile" the package itself, which is independent from the version of Node we use in the functions the package generates. So I don't think we should be changing that as part of this PR.

eduardoboucas (approved):

PR #1460, opened 2023-06-12
feat: expose the api version in the results

danez:
#### Summary

Expose the property `runtimeAPIVersion` in the result of zip* and list* methods

This property has multiple meanings:

- list*
  - `undefined`: we do not know, because we did not check/parse the file
  - `2`: we detected version 2
  - `1`: not detected version 2 (might still be an invalid function)

- zip*
  - `undefined`: something went wrong, parsing error etc.
  - `2`: we detected version 2
  - `1`: not detected version 2 (might still be an invalid function)

- [ ] Docs in README.md


README.md
```diff
@@ -275,6 +283,16 @@ Additionally, the following properties also exist for Node.js functions:
...
+- `runtimeAPIVersion` `number | undefined`
+
+  When a function with the new API will be detected this is set to `2`. Otherwise it is set to `1`. `undefined` is only
+  returned in case of an error.
+
+- `runtimeVersion` `string | undefined`
+
+  Which exact runtime this function should be using. (e.g. `nodejs18.x`). This value is detected based on the input
+  `nodeVersion`.
+
```
danez, line 295: These are the only two actual changes in this file, the rest is formatting from prettier.

`runtimeVersion` was introduced in #1367, but never documented

src/main.ts
```diff
@@ -129,21 +130,22 @@ export const listFunctionsFiles = async function (
...
     displayName: config.name,
+    extension,
+    generator: config.generator,
     mainFile,
+    name,
     runtime: runtime.name,
-    extension,
+    runtimeAPIVersion: inSourceConfig ? inSourceConfig?.runtimeAPIVersion ?? 1 : undefined,
     schedule: inSourceConfig?.schedule ?? config.schedule,
-    generator: config.generator,
```
danez, line 146: sorting + `runtimeAPIVersion` added :)

src/utils/format_result.ts
```diff
@@ -15,6 +16,7 @@ export const formatZipResult = (archive: FunctionArchive) => {
     inSourceConfig: undefined,
     runtime: archive.runtime.name,
     schedule: archive.inSourceConfig?.schedule ?? archive?.config?.schedule,
+    runtimeAPIVersion: archive.inSourceConfig?.runtimeAPIVersion,
```
Skn0tt, line 19: Should there also be a fallback to `1` here?
```suggestion
    runtimeAPIVersion: archive.inSourceConfig?.runtimeAPIVersion ?? 1,
```

Skn0tt (approved):
Looks good, especially the broad test coverage! Left a small comment.

src/utils/format_result.ts
danez: No, the decision if it is 1 or 2 happens during parsing of the file. At this point if it is undefined it means we do not know the version, because of parsing errors or other things gone wrong. And i do not want to return a version that might be wrong.

PR #1461, opened 2023-06-12
fix: write bootstrap also when archive format is not zip

danez:

#### Summary

We did not write the bootstrap when the archiveFormat was not zip.

---

      passes our tests.

### DIFF
--- a/src/runtimes/node/utils/zip.ts
+++ b/src/runtimes/node/utils/zip.ts
@@ -53,7 +53,7 @@ interface ZipNodeParameters {
   srcFiles: string[]
 }
 
-const ensureBootstrapPresent = function (srcFiles: string[], aliases: Map<string, string>) {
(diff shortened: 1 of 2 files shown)


src/runtimes/node/utils/zip.ts
```diff
@@ -53,6 +53,16 @@ interface ZipNodeParameters {
   srcFiles: string[]
 }
 
+const ensureBootstrapPresent = function (srcFiles: string[], aliases: Map<string, string>) {
```
eduardoboucas, line 56: nit: A method starting with `ensure` makes me think that it'll check whether something exists and do something only if it doesn't (like `ensureDirectory`). That's not what's happening here, as we're always adding these files to the zip. I'd say a name like `addBootstrapFiles` would be more fitting.

eduardoboucas (approved):

eduardoboucas (approved):

PR #1472, opened 2023-06-23
feat: expose entryFilename

Skn0tt:

#### Summary

As proposed by @eduardoboucas in https://github.com/netlify/cli/pull/5804#discussion_r1238735216, this PR exposes the `entryFilename` for usage from the CLI.

---

      passes our tests.


src/runtimes/runtime.ts
```diff
@@ -54,6 +54,7 @@ export interface ZipFunctionResult {
   nodeModulesWithDynamicImports?: string[]
   path: string
   runtimeVersion?: string
+  entryFilename?: string
```
eduardoboucas, line 57: I think we should always be able to populate `entryFilename`. It should be either `___netlify-entry-point.mjs` or `<func name>.js`. If we always populate it, I think consuming this in the CLI will be easier.

Skn0tt: good point. done in [9e25e9a](https://github.com/netlify/zip-it-and-ship-it/pull/1472/commits/9e25e9abecd41fa016a02b6b503aeb6274e1cb3e). I'm unsure about the `entryFilename` for Go and Rust functions - my assumption is that we output an executable binary, so I set `entryFilename` to `""` - that way, `join(path, entryFilename)` will always point to the right file, independent of runtime.

eduardoboucas (approved):
🚀

Issue #1374, opened 2023-03-22
Functions depending on aws-sdk@2.x doesn't work on Lambda's Node.js 18.x runtime


evisong:
**- Do you want to request a _feature_ or report a _bug_?**
_bug_

**- What is the current behavior?**
In my function, one of the NPM dependencies depends on legacy `aws-sdk@2.x`, after bundled with `zip-it-and-ship-it` and deployed on AWS Lambda with Node.js 18.x runtime, the Lambda function fails to execute due to failing to find `aws-sdk`.

According to #538:

> The aws-sdk Node module is a special case. Because the module is injected automatically in the AWS Lambda environment, we want to leave any require/import calls untouched, but also we don't want to include its supporting files in the bundle.

However in AWS Lambda's recommended Node.js 18.x runtime, this is no longer true. Please refer to https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/:

> If your existing Lambda functions are using the included SDK v2, then you must update your function code to use the SDK v3 when upgrading to the Node.js 18 runtime. This is the recommended approach when upgrading existing functions to Node.js 18. Alternatively, you can use the Node.js 18 runtime without updating your existing code if you deploy the SDK v2 together with your function code.

So if I want to make use of Node.js 18.x but have to stick to legacy `aws-sdk`, I'll have to bundle it in the zip.

**- If the current behavior is a bug, please provide the steps to reproduce.**
1. Init an NPM project, add `aws-sdk` as a dependency;
2. Create an `functions/myfunc/index.mjs`, import `aws-sdk` and write whatever code to use it;
3. Bundle the project with `zip-it-and-ship-it functions dist`;
4. Create a Lambda function on AWS, configure with its recommended Node.js 18.x runtime;
5. Upload the `dist/myfunc.zip` to Lambda, and test;
6. The Lambda function fails with the following error:

```
2023-03-22T04:38:14.971Z	undefined	ERROR	Uncaught Exception 	
{
    "errorType": "Runtime.ImportModuleError",
    "errorMessage": "Error: Cannot find module 'aws-sdk'\nRequire stack:\n- /var/task/index.js\n- /var/task/myfunc.js\n- /var/runtime/index.mjs",
    "stack": [
        "Runtime.ImportModuleError: Error: Cannot find module 'aws-sdk'",
        "Require stack:",
        "- /var/task/index.js",
        "- /var/task/myfunc.js",
[...]

PR #1481, opened 2023-06-28
feat: correctly account for AWS SDK v3

Skn0tt:

#### Summary

Closes https://github.com/netlify/zip-it-and-ship-it/issues/1374.

---

      passes our tests.


tests/main.test.ts
```diff
@@ -195,24 +195,85 @@ describe('zip-it-and-ship-it', () => {
...
+            '*': {
+              ...options.config['*'],
+              nodeVersion: '16',
+            },
+          },
+        })
+        const { files } = await zipNode(fixtureName, { opts })
+
+        const [{ unzipPath, bundler, entryFilename }] = files
+        if (bundler === 'esbuild') {
```
Skn0tt, line 256: The previous test case silently ignored this case. It still passed, since we only asserted on a file being *nonexistant*. But now we're also asserting on a different file existing.

src/runtimes/node/utils/node_version.ts
```diff
@@ -14,6 +15,7 @@ export const getNodeSupportMatrix = (configVersion?: string): NodeVersionSupport
 
   return {
     esm: versionNumber >= 14,
+    awsSDKV3: versionNumber >= 18,
```
eduardoboucas, line 18: It feels a bit weird to me to be using the Node support matrix to signal whether we need to bundle a specific version of the AWS SDK, because that has nothing to do with Node itself, just a Lambda implementation detail.

Skn0tt: Yeah, I get that - but at the other hand, it is a Node.js-only issue, and it's totally dependent on the version of Node.js. That's why I put it into here. But if you prefer putting it into a different function name, I can refactor it 🤷

eduardoboucas: Up to you. If you think this simplifies things, let's keep it.

Regardless, it feels that making this a `number` rather than `boolean` (e.g. `awsSDKVersion: 3`) would give us a bit more flexibility in case Lambda ships a future version of Node with yet another major version of their SDK? With the current implementation we'd need to keep adding more properties and end up with a potential inconsistent state?

Skn0tt: Agree, but let's make that update when AWS ships the next version.

eduardoboucas (approved):

Issue #1487, opened 2023-07-04
Auto-enable `zisi_pure_esm` and `zisi_pure_esm_mjs` when v2 API is detected


Skn0tt:
When the v2 API is detected, we should automatically enable these two featureflags.

PR #1489, opened 2023-07-04
fix: auto-enable zisi_pure_esm, zisi_pure_esm_mjs when v2 api is used

Skn0tt:
#### Summary

Fixes https://github.com/netlify/zip-it-and-ship-it/issues/1487

---

      passes our tests.


src/runtimes/node/index.ts
```diff
@@ -62,6 +65,12 @@ const zipFunction: ZipFunction = async function ({
 
   const inSourceConfig = await findISCDeclarationsInPath(mainFile, name, featureFlags)
   const runtimeAPIVersion = inSourceConfig.runtimeAPIVersion === 2 ? 2 : 1
+
+  if (runtimeAPIVersion === 2) {
```
eduardoboucas, line 69: Rather than overriding the value of feature flags if we detect the V2 API, I think we should take the value of `runtimeAPIVersion` into account when deciding the module format to use.

I think this was already in place for NFT (https://github.com/netlify/zip-it-and-ship-it/blob/main/src/runtimes/node/bundlers/nft/es_modules.ts#L80-L84), but I believe you changed it to use esbuild instead, so we lost that?

Skn0tt: Good idea! done in https://github.com/netlify/zip-it-and-ship-it/pull/1489/commits/a0ad3a3d90613bb7cdf27ac5c449e70c8345a89e

eduardoboucas (approved):

PR #1499, opened 2023-07-10
feat: log to system-log if v2 function found

danez:
Logs to systemLog if the feature flag is on and we find a v2 function.


src/utils/logger.ts
```diff
@@ -1,7 +1,7 @@
 // adapted from https://github.com/netlify/edge-bundler/blob/main/node/logger.ts
 type LogFunction = (...args: unknown[]) => void
 
-const noopLogger: LogFunction = () => {
+export const noopLogger: LogFunction = () => {
```
danez, line 4: Instead of using `getLogger()` from below, I find it way cleaner to directly use the noop logger in the function declaration

```ts
function myfunction (systemLog?: Logfunction = noopLogger) {}
```

eduardoboucas (approved):

PR #1505, opened 2023-07-13
fix: output mjs and ts v2 functions with mjs extension

danez:
#### Summary

Fixes https://github.com/netlify/pod-dev-foundations/issues/559

This also fixes the tests, see inline comments


tests/v2api.test.ts
```diff
@@ -96,6 +110,10 @@ describe.runIf(semver.gte(nodeVersion, '18.13.0'))('V2 functions API', () => {
         }),
       })
 
+      for (const entry of files) {
+        expect(entry.runtimeAPIVersion).toBe(2)
+      }
```
danez, line 115: not related to the fix, but ensure that the `runtimeAPIVersion` returned by zipFunctions is 2.

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -107,18 +107,18 @@ export const bundleJsFile = async function ({
...
-    let __dirname=___nfyPathDirname(___nfyFileURLToPath(import.meta.url));
-    let require=___nfyCreateRequire(import.meta.url);
-  `
+import {createRequire as ___nfyCreateRequire} from "module";
+import {fileURLToPath as ___nfyFileURLToPath} from "url";
+import {dirname as ___nfyPathDirname} from "path";
+let __filename=___nfyFileURLToPath(import.meta.url);
+let __dirname=___nfyPathDirname(___nfyFileURLToPath(import.meta.url));
+let require=___nfyCreateRequire(import.meta.url);
+`
```
danez, line 121: Remove unnecessary whitespace from bundle.

src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -102,9 +104,10 @@ const traceFilesAndTranspile = async function ({
   name: string
   runtimeAPIVersion: number
 }) {
-  const isTypeScript = extname(mainFile) === '.ts'
+  const isTypeScript = tsExtensions.has(extname(mainFile))
   const tsAliases = new Map<string, string>()
   const tsRewrites = new Map<string, string>()
+  const tsFormat = runtimeAPIVersion === 2 ? MODULE_FORMAT.ESM : MODULE_FORMAT.COMMONJS
```
eduardoboucas, line 110: Still planning on reading from `tsconfig.json` in a follow-up.

Skn0tt (dismissed):

package.json
```diff
@@ -66,6 +66,7 @@
     "execa": "^6.0.0",
     "filter-obj": "^5.0.0",
     "find-up": "^6.0.0",
+    "get-tsconfig": "^4.6.2",
```
Skn0tt, line 69: nice, this package looks like it's solving sooo much for us! I got a bit suspicious because it says in the README that it "validates" the tsconfig (which we wouldn't want to do here, I think) - but I couldn't find anything about that in the code.

src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -105,9 +106,9 @@ const traceFilesAndTranspile = async function ({
   runtimeAPIVersion: number
 }) {
   const isTypeScript = tsExtensions.has(extname(mainFile))
+  const tsFormat = isTypeScript ? getTSModuleFormat(mainFile) : MODULE_FORMAT.COMMONJS
```
Skn0tt, line 109: it's confusing to me that a variable called `tsFormat` has a case for when the file isn't `ts`. in the case of `!isTypeScript`, this value will be unused if i'm reading this right. does that make sense?

src/runtimes/node/utils/tsconfig.ts
```diff
@@ -0,0 +1,19 @@
...
+export const getModuleFormat = (path: string) => {
+  const file = getTsconfig(path)
+
+  if (!file) {
+    return MODULE_FORMAT.COMMONJS
+  }
+
+  const moduleProp = file.config.compilerOptions?.module?.toLowerCase() ?? ''
+
+  return esmModuleValues.has(moduleProp) ? MODULE_FORMAT.ESM : MODULE_FORMAT.COMMONJS
```
Skn0tt, line 18: According to https://www.typescriptlang.org/tsconfig#module, it's a bit more complicated I think:

Screenshot 2023-08-02 at 15 43 59

It looks like we also need to emit ESM for `es6` and `es2015`. Maybe it's easier to phrase this in terms of when we need to emit CJS?

src/runtimes/node/bundlers/nft/index.ts
eduardoboucas: It's because we're checking the value of `isTypeScript` twice: once here, and then again when it comes to returning the result (https://github.com/netlify/zip-it-and-ship-it/blob/220b0b03e0ec964635ed3245c206304bf6745022/src/runtimes/node/bundlers/nft/index.ts#L169). If we leave `tsFormat` as `undefined`, the compiler will complain that we're missing a value for `moduleFormat`, unaware that if we're in that branch we've already set a value for `tsFormat`.

The quick way around it was to have a fallback for `tsFormat` here, although we'll never use it.

If there's a more elegant way of doing it, I'm happy to change.

src/runtimes/node/utils/tsconfig.ts
```diff
@@ -0,0 +1,19 @@
+import { getTsconfig } from 'get-tsconfig'
```
lukasholzer, line 1: nice that it resolves extends!

src/runtimes/node/utils/tsconfig.ts
lukasholzer: yea if `module` is es2015 then it results in esm as well:
https://www.typescriptlang.org/play?target=99&moduleResolution=99&module=5&jsx=0#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVwBbAA2hEAFAJRwDeAUHHGjgM4SHAB0JA5qZgLaRY7PsBgBDdkiiFyNAL5A
CleanShot 2023-08-02 at 17 16 32

src/runtimes/node/utils/zip.ts
```diff
@@ -229,6 +238,58 @@ const addStat = async function (cache: RuntimeCache, srcFile: string) {
...
+  mainFile: string
+  name: string
+  userNamespace: string
+}) {
+  const functionDirectory = dirname(mainFile)
+
+  let packageJSON: PackageJsonFile | null = null
+
+  try {
+    packageJSON = await getClosestPackageJson(functionDirectory)
```
lukasholzer, line 259: do we stop at the repository root/site root here?

eduardoboucas: Good question. I don't think that `get-tsconfig` has any option to define boundaries, so I think we'll need to look at the path it finds and assert whether it's inside or outside the repository root. I'll do that in a bit.

src/runtimes/node/utils/tsconfig.ts
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/1505/commits/217633cb74db61f145105a820d16a91405eb3692.

src/runtimes/node/utils/zip.ts
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/1505/commits/ba79d06bbe0ab7b4e167ca621ab2645e0abe6bbb.

src/runtimes/node/utils/tsconfig.ts
eduardoboucas: Yeah, it's pretty handy!

src/runtimes/node/utils/zip.ts
eduardoboucas: Wait, I've done this for the `tsconfig.json`, but your comment is about `package.json`. Let me to do that here too.

eduardoboucas: Done that in https://github.com/netlify/zip-it-and-ship-it/pull/1505/commits/71bb5eee988c6a8199947441a5a49b3dd9b02e60.

src/runtimes/node/utils/zip.ts
```diff
@@ -48,6 +51,8 @@ interface ZipNodeParameters {
   filename: string
   mainFile: string
   moduleFormat: ModuleFormat
+  name: string
+  repositoryRoot?: string
```
lukasholzer, line 55: we have to pass this than in the CLI, right?

lukasholzer (approved):
Awesome! Great Job 🥳 
There is so much 🧠 work in this!

src/runtimes/node/utils/zip.ts
eduardoboucas: Netlify Build is already passing it down to ZISI: https://github.com/netlify/build/blob/a20207ec7cdc1101a875d9411355bd48f12733a3/packages/build/src/plugins_core/functions/zisi.ts#L68

PR #1516, opened 2023-08-03
feat: extend `package.json` logic to `archiveFormat: "none"`

eduardoboucas:
#### Summary

This is a follow-up to #1505 extending the logic for planting a `package.json` file with `{"type": "module"}` to the generation of a directory (as opposed to a ZIP file), which is used for local development in the CLI.

### DIFF
--- a/src/runtimes/node/utils/zip.ts
+++ b/src/runtimes/node/utils/zip.ts
@@ -128,18 +128,15 @@ const createDirectory = async function ({
 
   if (tsExtensions.has(extension) && moduleFormat === MODULE_FORMAT.ESM && runtimeAPIVersion === 2) {
     const packageJSON = await ensurePackageJSONWithModuleType({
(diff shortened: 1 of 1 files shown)


src/runtimes/node/utils/zip.ts
```diff
@@ -291,7 +317,10 @@ const ensurePackageJSONWithModuleType = async function ({
   const normalizedPath = normalizeFilePath({ commonPrefix: basePath, path, userNamespace })
   const contents = JSON.stringify({ type: 'module' })
 
-  addZipContent(archive, contents, normalizedPath)
+  return {
```
eduardoboucas, line 320: I've changed this function such that instead of adding the file to the ZIP, it simply returns its content and path. That way, the callers (`createDirectory` and `createZipArchive`) have the power to do what they want with the file.

PR #1520, opened 2023-08-09
feat: log dynamic import expressions

eduardoboucas:
#### Summary

As of version 0.19.0, [esbuild can now process some dynamic imports](https://esbuild.github.io/api/#glob), albeit just a subset of the expressions our custom plugin can process. I'd like to understand how many of the expressions people are using in the wild would be supported natively by esbuild, with the view of decommissioning our fork entirely.

With https://github.com/netlify/zip-it-and-ship-it/pull/1445, we started logging any instances where the dynamic imports plugin from our esbuild fork was used to process a dynamic import expression, but we're logging the paths that will be included in the bundle, which is not the most useful piece of informaiton.

This PR changes that so we log the actual dynamic import expression that was used. We can then compare those against the list of expressions that esbuild supports.

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -1569,13 +1569,25 @@ describe('zip-it-and-ship-it', () => {
     expect(() => func('two')).toThrowError()
     expect(files[0].nodeModulesWithDynamicImports).toHaveLength(0)
 
-    const pathsFromDynamicImports = [
-      join('node_modules', '@org', 'test', 'files', '**'),
-      join('node_modules', '@org', 'test', 'files', '**.js'),
-    ]
-
     expect(systemLog).toHaveBeenCalledWith(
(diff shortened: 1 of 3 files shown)


tests/main.test.ts
```diff
@@ -1569,13 +1569,25 @@ describe('zip-it-and-ship-it', () => {
...
     expect(files[0].nodeModulesWithDynamicImports).toHaveLength(0)
 
-    const pathsFromDynamicImports = [
-      join('node_modules', '@org', 'test', 'files', '**'),
-      join('node_modules', '@org', 'test', 'files', '**.js'),
-    ]
-
     expect(systemLog).toHaveBeenCalledWith(
-      `Functions bundling included paths by parsing dynamic import: ${pathsFromDynamicImports.join(', ')}`,
+      // eslint-disable-next-line no-template-curly-in-string
```
lukasholzer, line 1573: I would burn this lint rule 🔥

PR #1523, opened 2023-08-11
feat: add support for custom routes

eduardoboucas:
#### Summary

Reads the `path` in-source configuration property and writes the resulting routes into a `routes` field in the manifest. It only does so for v2 functions.

Part of https://github.com/netlify/pod-dev-foundations/issues/572.


src/utils/urlpattern.ts
```diff
@@ -0,0 +1,7 @@
+import { URLPattern } from 'urlpattern-polyfill'
```
eduardoboucas, line 1: Taken from https://github.com/netlify/edge-bundler/blob/main/node/utils/urlpattern.ts.

tests/fixtures-esm/v2-api-with-path/with-named-group.mjs
```diff
@@ -0,0 +1,10 @@
+export default async () =>
+  new Response('<h1>Hello world</h1>', {
+    headers: {
+      'content-type': 'text/html',
+    },
+  })
+
+export const config = {
+  path: '/products/:id',
```
lukasholzer, line 9: do we want to add a test for multiple named groups as well?

like `/products/:category/:variant/:sku`

lukasholzer (approved):
Nothing blocking but I would add additional tests for the `path` property covering some more scenarios like:

- invalid url patterns 
- paths starting without a slash
- longer nested URLs
- deep nested URLs like `/product/:id/view`

tests/fixtures-esm/v2-api-with-path/with-named-group.mjs
eduardoboucas: Great idea. Done in https://github.com/netlify/zip-it-and-ship-it/pull/1523/commits/b7fbde318debce338ca3a4a1193ece0226ca6271.

lukasholzer (approved):
Awesome!

PR #1530, opened 2023-08-16
fix: tighten up V2 API detection mechanism

eduardoboucas:
#### Summary

This is a small tweak to the V2 API detection mechanism. To make sure we reduce false positives during the initial phase of the rollout, we should adopt the V1 API as long as the entry file has a `handler` export.

This was always the intention as shown by [this check](https://github.com/netlify/zip-it-and-ship-it/blob/e7ac14fb5b8a7745fe59bbc370d7ae4faf296b28/src/runtimes/node/in_source_config/index.ts#L82), but the `handlerExports` is a bit misleading. Right now, it's only finding exports that are function calls, because of [this](https://github.com/netlify/zip-it-and-ship-it/blob/e7ac14fb5b8a7745fe59bbc370d7ae4faf296b28/src/runtimes/node/parser/exports.ts#L186-L190).

With this PR we start capturing all exports and categorising them based on their type.

### DIFF
--- a/src/runtimes/node/in_source_config/index.ts
+++ b/src/runtimes/node/in_source_config/index.ts
@@ -153,7 +153,9 @@ export type ISCExportWithCallExpression = {
 }
 export type ISCExportWithArrowFunctionExpression = { type: 'arrow-function-expression' }
 export type ISCExportWithFunctionExpression = { type: 'function-expression' }
+export type ISCExportOther = { type: 'other' }
 export type ISCExport =
   | ISCExportWithArrowFunctionExpression
   | ISCExportWithCallExpression
   | ISCExportWithFunctionExpression
+  | ISCExportOther
--- a/src/runtimes/node/in_source_config/index.ts
+++ b/src/runtimes/node/in_source_config/index.ts
@@ -97,7 +97,12 @@ export const findISCDeclarations = (
   }
 
   const iscExports = handlerExports
-    .map(({ args, local: exportName }) => {
+    .map((exp) => {
+      if (exp.type !== 'call-expression') {
+        return null
+      }
+
+      const { args, local: exportName } = exp
       const matchingImport = imports.find(({ local: importName }) => importName === exportName)
 
       if (matchingImport === undefined) {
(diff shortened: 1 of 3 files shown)


src/runtimes/node/in_source_config/index.ts
```diff
@@ -97,7 +97,12 @@ export const findISCDeclarations = (
   }
 
   const iscExports = handlerExports
-    .map(({ args, local: exportName }) => {
+    .map((exp) => {
+      if (exp.type !== 'call-expression') {
```
Skn0tt: just as a note to myself: limiting this to `call-expression` here is fine, because it's not the immediate `export` node, but the resolved value that's being exported. so the code from https://docs.netlify.com/functions/scheduled-functions/ will still work with this change:

```ts
import type { Handler, HandlerEvent, HandlerContext } from "@netlify/functions";
import { schedule } from "@netlify/functions";

const myHandler: Handler = async (event: HandlerEvent, context: HandlerContext) => {
  console.log("Received event:", event);

    return {
        statusCode: 200,
    };
};

const handler = schedule("@hourly", myHandler)

export { handler };```

**Review by Skn0tt — APPROVED** [review]
Looks good!

File: src/runtimes/node/in_source_config/index.ts (hunk 1 above)
Comment (eduardoboucas) [reply src/runtimes/node/in_source_config/index.ts]: That's right! I've pushed https://github.com/netlify/zip-it-and-ship-it/pull/1530/commits/389445509c91344a3951f60b7edc1aa180535e04 with a couple of tests to prove that.

Comment (eduardoboucas) [reply src/runtimes/node/in_source_config/index.ts]: Wait, something is broken ⏳

Comment (eduardoboucas) [reply src/runtimes/node/in_source_config/index.ts]: Ah, the tests were failing because this doesn't work in CJS, only the ESM counterpart:

```js
const { schedule } = require("@netlify/functions")

const handler = schedule("@daily", () => {})

exports.handler = handler
```

But this also doesn't work in current `main` because we only get `getExportsFromBindings` on the ESM flow, for some reason: https://github.com/netlify/zip-it-and-ship-it/blob/9c24f23ea2acf2cc258914f37a5491c89107cf81/src/runtimes/node/parser/exports.ts#L83.

We can fix that, but it's neither urgent nor part of this PR. I've skipped that new test for now.

minivan (approved):

PR #1532, opened 2023-08-17
feat: use upstream ESBuild instead of Netlify Fork

Skn0tt:

#### Summary

See https://github.com/netlify/pod-dev-foundations/issues/576.

This PR will switch ZISI from using our own fork of esbuild to the upstream version of it. To ensure no breaking changes, this PR starts with two code branches for the fork + upstream, and we'll run our full test suite against both. Then we'll add more test cases to increase our confidence that there's no breaking changes between the two. Because we don't want to ship both at the same time (it'd mean >10mb more dependencies in the Netlify CLI), i'll remove the old code path before merging.

---

      passes our tests.


eduardoboucas:
I'm a bit confused. If the idea is to have only the upstream version before merging, why not do that right away and see if the tests pass? We know that tests pass with the current version, so I'm not sure I understand what we get from testing both here?

Skn0tt:
I agree that in the current state, this is nonsense. But i'm planning on adding more tests to cover more of the scenarios we're seeing in the wild in the coming days. To work on those tests, it's good to see both versions running side-by-side.

eduardoboucas:
Ah, got it! Makes sense. 👍🏻

tests/esbuild_migration.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    const {
+      files: [{ unzipPath, entryFilename }],
+    } = await zipNode('cardinal-require-terminator', { opts })
+    const res = await importFunctionFile(`${unzipPath}/${entryFilename}`)
+    expect(res).toEqual('Cardinal is unavailable!')
+  })
+
+  /**
+   * We're seeing one site do `require(`../../data/${req.params.foo}/${req.params.bar}`)`.
+   * It's not supported on our fork, but it is on upstream!
```
Skn0tt, line 93: The fact that it's supported by upstream means that this test will fail in CI. If you see that, it's expected.

Skn0tt:
I believe I covered all relevant import patterns. After our small workaround, we're now on-par with the fork, and can move forward with this.

After we've reviewed the test suite, i'll remove the code path for the fork, so we only ship a single esbuild when we merge this.

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -147,6 +155,31 @@ let require=___nfyCreateRequire(import.meta.url);
       srcFile,
       outputExtension,
     })
+
+    for (const [filename] of Object.entries(metafile.outputs)) {
+      const content = await readFile(filename, { encoding: 'utf-8' })
```
eduardoboucas, line 160: Rather than esbuild writing to disk and then us reading + rewriting the files, you could [set `write: false`](https://esbuild.github.io/api/#write) to make it return the contents, which you can then patch and write to disk just once.

On the other hand, the current approach feels less intrusive, in the sense that (hopefully) a fix for this is added upstream and we can just remove this whole block without changing the way files are written.

No strong opinion, just wanted to present that option.

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -147,6 +155,31 @@ let require=___nfyCreateRequire(import.meta.url);
       srcFile,
       outputExtension,
     })
+
+    for (const [filename] of Object.entries(metafile.outputs)) {
```
eduardoboucas, line 159: I think `metafile.outputs` will contain all the files written by esbuild, which can include a source map. It would be worth checking the extension and limit this operation to JavaScript files.

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -147,6 +155,31 @@ let require=___nfyCreateRequire(import.meta.url);
...
+      const content = await readFile(filename, { encoding: 'utf-8' })
+
+      const updated = content.replace(
+        `
+var __glob = (map) => (path) => {
```
eduardoboucas, line 164: If there is a symbol called `__glob` defined in user space (whether by a function directly or by a dependency), esbuild will rename this symbol to one that is available, following the format `__glob2`, `__glob3`, etc. If that happens, we won't match and will fail to patch this function.

We could potentially solve this by passing a regular expression to `content.replace`, probably at the cost of a more expensive matching.

This feels like an edge case though, so I understand if you don't want to address it.

eduardoboucas (commented):
Great work! Left a few comments, but none are blocking. Not approving because I know you still need to do the work of removing the old version.

One small nit: I think the project is capitalised as "esbuild", not "ESBuild". 🤓

package.json
```diff
@@ -63,6 +63,7 @@
     "common-path-prefix": "^3.0.0",
     "cp-file": "^10.0.0",
     "es-module-lexer": "^1.0.0",
+    "esbuild": "^0.19.2",
```
lukasholzer, line 66: Do we want to restrict it to the patch version as esbuild does breaking changes in minor versions? like `~0.19.2`

src/feature_flags.ts
```diff
@@ -33,6 +33,8 @@ export const defaultFlags = {
 
   // Create unique entry file instead of a file that is based on the function name.
   zisi_unique_entry_file: false,
+
+  zisi_esbuild_upstream: false,
```
lukasholzer, line 37: we can leverage real feature flags from devcycle as well no need for defining constants here

src/runtimes/node/bundlers/esbuild/bundler.ts
lukasholzer: you could use a variable name that is so unlikely to be used including emojis. `__glob💩-thingy`

Skn0tt: Could you share some code where this happens? I've added a test on this in [2dd12bb](https://github.com/netlify/zip-it-and-ship-it/pull/1532/commits/2dd12bb17043b3521a81123c7c5b3c4f0aa6b5b9), and it doesn't break. Taking a look at the output, it looks like it renames the user-defined `__glob`, not its own one:

Screenshot 2023-08-24 at 13 02 36

package.json
Skn0tt: That's a good idea! I'll lock it down to an exact version, no need of ambiguity here.

Skn0tt: Done in [2dd12bb](https://github.com/netlify/zip-it-and-ship-it/pull/1532/commits/2dd12bb17043b3521a81123c7c5b3c4f0aa6b5b9).

src/runtimes/node/bundlers/esbuild/bundler.ts
eduardoboucas: Ah, I was mistaken. I was convinced that esbuild would keep the name of the user symbol and rename the shim, but it's the other way around. Ignore me!

tests/esbuild_migration.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  testMany('template strings', esbuildConfigs, async (options) => {
+    const fixtureName = 'require-template-string'
+    const opts = merge(options, {
+      basePath: join(FIXTURES_DIR, fixtureName),
+    })
+    await zipNode(fixtureName, { opts })
+  })
+
+  /**
+   * If user code defines `__glob` in global scope, esbuild will rename its `__glob` helper.
```
eduardoboucas, line 108: Based on your findings, esbuild will NOT rename its helper, but will rename the user symbol?

src/runtimes/node/bundlers/esbuild/bundler.ts
Skn0tt: I like that the current approach is less intrusive. Let's stick with that for now.

src/runtimes/node/bundlers/esbuild/bundler.ts
Skn0tt: Good idea, done in https://github.com/netlify/zip-it-and-ship-it/pull/1532/commits/695f62241865f52faae5108e9fd6311c9cb1ccfd.

tests/esbuild_migration.test.ts
Skn0tt:
```suggestion
   * If user code defines `__glob` in global scope, esbuild will rename the user symbol, not its helper.
```

Skn0tt:
Ha, found the fix - the test is matching on `_esbuild`, which we're incorrectly appending: https://github.com/netlify/zip-it-and-ship-it/blob/695f62241865f52faae5108e9fd6311c9cb1ccfd/tests/main.test.ts#L2193 That's why it fails on `bundler_none_upstream_esbuild`.

I'll start with rewriting the PR now!

Skn0tt:
Et voila, esbuild upstream is ready!

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -58,34 +69,16 @@ export const bundleJsFile = async function ({
   const targetDirectory = await tmpName()
 
   // De-duping external and ignored modules.
-  const external = [...new Set([...externalModules, ...ignoredModules])]
+  const external = [
+    ...new Set([...externalModules, ...ignoredModules, ...includedFilesToEsbuildExternals(config.includedFiles ?? [])]),
```
eduardoboucas, line 73: Could you add some context on why we're now adding paths from `includedFiles` into esbuild's `external`?

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -135,7 +128,8 @@ let require=___nfyCreateRequire(import.meta.url);
       outdir: targetDirectory,
       outExtension: { '.js': outputExtension },
       platform: 'node',
-      plugins,
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      plugins: plugins as any,
```
eduardoboucas, line 132: Can't we just do this?

```suggestion
      plugins,
```

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -147,9 +141,40 @@ let require=___nfyCreateRequire(import.meta.url);
...
+          )
+
+          await writeFile(filename, updated, { encoding: 'utf-8' })
+        }),
+    )
+
     const inputs = Object.keys(metafile.inputs).map((path) => resolve(path))
     const cleanTempFiles = getCleanupFunction([...bundlePaths.keys()])
-    const additionalPaths = [...dynamicImportsIncludedPaths, ...includedFilesFromModuleDetection]
+    const additionalPaths = [...includedFilesFromModuleDetection]
```
eduardoboucas, line 177: Not sure we need to create a new array now that `includedFilesFromModuleDetection` is the only source?

eduardoboucas (commented):
OMG I'm so happy about this.

I left a couple of comments, but generally it's looking great.

eduardoboucas (dismissed):
🎉 🎉 🎉 

Great work!

mraerino (approved):

PR #1550, opened 2023-08-31
fix: remove zisi_functions_api_v2 flag

Skn0tt:
This flag has been rolled out for a while now, without any problems. Let's remove it, so the functionality becomes available in the CLI as well.

Part of https://github.com/netlify/pod-dev-foundations/issues/578.


eduardoboucas (approved):
🎉

PR #1557, opened 2023-09-04
fix: resolve ignoredFiles with multiple globs on behalf of esbuild

Skn0tt:

#### Summary

Follow-Up on https://github.com/netlify/zip-it-and-ship-it/pull/1546. Instead of warning on double-glob patterns, we resolve them using `glob`.

Part of https://github.com/netlify/pod-dev-foundations/issues/576.

---

      passes our tests.


src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -29,45 +29,49 @@ const RESOLVE_EXTENSIONS = ['.js', '.jsx', '.mjs', '.cjs', '.ts', '.tsx', '.mts'
...
-    // only keep negated patterns
+const includedFilesToEsbuildExternals = async (includedFiles: string[], baseDir: string) => {
+  const exclusions = includedFiles
     .filter((pattern) => pattern.startsWith('!'))
-    // warn about invalid patterns
-    .filter((pattern) => {
-      const containsMultipleStars = pattern.indexOf('*') !== pattern.lastIndexOf('*')
+    .map((pattern) => pattern.slice(1))
+    // esbuild expects relative paths
+    .map((pattern) => `./${pattern}`)
```
eduardoboucas, line 41: Is this going to work on Windows?

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -29,45 +29,49 @@ const RESOLVE_EXTENSIONS = ['.js', '.jsx', '.mjs', '.cjs', '.ts', '.tsx', '.mts'
...
 
-      if (containsMultipleStars) {
-        const warning = `Found invalid pattern in includedFiles: ${pattern}. Only one * is allowed when using esbuild. See https://ntl.fyi/esbuild-double-glob for more.`
+  // esbuild treats * the same as glob treats **, so this replacement is safe
+  const singleStar = exclusions.map((pattern) => pattern.replace('**', '*').replace(/\*(\\\*)+/, '*'))
```
eduardoboucas, line 44: nit: I find this variable a bit misleading, because it doesn't necessarily have anything to do with single stars. It's just a version of `exclusions` with normalised paths. Could we do this in another `.map` in the `exclusions` declaration above, as opposed to creating a new variable?

eduardoboucas (dismissed):

src/runtimes/node/bundlers/esbuild/bundler.ts
Skn0tt: Didn't really think about that, but I guess it does, yeah - the unit tests are all green, also on windows.

src/runtimes/node/bundlers/esbuild/bundler.ts
Skn0tt: The idea is that this replaces all occurences of `**` with `*`, and all occurences of `*/*` with `*`. So `singleStar` is intended to mean "it only contains singular globs", not "it only contains one glob". I wanted the code to explain that somewhere, so I added the variable and the comment - but it reads like you interpreted it differently, so that didn't work. How would you phrase this?

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -29,45 +29,49 @@ const RESOLVE_EXTENSIONS = ['.js', '.jsx', '.mjs', '.cjs', '.ts', '.tsx', '.mts'
...
-    .map((pattern) => pattern.slice(1))
-    // esbuild expects a relative path for local files
-    .map((pattern) => `./${pattern}`)
-    // esbuild doesn't like **
-    .map((pattern) => pattern.replace(/\*\*/g, '*'))
+      result.push(moduleName)
+      continue
+    }
+
+    const hasMultipleGlobs = pattern.indexOf('*') !== pattern.lastIndexOf('*')
```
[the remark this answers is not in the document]
Skn0tt: See above - all occurences of `**` should be replaced with `*` at this point, so it should be working as intended.

src/runtimes/node/bundlers/esbuild/bundler.ts
eduardoboucas: I would add a comment explaining what you said just now, but keep it as part of the group of maps above.

```js
  const exclusions = includedFiles
    .filter((pattern) => pattern.startsWith('!'))
    .map((pattern) => pattern.slice(1))
    // esbuild expects relative paths
    .map((pattern) => `./${pattern}`)
    // replace all occurences of ** with *, and all occurences of */* with *
    .map((pattern) => pattern.replace('**', '*').replace(/\*(\\\*)+/, '*'))
```

src/runtimes/node/bundlers/esbuild/bundler.ts
```diff
@@ -29,45 +29,48 @@ const RESOLVE_EXTENSIONS = ['.js', '.jsx', '.mjs', '.cjs', '.ts', '.tsx', '.mts'
...
-        if (failOnDouble) {
-          throw new FunctionBundlingUserError(warning, {
-            functionName,
-            runtime: RUNTIME.JAVASCRIPT,
-            bundler: NODE_BUNDLER.ESBUILD,
-          })
-        }
+  for (const pattern of exclusions) {
+    // esbuild expects modules to be passed in as module names, not paths
+    const nodeModulesMatch = pattern.match(/^\.\/node_modules\/(([^/]+)(\/[^/]*)?)*/)
```
eduardoboucas, line 49: If I understand the logic correctly, something like `!node_modules/react-native/**/*` would have been transformed into `./node_modules/react-native/*` by the time it gets here?

If that's true, then I think the module name should come from the 3rd position in the match, not the 2nd?

https://jsbin.com/zamayipaka/edit?js,console

Skn0tt: Hmmm - I also want it to work for `@react-native/foo`. I believe the edit in https://github.com/netlify/zip-it-and-ship-it/pull/1557/commits/37450e26bca6235da2d73840d01e4d28463e2904 fixes that, at least it looks right when I dabble with it in https://jsbin.com/wanonatoro/edit?js,console :)

eduardoboucas (approved):

PR #1559, opened 2023-09-05
feat: support go provided.al2 runtime

Skn0tt:

#### Summary

AWS will deprecate the `go1.x` runtime by end of year, asking customers to deploy to `provided.al2` going forward. It requires ZIPs to contain a single file called `bootstrap`, similar to Rust. This PR adds a featureflag to ZISI that makes that switch, and writes the result of it into `manifest.json` so we can detect this during deployment.

Part of https://github.com/netlify/pod-dev-foundations/issues/581.

---

      passes our tests.

### DIFF
--- a/src/runtimes/go/index.ts
+++ b/src/runtimes/go/index.ts
@@ -113,6 +113,7 @@ const zipFunction: ZipFunction = async function ({
   srcPath,
   stat,
   isInternal,
+  featureFlags,
 }) {
   const destPath = join(destFolder, filename)
   const isSource = extname(mainFile) === '.go'
@@ -139,13 +140,18 @@ const zipFunction: ZipFunction = async function ({
     const zipPath = `${destPath}.zip`
     const zipOptions = {
       destPath: zipPath,
-      filename: basename(destPath),
+      filename: featureFlags.zisi_golang_use_al2 ? 'bootstrap' : basename(destPath),
       runtime,
     }
 
     await zipBinary({ ...zipOptions, srcPath: binary.path, stat: binary.stat })
 
-    return { config, path: zipPath, entryFilename: zipOptions.filename }
+    return {
+      config,
+      path: zipPath,
+      entryFilename: zipOptions.filename,
+      goUseAL2Runtime: featureFlags.zisi_golang_use_al2,
+    }
   }
 
   // We don't need to zip the binary, so we can just copy it to the right path.
--- a/src/runtimes/go/index.ts
+++ b/src/runtimes/go/index.ts
@@ -150,7 +150,7 @@ const zipFunction: ZipFunction = async function ({
       config,
       path: zipPath,
       entryFilename: zipOptions.filename,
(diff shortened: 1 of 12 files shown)


src/runtimes/go/index.ts
```diff
@@ -139,13 +140,18 @@ const zipFunction: ZipFunction = async function ({
...
     }
 
     await zipBinary({ ...zipOptions, srcPath: binary.path, stat: binary.stat })
 
-    return { config, path: zipPath, entryFilename: zipOptions.filename }
+    return {
+      config,
+      path: zipPath,
+      entryFilename: zipOptions.filename,
+      goUseAL2Runtime: featureFlags.zisi_golang_use_al2,
```
eduardoboucas, line 153: What will we do with this property?

Skn0tt: We'll read it in open-api, so we can propagate it through nf-server into BitBalloon and functions-origin. They can then read it and use `provided.al2` as the AWS runtime, if it's present. If you have an idea for a solution where we don't need to propagate this so far, i'd be happy to hear it!

eduardoboucas: Could we use the existing `runtimeVersion` property? https://github.com/netlify/zip-it-and-ship-it/blob/626d63d4de399693d5f1cbd329e95c6b62651166/src/manifest.ts#L16

Skn0tt: I guess so, but that'd still require changes to all three services. But that field is probably better-suited. I'll make that update!

Skn0tt: done in https://github.com/netlify/zip-it-and-ship-it/pull/1559/commits/80a3c66ba1ddb49d6867b0de7debdc1eddab4ddb

src/runtimes/go/index.ts
```diff
@@ -139,13 +140,18 @@ const zipFunction: ZipFunction = async function ({
...
     }
 
     await zipBinary({ ...zipOptions, srcPath: binary.path, stat: binary.stat })
 
-    return { config, path: zipPath, entryFilename: zipOptions.filename }
+    return {
+      config,
+      path: zipPath,
+      entryFilename: zipOptions.filename,
+      runtimeVersion: featureFlags.zisi_golang_use_al2 ? 'provided.al2' : undefined,
```
eduardoboucas, line 153: Just for consistency with what we do in Node.js, it might be nice to always return the runtime version, even if it's the old one.

Not a big deal, though.

```suggestion
      runtimeVersion: featureFlags.zisi_golang_use_al2 ? 'provided.al2' : 'go1.x',
```

eduardoboucas (approved):

src/runtimes/go/index.ts
Skn0tt: Thought about that as well, but I didn't want to introduce changes when the flag is off. `undefined` should do that.

PR #1580, opened 2023-10-02
fix: infer TS module format from `package.json`

eduardoboucas:
#### Summary

Currently we're only looking at `tsconfig.json` to infer the module type. We need to also look at `package.json` as a fallback.

### DIFF
--- a/src/runtimes/node/bundlers/nft/index.ts
+++ b/src/runtimes/node/bundlers/nft/index.ts
@@ -97,7 +97,7 @@ const getTSModuleFormat = async (mainFile: string, repositoryRoot?: string): Pro
   }
 
   // At this point, we need to infer the module type from the `type` field in
-  // the closes `package.json`.
+  // the closest `package.json`.
   try {
     const packageJSON = await getClosestPackageJson(dirname(mainFile), repositoryRoot)
 
--- a/src/runtimes/node/bundlers/nft/index.ts
+++ b/src/runtimes/node/bundlers/nft/index.ts
@@ -10,9 +10,10 @@ import { cachedReadFile, getPathWithExtension } from '../../../../utils/fs.js'
 import { minimatch } from '../../../../utils/matching.js'
 import { getBasePath } from '../../utils/base_path.js'
 import { filterExcludedPaths, getPathsOfIncludedFiles } from '../../utils/included_files.js'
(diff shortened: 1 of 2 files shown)


src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -85,6 +86,31 @@ const getIgnoreFunction = (config: FunctionConfig) => {
   }
 }
 
+// Returns the module format that should be used when transpiling a TypeScript
+// file.
```
lukasholzer, line 90: NIT: comments for functions/methods or variables should always be JSDoc as the IntelliSense can pick them up. A regular code block comment should be just used inside functions as it is not part of the AST

```suggestion
/** 
 * Returns the module format that should be used when transpiling a TypeScript
 * file.
 */
```

src/runtimes/node/utils/tsconfig.ts
```diff
@@ -6,22 +6,27 @@ import { MODULE_FORMAT } from './module_format.js'
 
 const esmModuleValues = new Set(['es6', 'es2015', 'es2020', 'es2022', 'esnext', 'node16', 'nodenext'])
 
-// Returns the module format that should be used for a TypeScript file at a
-// given path, by reading the associated `tsconfig.json` file if it exists.
+// Looks for a `tsconfig.json` file on a given path and, if one exists, returns
+// the module format inferred from the `module` property. If no file is found
+// or if no `module` property is defined, the function returns `undefined`.
```
lukasholzer, line 11:
```suggestion
/**
 * Looks for a `tsconfig.json` file on a given path and, if one exists, returns
 * the module format inferred from the `module` property. If no file is found
 * or if no `module` property is defined, the function returns `undefined`.
 */
```

Same here

lukasholzer (approved):
Looks good to me, that's what I meant in Slack

src/runtimes/node/bundlers/nft/index.ts
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/1580/commits/99f27bafdce87d3667f82c7619a9f1a7f92f84b1.

lukasholzer (approved):

PR #1584, opened 2023-10-03
feat: bundle TypeScript files in V2 API

eduardoboucas:
#### Summary

When creating functions with the V2 API, we want to rewrite import paths so that they resolve to the right `.js` file, regardless of whether a `.js` extension, `.ts` extension, or no extension was used.


jenae-janzen:
Overall I think it looks good, but the tests are failing 🤔

tests/helpers/main.ts
```diff
@@ -91,13 +91,19 @@ export const zipFixture = async function (
...
 ): Promise<ZipReturn> {
   const srcFolders = Array.isArray(fixture)
     ? fixture.map((srcFolder) => `${fixtureDir}/${srcFolder}`)
     : `${fixtureDir}/${fixture}`
 
-  const files = await zipFunctions(srcFolders, tmpDir, opts)
+  let basePath
+
+  if (!opts.basePath && typeof fixture === 'string') {
+    basePath = resolve(fixtureDir, fixture)
```
eduardoboucas, line 103: All tests should have a `basePath` property, otherwise we'll look for modules and `tsconfig.json` files that belong to the parent zip-it-and-ship-it module, not the fixture.

The proper way would be to add it to each test individually, but I'm adding this as a fallback temporarily and will follow up in a separate PR.

This is also why you'll see some changes around Node modules in the fixtures. We were relying on modules from zip-it-and-ship-it in some tests, which we shouldn't.

eduardoboucas:
> Overall I think it looks good, but the tests are failing 🤔

@jenae-janzen fixed!

jenae-janzen (dismissed):

src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -113,6 +113,37 @@ const getTSModuleFormat = async (mainFile: string, repositoryRoot?: string): Pro
...
 
+const getTypeScriptTransformer = async (runtimeAPIVersion: number, mainFile: string, repositoryRoot?: string) => {
+  const isTypeScript = tsExtensions.has(extname(mainFile))
+
+  if (!isTypeScript) {
+    return
+  }
+
+  const aliases = new Map<string, string>()
+  const rewrites = new Map<string, string>()
```
lukasholzer, line 124: why are we returning two empty maps in this case of v2?

eduardoboucas: So that we can mutate them here: https://github.com/netlify/zip-it-and-ship-it/blob/fd3755c3c9b9d6cda44a05e804661fc806a8ed33/src/runtimes/node/bundlers/nft/index.ts#L195.

src/runtimes/node/utils/plugin_rewrite_imports.ts
```diff
@@ -0,0 +1,44 @@
+import { extname } from 'path'
+
+import { Plugin } from 'esbuild'
+
+const rewriteTSExtensionPlugin: Plugin = {
+  name: 'rewrite-imports',
+  setup(build) {
+    // Targeting relative imports only, since Node modules will be marked as
+    // external automatically because of `packages: "external"`.
+    build.onResolve({ filter: /^\./ }, (args) => {
```
lukasholzer, line 10: whats about typescript path aliases? Can you maybe write a test for that?

tests/fixtures-esm/v2-api-ts-aliases/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "@lib/*": ["super-long-and-awkward-name/*"]
```
lukasholzer, line 5: thanks for adding this test!

tests/fixtures-esm/v2-api-ts-aliases/function.ts
```diff
@@ -0,0 +1,10 @@
+import cjs from 'cjs-module'
+import esm from 'esm-module'
+
+import { name as helper1 } from '@lib/helper1.ts'
```
lukasholzer, line 4: noice!

src/runtimes/node/bundlers/nft/transpile.ts
```diff
@@ -23,8 +24,44 @@ export const transpile = async ({ config, format, name, path }: TranspileOptions
...
+}
+
+export const transpileTS = async ({ config, format, name, path }: TranspileTSOptions) => {
+  // The version of ECMAScript to use as the build target. This will determine
+  // whether certain features are transpiled down or left untransformed.
+  const nodeTarget = getBundlerTarget(config.nodeVersion)
+
+  try {
+    const transpiled = await build({
+      bundle: true,
```
lukasholzer, line 60: does this mean we are now bundling non v2 as well? or should we gate the bundling to only the v2 interface to not break old and existing sites without opting into the new interface?

lukasholzer (approved):
Just a question not going to block this :)

src/runtimes/node/bundlers/nft/transpile.ts
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/1584/commits/215595ac0e1337ed8e585fdf5e61c60a851d74a0.

lukasholzer (approved):
LGTM!

PR #1585, opened 2023-10-04
feat: add `tsConfig` property to `listFunctions` output

eduardoboucas:
#### Summary

Adds two additional properties to the return value of the `listFunctions` method:

- `inputModuleFormat`: Contains `cjs` or `esm` based on the syntax we detect during the static analysis stage. This is the module format that was used to author the files, which may differ from the module format used in the output file (e.g. a TypeScript file with ESM syntax may emit a CJS file)

- `tsConfig`: If the function entry point is a TypeScript file using the V2 API, we'll output a `tsConfig` property containing the path to the `tsconfig.json` that was loaded, as well as the full resolved config (including resolved `extends`)


src/runtimes/node/in_source_config/index.ts
```diff
@@ -8,22 +8,27 @@ import { getRoutesFromPath, Route } from '../../../utils/routes.js'
...
 
 export type ISCValues = {
-  invocationMode?: InvocationMode
   routes?: Route[]
-  runtimeAPIVersion?: number
   schedule?: string
   methods?: string[]
 }
 
+export interface StaticAnalysisResult extends ISCValues {
```
eduardoboucas, line 26: We were using the term "ISC" to contain properties that are not part of the in-source configuration spec, like `invocationMode` or `runtimeAPIVersion`. They were grouped together because we extract both those properties _and_ the ISC properties using the same static analysis step, but functionally they are different things.

So now we have a `ISCValues` type that holds just the in-source configuration properties, and a `StaticAnalysisResult` that holds those plus some additional metadata properties we'll extract using the static analysis mechanism.

A big part of this diff is just renaming ISC-related terms to a more generic "static analysis result".

src/utils/cache.ts
```diff
@@ -13,18 +13,25 @@ interface NFTCache {
...
   lstatCache: LstatCache
+
   // Cache fs.readdir() calls
   readdirCache: ReaddirCache
+
   // NFT cache, which should not be used in zisi and only supplied to NFT
   // this cache shares the file cache with zisi
   nftCache: NFTCache
 
+  // Used by `get-tsconfig` for caching the retrieval of `tsconfig.json` files
```
lukasholzer, line 27: NIT: class properties should always be annotated with JSDoc to benefit from this documentation in the IntelliSense when you use it with `this.tsConfigCache`

src/runtimes/node/parser/exports.ts
```diff
@@ -8,21 +8,37 @@ import type {
...
+/**
+ * Traverses a list of nodes and returns:
+ *
+ * 1. Named `config` object export
+ * 2. Default function export
+ * 3. Named `handler` function exports
+ * 4. The module format syntax used in the file: if any `import` or `export`
+ *    declarations are found, this is ESM; if not, this is CJS
+ */
+export const traverseNodes = (nodes: Statement[], getAllBindings: BindingMethod) => {
```
lukasholzer, line 25: 😍

src/runtimes/node/parser/exports.ts
```diff
@@ -8,21 +8,37 @@ import type {
...
   let configExport: Record<string, unknown> = {}
   let defaultExport: ExportDefaultDeclaration | undefined
+  let inputModuleFormat: ModuleFormat = MODULE_FORMAT.COMMONJS
 
   nodes.forEach((node) => {
+    if (
+      node.type === 'ImportDeclaration' ||
+      node.type === 'ExportNamedDeclaration' ||
+      node.type === 'ExportDefaultDeclaration' ||
+      node.type === 'ExportAllDeclaration'
```
lukasholzer, line 37: oh, this looks a little bit brittle to me is this the AST from esbuild? – according to the naming I think it is an acorn AST. Are there enums for the values?

But can we define the inputModule format to ESM based on that? Whats about typescript or is this already the transpiled code we are parsing here?

lukasholzer (approved):
Looks good to me – hard to understand the full context often as the codebase lacks a lot of documentation sadly on what properties are or what they should do

src/runtimes/node/parser/exports.ts
eduardoboucas: > oh, this looks a little bit brittle to me is this the AST from esbuild? – according to the naming I think it is an acorn AST. Are there enums for the values?

esbuild doesn't expose an AST. This is from Babel, as you can see here: https://github.com/netlify/zip-it-and-ship-it/blob/d98dd20825b80d78c6ff0e5c4f2b51c7117d6af5/src/runtimes/node/parser/index.ts#L4.

Can you elaborate on why you think this is brittle? The nodes are fully typed, so if there's a mismatch in this or in a future version of the AST lexicon, we'll know?

> But can we define the inputModule format to ESM based on that? Whats about typescript or is this already the transpiled code we are parsing here?

As I explained in the PR body, we're not trying to infer the module type that we'll emit. We're simply trying to see what kind of syntax was used when authoring the function. If we see an `import` or an `export`, we know it's ESM syntax.

For TypeScript functions this won't matter, but we can use this information to detect when people are writing a `.js` function with ESM without a `type: "module"`. That's the main purpose of surfacing this information.

eduardoboucas:
> hard to understand the full context often as the codebase lacks a lot of documentation sadly on what properties are or what they should do

What parts of the codebase do you feel are lacking documentation?

eduardoboucas:
After discussing with @lukasholzer, we're not sure this is still needed in its current form. I still think some of the cleaning up we're doing here would be great to land, so I'm moving to draft until I have time to pick this up again.

src/runtimes/node/in_source_config/index.ts
```diff
@@ -84,10 +73,32 @@ const normalizeMethods = (input: unknown, name: string): string[] | undefined =>
   })
 }
 
-export const findISCDeclarations = (
+/**
+ * Loads a file at a given path, parses it into an AST, and returns a series of
+ * data points, such as in-source configuration properties and other metadata.
+ */
+export const parseFile = async (
```
eduardoboucas, line 80: This was renamed from `findISCDeclarationsInPath`. No functionality has changed.

src/runtimes/node/in_source_config/index.ts
```diff
@@ -101,27 +112,28 @@ export const findISCDeclarations = (
   let scheduleFound = false
 
   const getAllBindings = createBindingsMethod(ast.body)
-  const { configExport, defaultExport, handlerExports } = getExports(ast.body, getAllBindings)
-  const isV2API = handlerExports.length === 0 && defaultExport !== undefined
+  const { configExport, handlerExports, hasDefaultExport, inputModuleFormat } = traverseNodes(ast.body, getAllBindings)
```
eduardoboucas, line 115: We were not using the contents of `defaultExport` for anything, just checking whether it exists or not. So I've changed it to a boolean that is now `true` if ESM or CJS default exports are found.

src/runtimes/node/parser/exports.ts
```diff
@@ -9,60 +9,87 @@ import type {
...
     }
   })
 
-  return { configExport, defaultExport, handlerExports }
+  return { configExport, handlerExports, hasDefaultExport, inputModuleFormat }
 }
 
 // Finds the main handler export in a CJS AST.
-const getMainExportFromCJS = (node: Statement) => {
+const getCJSExports = (node: Statement, name: string) => {
```
eduardoboucas, line 89: This is now generic so it can look for exports with different names.

src/runtimes/node/parser/exports.ts
```diff
@@ -74,16 +101,20 @@ const getMainExportFromCJS = (node: Statement) => {
...
 }
 
-// Finds the main handler export in an ESM AST.
-const getMainExportFromESM = (node: Statement, getAllBindings: BindingMethod) => {
+/**
+ * Finds a named ESM export with a given name. It's capable of finding exports
+ * with a variable declaration (`export const foo = "bar"`), but also resolve
+ * bindings and find things like `const baz = "1"; export { baz as foo }`.
+ */
+const getNamedESMExport = (node: Statement, name: string, getAllBindings: BindingMethod) => {
```
eduardoboucas, line 109: Renamed from `getMainExportFromESM`. This is now generic so it can look for exports with different names.

src/runtimes/node/parser/exports.ts
```diff
@@ -93,35 +124,41 @@ const getMainExportFromESM = (node: Statement, getAllBindings: BindingMethod) =>
...
-// either with Identifier `export { handler }`
-// or with StringLiteral `export { x as "handler" }`
-const isHandlerExport = (node: ExportNamedDeclaration['specifiers'][number]): node is ExportSpecifier => {
+/**
+ * Check if the node is an `ExportSpecifier` that has a named export with
+ * the given name, either as:
+ * - `export { handler }`, or
+ * - `export { x as "handler" }`
+ */
+const isNamedExport = (node: ExportNamedDeclaration['specifiers'][number], name: string): node is ExportSpecifier => {
```
eduardoboucas, line 141: Renamed from `isHandlerExport`. This is now generic so it can look for exports with different names.

minivan (approved):

src/runtimes/node/parser/exports.ts
```diff
@@ -187,13 +224,20 @@ const parsePrimitive = (exp: Expression | PatternLike): PrimitiveResult => {
...
+ * For example, the following would resolve correctly to the handler function:
+ *
+ * `let handler; handler = () => {}; export { handler }`
+ */
+const getExportsFromBindings = (
+  specifiers: ExportNamedDeclaration['specifiers'],
+  name: string,
+  getAllBindings: BindingMethod,
+) => {
+  const specifier = specifiers.find((node) => isNamedExport(node, name))
```
lukasholzer, line 238: here you get only the named exports what about doing an object literal export over a default?

```javascript
const handler = () => {}

export default {
  handler
}
```

lukasholzer: same would apply to the cjs where it's:

```javascript
const handler = () => {}

exports.default {
  handler
}
```

lukasholzer (commented):
I would add a test for the default export with a object literal

PR #1586, opened 2023-10-04
feat: support multiple paths in V2 functions

eduardoboucas:
#### Summary

Allow an array of strings in the `path` in-source configuration property of V2 functions.

The types [already allow this](https://github.com/netlify/functions/blob/3270f5f7823aef8f27e62ab39322514481b7b171/src/function/v2.ts#L10), as well as the rest of the infrastructure.

### DIFF
--- a/src/utils/routes.ts
+++ b/src/utils/routes.ts
@@ -19,10 +19,6 @@ const isPathLiteral = (path: string) => {
 }
 
 const getRoute = (path: unknown, functionName: string, methods: string[]): Route | undefined => {
-  if (!path) {
-    return
-  }
-
   if (typeof path !== 'string') {
     throw new FunctionBundlingUserError(`'path' property must be a string, found '${typeof path}'`, {
       functionName,
--- a/src/utils/routes.ts
+++ b/src/utils/routes.ts
@@ -1,6 +1,7 @@
 import { RUNTIME } from '../runtimes/runtime.js'
 
 import { FunctionBundlingUserError } from './error.js'
+import { nonNullable } from './non_nullable.js'
 import { ExtendedURLPattern } from './urlpattern.js'
 
 export type Route = { pattern: string; methods: string[] } & ({ literal: string } | { expression: string })
@@ -17,9 +18,9 @@ const isPathLiteral = (path: string) => {
   return parts.every((part) => !isExpression(part))
 }
 
-export const getRoutesFromPath = (path: unknown, functionName: string, methods: string[]): Route[] => {
+const getRoute = (path: unknown, functionName: string, methods: string[]): Route | undefined => {
   if (!path) {
-    return []
+    return
   }
 
   if (typeof path !== 'string') {
@@ -37,7 +38,7 @@ export const getRoutesFromPath = (path: unknown, functionName: string, methods: 
   }
 
   if (isPathLiteral(path)) {
-    return [{ pattern: path, literal: path, methods }]
+    return { pattern: path, literal: path, methods }
   }
 
   try {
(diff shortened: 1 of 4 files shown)


tests/unit/runtimes/node/in_source_config.test.ts
```diff
@@ -262,7 +262,7 @@ describe('V2 API', () => {
     })
 
     test('With an invalid pattern', () => {
-      expect.assertions(4)
+      expect.assertions(8)
```
lukasholzer, line 265: oof 🫣

tests/unit/runtimes/node/in_source_config.test.ts
```diff
@@ -317,5 +336,26 @@ describe('V2 API', () => {
...
       ])
     })
+
+    test('Using multiple paths', () => {
+      const source = `export default async () => {
+        return new Response("Hello!")
+      }
+  
+      export const config = {
+        path: ["/store/:category/products/:product-id", "/super-awesome-campaign"]
```
lukasholzer, line 346: can we make the second one a named group as well?

+ Can we add a test for what if both routes are the same?
+ What if I add falsy values like `null` or `undefined` to the array? Do we throw type errors if numbers or whatever is provided? (You have a non nullable function but no test cases for it)

src/utils/routes.ts
```diff
@@ -52,11 +49,22 @@ export const getRoutesFromPath = (path: unknown, functionName: string, methods:
...
+
+export const getRoutes = (input: unknown, functionName: string, methods: string[]): Route[] => {
+  if (!input) {
+    return []
+  }
+
+  const paths = Array.isArray(input) ? input : [input]
+  const routes = paths.map((path) => getRoute(path, functionName, methods ?? [])).filter(nonNullable)
+
+  return routes
```
lukasholzer, line 69: I would do `return [...new Set(routes)]` to remove duplicates

src/utils/routes.ts
```diff
@@ -52,11 +49,22 @@ export const getRoutesFromPath = (path: unknown, functionName: string, methods:
...
+    return { pattern: path, expression: normalizedRegex, methods }
   } catch {
     throw new FunctionBundlingUserError(`'${path}' is not a valid path according to the URLPattern specification`, {
       functionName,
       runtime: RUNTIME.JAVASCRIPT,
     })
   }
 }
+
+export const getRoutes = (input: unknown, functionName: string, methods: string[]): Route[] => {
```
lukasholzer, line 61: NIT: would love a small jsdoc that explains the parameters and the method. So that a tipsy on call engineer at 2am in the morning can understand it as well :D

src/utils/routes.ts
```diff
@@ -17,11 +18,7 @@ const isPathLiteral = (path: string) => {
   return parts.every((part) => !isExpression(part))
 }
 
-export const getRoutesFromPath = (path: unknown, functionName: string, methods: string[]): Route[] => {
-  if (!path) {
-    return []
-  }
-
+const getRoute = (path: unknown, functionName: string, methods: string[]): Route | undefined => {
```
lukasholzer, line 21: NIT: small jsdoc would be nice

lukasholzer (approved):
Some small comments and request for tests but nothing blocking ;)

src/utils/routes.ts
eduardoboucas: Done in https://github.com/netlify/zip-it-and-ship-it/pull/1586/commits/9e2cac92e2a50eb82f1ae5f112651b3933522d81.

eduardoboucas: We can't do it on `routes` because it's an array of objects. I've done it in `paths`.

tests/unit/runtimes/node/in_source_config.test.ts
eduardoboucas: I've added more tests to capture this.

tests/unit/runtimes/node/in_source_config.test.ts
eduardoboucas: I didn't think it was worth splitting into another test.

src/utils/routes.ts
eduardoboucas: Done.

src/utils/routes.ts
eduardoboucas: Done.

eduardoboucas:
@lukasholzer I think I addressed all your feedback. I ended up having to make some changes in the AST parsing, because when processing arrays we were only looking for strings, so any other primitive values were ignored (so we couldn't throw a type error if people tried to use them).

This has now been addressed.

tests/unit/runtimes/node/in_source_config.test.ts
eduardoboucas: I've split this into two tests now.

lukasholzer (approved):
Thanks for the added test cases! 🚀

PR #1587, opened 2023-10-06
feat: infer TypeScript format from `package.json` only

eduardoboucas:
#### Summary

When compiling TypeScript functions, infer the module type only from `package.json` and not from `tsconfig.json`.

Also ensures that the entry files we generate for V2 functions have an unambiguous extension depending on their module format (i.e. either `.cjs` or `.mjs`).


minivan (dismissed):

src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -90,12 +89,22 @@ const getIgnoreFunction = (config: FunctionConfig) => {
...
  * file.
  */
-const getTSModuleFormat = async (mainFile: string, repositoryRoot?: string): Promise<ModuleFormat> => {
-  const fromTsConfig = getModuleFormatFromTsConfig(mainFile, repositoryRoot)
+const getTSModuleFormat = async (
+  mainFile: string,
+  runtimeAPIVersion: number,
+  repositoryRoot?: string,
+): Promise<ModuleFormat> => {
+  // TODO: This check should go away. We should always respect the format from
```
eduardoboucas, line 97: I want to make sure that we're not touching V1 functions with this release.

lukasholzer (approved):
I love how much code got removed! 🚢

PR #1589, opened 2023-10-07
feat: improvements to V2 functions

eduardoboucas:
#### Summary

This PR has 3 small changes made separately in the following commits:

- https://github.com/netlify/zip-it-and-ship-it/commit/c52807a4a04e329313bbd83f3bb9971fc29d3274 + https://github.com/netlify/zip-it-and-ship-it/pull/1589/commits/802e351c15756436af42a4fc67a4d62f797a68c1: A new `outputModuleFormat` is added to the function objects returned by zip-it-and-ship-it, containing the module format that the function must run as (`cjs` or `esm`)

- https://github.com/netlify/zip-it-and-ship-it/commit/d1b747c3678245a86e0f4855254faef32fb31e24: The `inputs` array in the returned function objects contains the list of paths that acted as inputs to the output function. This is used by the CLI to know which files it needs to watch and associate with each function. Since https://github.com/netlify/zip-it-and-ship-it/pull/1584, we're bundling some of the files together, but the originating files still need to be part of `inputs`.

- https://github.com/netlify/zip-it-and-ship-it/commit/c0be11ca9c7dbf6c3fce5032d10b49926831469d: We're forcing the bundler to NFT in V2 functions, since that's the only bundler that is optimised for this flow, and we don't want people manually configuring bundlers anymore. We still allow the `none` bundler, in case users (or integrations) want to take over the bundling process.

- https://github.com/netlify/zip-it-and-ship-it/pull/1589/commits/4d6cae358ebfac14ed726b44c05fcc73a4467f11: Adds a shim that makes `require`, `__dirname`, and `__filename` available when bundling local imports to ESM.

### DIFF
--- a/src/runtimes/node/bundlers/nft/transpile.ts
+++ b/src/runtimes/node/bundlers/nft/transpile.ts
@@ -5,6 +5,7 @@ import { build, BuildOptions } from 'esbuild'
 import type { FunctionConfig } from '../../../../config.js'
 import { FunctionBundlingUserError } from '../../../../utils/error.js'
 import { RUNTIME } from '../../../runtime.js'
+import { CJS_SHIM } from '../../utils/esm_cjs_compat.js'
 import { ModuleFormat, MODULE_FORMAT } from '../../utils/module_format.js'
 import { getBundlerTarget } from '../esbuild/bundler_target.js'
 import { NODE_BUNDLER } from '../types.js'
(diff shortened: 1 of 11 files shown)


minivan (approved):

src/runtimes/node/utils/esm_cjs_compat.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * https://github.com/evanw/esbuild/pull/2067
+ */
+export const CJS_SHIM = `
+import {createRequire as ___nfyCreateRequire} from "module";
+import {fileURLToPath as ___nfyFileURLToPath} from "url";
+import {dirname as ___nfyPathDirname} from "path";
+let __filename=___nfyFileURLToPath(import.meta.url);
+let __dirname=___nfyPathDirname(___nfyFileURLToPath(import.meta.url));
+let require=___nfyCreateRequire(import.meta.url);
```
lukasholzer, line 13: Nice so we can even shim `require.resolve` 🥳

lukasholzer (approved):
LGTM!

src/runtimes/node/bundlers/nft/transpile.ts
```diff
@@ -52,22 +55,34 @@ export const transpileTS = async ({ bundle = false, config, format, name, path }
...
+  const bundleOptions: BuildOptions = {
+    bundle: false,
+  }
 
-  const bundleOptions: BuildOptions = bundle ? { bundle: true, packages: 'external' } : { bundle: false }
+  if (bundle) {
+    bundleOptions.bundle = true
+    bundleOptions.packages = 'external'
+
+    if (format === MODULE_FORMAT.ESM) {
```
lukasholzer, line 66: so we emit this shim in any case even if the srcs are esm?

eduardoboucas: Yeah, I don't think we have an easy way to determine whether we have a CJS source or not. I mean, we could do it, but not sure if it's worth the hassle?

PR #1597, opened 2023-10-09
feat: always transform V2 functions

eduardoboucas:
#### Summary

I realised that we were only transforming V2 functions (i.e. running them through esbuild) when the extension was a TypeScript extension. But we should also do it for JavaScript functions, since bundling local imports is essential to ensure imports between ESM and CJS work as expected.

I ended up renaming the "TS transformer" to a generic transformer, and I moved it into its own file.

### DIFF
--- a/src/runtimes/node/bundlers/nft/index.ts
+++ b/src/runtimes/node/bundlers/nft/index.ts
@@ -1,4 +1,4 @@
-import { basename, dirname, join, normalize, resolve, extname } from 'path'
+import { basename, dirname, join, normalize, resolve } from 'path'
 
 import { nodeFileTrace } from '@vercel/nft'
 import resolveDependency from '@vercel/nft/out/resolve-dependency.js'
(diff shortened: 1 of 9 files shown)


src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -90,90 +89,6 @@ const getIgnoreFunction = (config: FunctionConfig) => {
   }
 }
 
-/**
- * Returns the module format that should be used when transpiling a TypeScript
- * file.
- */
-const getTSModuleFormat = async (
```
eduardoboucas, line 97: This is now called `getModuleFormat` in `transformer.ts`.

src/runtimes/node/bundlers/nft/index.ts
```diff
@@ -90,90 +89,6 @@ const getIgnoreFunction = (config: FunctionConfig) => {
...
-type TypeScriptTransformer = {
-  aliases: Map<string, string>
-  bundle?: boolean
-  bundledPaths?: string[]
-  format: ModuleFormat
-  newMainFile?: string
-  rewrites: Map<string, string>
-}
-
-const getTypeScriptTransformer = async (
```
eduardoboucas, line 139: This is now called `getTransformer` in `transformer.ts`.

src/runtimes/node/bundlers/nft/transpile.ts
```diff
@@ -42,52 +39,3 @@ export const transpileESMToCJS = async ({ config, name, path }: TranspileESMToCJ
...
-
-interface TranspileTSOptions {
-  bundle?: boolean
-  config: FunctionConfig
-  format?: ModuleFormat
-  name: string
-  path: string
-}
-
-export const transpileTS = async ({ bundle = false, config, format, name, path }: TranspileTSOptions) => {
```
eduardoboucas, line 54: This is now called `transform` in `transformer.ts`.

Skn0tt (approved):
I found the diff hard to review - but I think it looks good. And since you didn't touch a lot of tests, i'm trusting this to be fine.

PR #1614, opened 2023-10-13
feat: include side files with NFT

eduardoboucas:
#### Summary

In the legacy `zisi` bundler , functions defined in a sub-directory had any sibling files recursively added to the bundle., regardless of whether they were statically required by the function or not.

As we try to remove this bundler entirely without breaking existing workflows, this PR ports that behaviour over to NFT, only when the default bundler is used (so that we don't affect functions explicitly opting in to `nft`).


tests/fixtures/directory-side-files/function-v1/Desktop.ini
Skn0tt, line 1: This looks like a remnant from Windows testing

src/runtimes/node/bundlers/nft/side_files.ts
```diff
@@ -0,0 +1,25 @@
+import type { Stats } from 'fs'
+import { basename } from 'path'
+
+import { isJunk } from 'junk'
```
Skn0tt, line 4: TIL about `junk`!

tests/fixtures/directory-side-files/function-v1/Desktop.ini
Skn0tt: oh it's not, it's junk!

Skn0tt (approved):
Looks good!

eduardoboucas:
@Skn0tt I pushed https://github.com/netlify/zip-it-and-ship-it/pull/1614/commits/a8af2875f0f2005cdd678ad67e0bbd3ca02f301d with a change to the test, ensuring we run for all bundler variants and don't include the side files unless the default bundler is used.

Can you re-approve, please?

Skn0tt (approved):

PR #1620, opened 2023-10-17
fix: handle named export declaration with two specifiers for v2

ascorbic:

#### Summary
Handles the case where there is an export in the pattern `export { config, handler as default }`. Also adds support for cases where the default export is an identifier, such as where `const handler = createRequestHandler()` (e.g. Remix)

---

      passes our tests.


src/runtimes/node/parser/exports.ts
```diff
@@ -59,12 +64,6 @@ export const traverseNodes = (nodes: Statement[], getAllBindings: BindingMethod)
       return
     }
 
-    if (isESMDefaultExport(node)) {
-      hasDefaultExport = true
-
-      return
-    }
-
```
ascorbic, line 67: Moving later and not returning because the node might be both a config export _and_ a default export.

src/runtimes/node/parser/exports.ts
```diff
@@ -39,6 +39,11 @@ export const traverseNodes = (nodes: Statement[], getAllBindings: BindingMethod)
     }
 
     const esmHandlerExports = getNamedESMExport(node, 'handler', getAllBindings)
+    const esmConfigExports = getNamedESMExport(node, 'config', getAllBindings)
+
+    if (esmConfigExports.length !== 0 && esmConfigExports[0].type === 'object-expression') {
+      configExport = esmConfigExports[0].object
+    }
```
ascorbic, line 46: Handle `export { config } `

tests/unit/runtimes/node/in_source_config.test.ts
```diff
@@ -187,6 +187,55 @@ describe('V2 API', () => {
...
+      const source = `
+      const handler = async () => ({ statusCode: 200, body: "Hello" })
+      const config = { schedule: "@daily" }
+      export { handler as default, config };`
+
+      const isc = parseSource(source, options)
+      expect(isc).toEqual({ inputModuleFormat: 'esm', routes: [], schedule: '@daily', runtimeAPIVersion: 2 })
+    })
+
+    test('the Remix handler', () => {
```
ascorbic, line 211: It's weird, so add it as a test case

src/runtimes/node/parser/exports.ts
```diff
@@ -39,6 +39,11 @@ export const traverseNodes = (nodes: Statement[], getAllBindings: BindingMethod)
     }
 
     const esmHandlerExports = getNamedESMExport(node, 'handler', getAllBindings)
+    const esmConfigExports = getNamedESMExport(node, 'config', getAllBindings)
```
eduardoboucas, line 42: I'm trying to think of the implications of having both this and this: https://github.com/netlify/zip-it-and-ship-it/blob/c8f1b138495bbd4e102177f10523b8bdd6d71c3c/src/runtimes/node/parser/exports.ts#L79

I think it just means that if you do:

```js
export { config }
export const config = { ... }
```

... then the second one takes precedence?

tests/unit/runtimes/node/in_source_config.test.ts
eduardoboucas, line 211: nit: I think it's fine to have the Remix case here, but I'd rather have a more generic name that describes the characteristics of the file.

```suggestion
    test('ESM file with the config and a default as a single named export', () => {
```

eduardoboucas (commented):
LGTM! I have just one question that I'd love to clarify before shipping.

src/runtimes/node/parser/exports.ts
ascorbic: That would be an error. You can't have two exports with the same name.

tests/unit/runtimes/node/in_source_config.test.ts
ascorbic: We have a generic test too

src/runtimes/node/parser/exports.ts
eduardoboucas: Right, we don't detect that here, so as far as the static analysis goes, the second config block would take precedence?

tests/unit/runtimes/node/in_source_config.test.ts
ascorbic: I've given it a name that I think covers the specific difference from the genertic test

src/runtimes/node/parser/exports.ts
ascorbic: It would have a runtime error though, so I'm not sure it matters what happens?

eduardoboucas (approved):
🚀

PR #1624, opened 2023-10-19
fix: remove golang_al2 flag

Skn0tt:
Part of https://github.com/netlify/pod-dev-foundations/issues/581. We've fully rolled out the flag successfully, so we can remove it now.


src/runtimes/go/index.ts
```diff
@@ -138,7 +137,7 @@ const zipFunction: ZipFunction = async function ({
     config,
     displayName: config?.name,
     generator: config?.generator || getInternalValue(isInternal),
-    runtimeVersion: featureFlags.zisi_golang_use_al2 ? 'provided.al2' : undefined,
+    runtimeVersion: 'provided.al2',
```
Skn0tt, line 140: We'll be able to remove this as well, once `provided.al2` is the default in our upload code. That'll have to wait until November 1st, though: https://answers.netlify.com/t/under-the-hood-changes-to-go-function-uploads-please-update-your-netlify-cli-version/104726

eduardoboucas (approved):

PR #1640, opened 2023-11-01
fix: handle symlinks when creating directory

eduardoboucas:
#### Summary

When creating a bundle into a directory, we need to respect symlinks and not just copy them as files.

Fixes https://github.com/netlify/cli/issues/6106

### DIFF
--- a/src/runtimes/node/utils/zip.ts
+++ b/src/runtimes/node/utils/zip.ts
@@ -1,6 +1,6 @@
 import { Buffer } from 'buffer'
 import { Stats } from 'fs'
-import { mkdir, rm, writeFile } from 'fs/promises'
+import { mkdir, readlink as readLink, rm, symlink, writeFile } from 'fs/promises'
 import os from 'os'
 import { basename, extname, join } from 'path'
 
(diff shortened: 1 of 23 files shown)


src/runtimes/node/utils/zip.ts
```diff
@@ -132,11 +135,29 @@ const createDirectory = async function ({
         return mkdirAndWriteFile(absoluteDestPath, rewrites.get(srcFile) as string)
       }
 
+      const stat = await cachedLstat(cache.lstatCache, srcFile)
+
+      // If the path is a symlink, find the link target and add the link to a
+      // `symlinks` map, which we'll later use to create the symlinks in the
+      // target directory. We can't do that right now because the target path
+      // may not have been copied over yet.
+      if (stat.isSymbolicLink()) {
```
eduardoboucas, line 144: We should probably do the same in `createZipArchive`, but starting with `createDirectory` for now.

@ascorbic what happened when you tried to deploy the function to production?

eduardoboucas: According to the test, it works: https://github.com/netlify/zip-it-and-ship-it/pull/1640/commits/07061e053162c76d76bdfd7d41317d47dc0f5e85.

ascorbic: Yeah, it worked fine in builds

Skn0tt (approved):

PR #1641, opened 2023-11-01
chore: add test for V2 pattern

eduardoboucas:
#### Summary

Adds a test case asserting that a specific pattern is detected as a V2 function.

Closes https://github.com/netlify/pod-dev-foundations/issues/606.


tests/unit/runtimes/node/in_source_config.test.ts
```diff
@@ -285,6 +286,17 @@ describe('V2 API', () => {
...
+
+    test('ESM file with a default export consisting of a function call', () => {
+      const source = `import * as build from "@remix-run/dev/server-build";
+      import { createRequestHandler } from "@netlify/remix-edge-adapter";
+      
+      export default createRequestHandler({ build });`
+
+      const isc = parseSource(source, options)
+
+      expect(isc).toEqual({ inputModuleFormat: 'esm', routes: [], runtimeAPIVersion: 2 })
```
eduardoboucas, line 298: @ascorbic I've used the source you shared in your example and it seems to be detected as a v2 function. Are you sure the detection is what's failing in your case?

minivan (approved):

PR #1643, opened 2023-11-02
fix: populate Netlify global before customer code

Skn0tt:
Fixes https://linear.app/netlify/issue/COM-101/unable-to-use-netlifyenv-in-functions-v2.

### DIFF
--- a/tests/v2api.test.ts
+++ b/tests/v2api.test.ts
@@ -468,7 +468,7 @@ describe.runIf(semver.gte(nodeVersion, '18.13.0'))('V2 functions API', () => {
     }
   })
 
-  testMany(
+  testMany.only(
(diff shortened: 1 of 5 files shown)


tests/v2api.test.ts
```diff
@@ -468,7 +468,7 @@ describe.runIf(semver.gte(nodeVersion, '18.13.0'))('V2 functions API', () => {
     }
   })
 
-  testMany(
+  testMany.only(
```
eduardoboucas, line 471:
```suggestion
  testMany(
```

Skn0tt:
Hmm, GitHub Actions seems to be having troubles 🤔 Screenshot 2023-11-02 at 17 46 47

PR #1657, opened 2023-11-17
fix: detect alt version of remix handler

Skn0tt:
While working on our Remix template, I found that the handler function that was generated for me isn't detected as v2, even though it should. This PR adds a test case for it.

It also adds the most straight-forward fix I found. It has the downside of creating a false-positive (will comment below) about that, but I think that's OK. Ideally though, we find a solution that doesn't require changes to the existing tests.


lukasholzer (dismissed):
LGTM!

eduardoboucas (dismissed):

eduardoboucas (approved):

PR #1658, opened 2023-11-19
fix(exports parser): handle `as default` export

lilnasy:
#### Summary

- Unblocks https://github.com/withastro/adapters/pull/31
- The parser failed to detect the following as a default export.
image

- This resulted in `runtimeAPIVersion` failing to be correctly detected as v2 [here](https://github.com/netlify/zip-it-and-ship-it/blob/b8a39b643d2c6101a20683e1a6823e1da188bf5a/src/runtimes/node/index.ts#L65)
- And an incorrect version being picked by the builder [here](https://github.com/netlify/cli/blob/main/src/lib/functions/runtimes/js/builders/zisi.mts#L74)
- The CLI then goes on to deploy critically misbehaving functions, like in [this case](https://655979f55ad363609619075c--bespoke-gumdrop-44a21e.netlify.app/)
image

The implementation may be done better, I don't have the full context of the codebase. Feel free to improve upon it.

### DIFF
--- a/src/runtimes/node/parser/exports.ts
+++ b/src/runtimes/node/parser/exports.ts
@@ -45,6 +45,15 @@ export const traverseNodes = (nodes: Statement[], getAllBindings: BindingMethod)
       configExport = esmConfigExports[0].object
     }
 
+    if (node.type === 'ExportNamedDeclaration') {
+        for (const exportSpecifier of node.specifiers) {
+            if (exportSpecifier.exported.name === 'default') {
+                hasDefaultExport = true;
+                return;
+            }
+        }    
+    }
+
     if (esmHandlerExports.length !== 0) {
       if (esmHandlerExports.some(({ type }) => type === 'default')) {
         hasDefaultExport = true


Skn0tt:
Hey @lilnasy, thanks for opening this PR - much appreciated! Coincidentally, I opened https://github.com/netlify/zip-it-and-ship-it/pull/1657 on friday which might address the same thing. I'll compare & review them sometime today.

Skn0tt:
Yup, your fix is pretty much the same as mine from Friday :D I've added the code from your screenshot as a test case to https://github.com/netlify/zip-it-and-ship-it/pull/1657/commits/7dfba88d56515afc1dd29a44b50f56a4cf6a3c86. Going to close your PR, and merge mine - will ping here again once it's deployed.

lilnasy:
@Skn0tt Thanks for that! Do you guys have stats on how many users are on older versions of `netlify-cli`?
I would probably have to workaround the issue with regex. Is there a way to explicitly inform the cli of the runtime version?

Skn0tt:
> will ping here again once it's deployed.

The update went out to the CLI in https://github.com/netlify/cli/releases/tag/v17.5.3 and was deployed to our CI build system a couple minutes ago.

> Is there a way to explicitly inform the cli of the runtime version?

I'm not aware of a way, nor did I find something in the code :/ This might be something we could build, will look into it.

> Do you guys have stats on how many users are on older versions of `netlify-cli`?

Just ran the numbers for you. It seems like in the past 24hrs of `netlify deploy` executions, only about 1/3 of them were on recent versions (`17.5.x`), and others were lagging behind quite a bit. But it should be pretty straight-forward for most to update their CLI version.

If the regex is straight-forward for you, then that might be the best bet. If it isn't, maybe you could try to detect the version of the Netlify CLI that's being used to build, and warn/error on versions below `17.5.3`? Let me know if you want to go that way, happy to help in how to detect the CLI version.

lilnasy:
It would be relatively reliable and easy to string replace. I will do that.

> This might be something we could build, will look into it
Yes, it will help in a similar situation in the future.

Skn0tt:
> Yes, it will help in a similar situation in the future.

Opened an internal issue about it, let's see when we get to it 🤷

Skn0tt:
@lilnasy it's now possible to detect the version of the Netlify CLI that's running using the `NETLIFY_CLI_VERSION` environment variable: https://github.com/netlify/cli/pull/6226

PR #1667, opened 2023-11-22
fix: integration test needs to use run.js instead of run.mjs

Skn0tt:
With https://github.com/netlify/cli/commit/5eb2133a277191c68e6d2129f0ebe6d5d85a9402, the CLI changed the file format of `run.mjs` to `run.js`. Gotta move our test along!


eduardoboucas (approved):
:shipit:

PR #1676, opened 2023-12-01
fix: drop runtimeVersion override for go functions

Skn0tt:
Part of https://github.com/netlify/pod-dev-foundations/issues/581. `provided.al2` is the default runtime for Go functions now, so we don't need the override here anymore. Put the change behind a flag, so we can rollout gradually.


lukasholzer (commented):
🧹

lukasholzer (approved):

PR #1701, opened 2024-01-26
fix: preserve symlinks from included files in the resulting zip archive

lukasholzer:

#### Summary

Fixes https://linear.app/netlify/issue/COM-318/function-bundle-is-not-preserving-symlinks-therefore-duplicating-node

---

      passes our tests.

### DIFF
--- a/src/runtimes/node/utils/included_files.ts
+++ b/src/runtimes/node/utils/included_files.ts
@@ -53,12 +53,11 @@ export const getPathsOfIncludedFiles = async (
     cwd: basePath,
     dot: true,
     ignore: excludePatterns,
-    followSymbolicLinks: false,
+    onlyFiles: false,
     // get directories as well to get symlinked directories,
     // to filter the regular non symlinked directories out mark them with a slash at the end to filter them out.
-    onlyFiles: false,
     markDirectories: true,
-    throwErrorOnBrokenSymbolicLink: true,
+    followSymbolicLinks: false,
   })
 
   // now filter the non symlinked directories out that got marked with a trailing slash
--- a/src/runtimes/node/utils/included_files.ts
+++ b/src/runtimes/node/utils/included_files.ts
@@ -53,9 +53,14 @@ export const getPathsOfIncludedFiles = async (
     cwd: basePath,
     dot: true,
     ignore: excludePatterns,
-    followSymbolicLinks: true,
+    followSymbolicLinks: false,
+    // get directories as well to get symlinked directories,
+    // to filter the regular non symlinked directories out mark them with a slash at the end to filter them out.
+    onlyFiles: false,
+    markDirectories: true,
     throwErrorOnBrokenSymbolicLink: true,
   })
 
-  return { excludePatterns, paths: pathGroups.map(normalize) }
+  // now filter the non symlinked directories out that got marked with a trailing slash
+  return { excludePatterns, paths: pathGroups.filter((path) => !path.endsWith('/')).map(normalize) }
 }
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -1,11 +1,12 @@
-import { mkdir, readFile, chmod, symlink, writeFile, rm } from 'fs/promises'
+import { mkdir, readFile, rm, symlink, writeFile } from 'fs/promises'
 import { dirname, isAbsolute, join, resolve } from 'path'
-import { arch, platform, version as nodeVersion } from 'process'
+import { arch, version as nodeVersion, platform } from 'process'
 
-import AdmZip from 'adm-zip'
 import cpy from 'cpy'
+import decompress from 'decompress'
 import merge from 'deepmerge'
 import { execa, execaNode } from 'execa'
+import glob from 'fast-glob'
 import isCI from 'is-ci'
 import { pathExists } from 'path-exists'
 import semver from 'semver'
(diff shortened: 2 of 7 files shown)


tests/symlinked_included_files.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    },
+    featureFlags: {},
+    repositoryRoot: basePath,
+    systemLog: console.log,
+    debug: true,
+    internalSrcFolder: undefined,
+  })
+
+  const unzippedPath = join(tmpDir, 'extracted')
+  await decompress(result!.path, unzippedPath)
```
eduardoboucas, line 55: Does the [existing util for unzipping](https://github.com/netlify/zip-it-and-ship-it/blob/8a3845d6e1500eba8d07c1e29b1176d5c9ca7d85/tests/helpers/main.ts#L142) not work here due to symlinks? If that's the case, it would be good to migrate all tests to use this new library, otherwise we're left with two distinct ways of doing the same thing.

Not saying you need to do that as part of this PR, but at least let's have an issue for it.

tests/symlinked_included_files.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      await readDirWithTypes(join(dir, dirent.name), files, dirent.name)
+    } else {
+      files[join(parent, dirent.name)] = dirent.isSymbolicLink()
+    }
+  }
+
+  return files
+}
+
+test('if symlinked directories from the included files are preserved', async () => {
```
eduardoboucas, line 27:
```suggestion
test('Symlinked directories from `includedFiles` are preserved', async () => {
```

tests/symlinked_included_files.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import decompress from 'decompress'
+import { dir as getTmpDir } from 'tmp-promise'
+import { expect, test } from 'vitest'
+
+import { ARCHIVE_FORMAT, zipFunction } from '../src/main.js'
+
+import { FIXTURES_ESM_DIR } from './helpers/main.js'
+
+const readDirWithTypes = async (dir: string, readFiles?: Record<string, boolean>, parent = '') => {
```
eduardoboucas, line 12: Should this be moved to a helper? Ideally with some JSDoc explaining what it does.

lukasholzer:
I had to use `decompress` instead of `AdmZip` used in other tests for decompressing as AdmZip is not restoring symlinks correctly. The symlinks are broken afterward with `AdmZip`, whereas with decompress, they are fully working (behaving like manual unzipping on Macos)

src/runtimes/node/utils/included_files.ts
```diff
@@ -53,9 +53,14 @@ export const getPathsOfIncludedFiles = async (
     cwd: basePath,
     dot: true,
     ignore: excludePatterns,
-    followSymbolicLinks: true,
+    followSymbolicLinks: false,
+    // get directories as well to get symlinked directories,
+    // to filter the regular non symlinked directories out mark them with a slash at the end to filter them out.
```
eduardoboucas, line 58: Can you move this to L60 so it's above the related property (`markDirectories`)?

tests/symlinked_included_files.test.ts
eduardoboucas: (Just saw [your comment](https://github.com/netlify/zip-it-and-ship-it/pull/1701#issuecomment-1912012374) about this.)

src/runtimes/node/utils/included_files.ts
lukasholzer: The related property is `onlyFiles`

lukasholzer: `get directories as well`

eduardoboucas: - "Get directories" = `onlyFiles`
- "Mark them with a slash" = `markDirectories`

Just trying to make sure these properties are easy to understand when people read the code.

src/runtimes/node/utils/included_files.ts
```diff
@@ -53,9 +53,13 @@ export const getPathsOfIncludedFiles = async (
     cwd: basePath,
     dot: true,
     ignore: excludePatterns,
-    followSymbolicLinks: true,
-    throwErrorOnBrokenSymbolicLink: true,
+    onlyFiles: false,
+    // get directories as well to get symlinked directories,
```
eduardoboucas, line 57: 😅 This is what I meant. Each property getting its own comment.

```suggestion
    // get directories as well to get symlinked directories
    onlyFiles: false,
```

tests/main.test.ts
```diff
@@ -2866,9 +2855,11 @@ describe('zip-it-and-ship-it', () => {
       }),
     })
 
-    const zip = new AdmZip(files[0].path)
+    const unzipPath = join(tmpDir, 'unzipped')
+
+    await decompress(files[0].path, unzipPath)
```
eduardoboucas, line 2860: Nice!

tests/symlinked_included_files.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import decompress from 'decompress'
+import { dir as getTmpDir } from 'tmp-promise'
+import { expect, test } from 'vitest'
+
+import { ARCHIVE_FORMAT, zipFunction } from '../src/main.js'
+
+import { FIXTURES_ESM_DIR } from './helpers/main.js'
+
+/** Small helper function, reading a directory recursively and returning a record with the files and if it is a symlink or not */
```
eduardoboucas, line 12: Comment looks good, but my question was about whether this should be moved to [a helper file](https://github.com/netlify/zip-it-and-ship-it/tree/main/tests/helpers) so it becomes easier to reuse in other tests.

eduardoboucas (approved):

tests/symlinked_included_files.test.ts
lukasholzer: I don't think we need it in other tests

eduardoboucas (approved):

PR #1711, opened 2024-02-20
feat: stop logging V2 function detection

eduardoboucas:
#### Summary

Cleaning up a system log line that I don't think is useful anymore, since the V2 detection mechanism has been proven to work and we have other ways of identifying the runtime version of a given function.


Skn0tt (approved):
✨ clean up that code ✨

PR #1713, opened 2024-02-28
chore: reenable windows symlink test

Skn0tt:
Reenables a test that was skipped before, presumably because of a Github Actions detail that I know how to fix now!

### DIFF
--- a/tests/symlinked_included_files.test.ts
+++ b/tests/symlinked_included_files.test.ts
@@ -68,7 +68,7 @@ test.skipIf(platform() === 'win32')('Symlinked directories from `includedFiles` 
   })
 })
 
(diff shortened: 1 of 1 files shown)


Skn0tt:
The now-enabled test was passing on my Windows VM. I'm expecting it to fail in CI, because the tmpdir that's created is located on the `C:/` drive, while the fixture is on the `D:/` drive. Windows doesn't like copying/symlinking across drives, so i'm expecting this to fail now. If it does, the fix will be to place the tmp dir on the `D:/` drive as well.

Skn0tt:
The other test is failing on Windows because of https://github.com/kevva/decompress/issues/41#issuecomment-1969067138.

eduardoboucas (approved):

PR #1717, opened 2024-04-01
chore: remove integration test

eduardoboucas:
#### Summary

We have an integration test in this repository that clones the latest version of the Next.js runtime and builds the demo site. This means that any change in the Next.js runtime repo may cause the test to fail without a clear indication as to why and without no clear path to resolution from the PR author.

This PR removes that test, which we should replace with the more comprehensive end-to-end test suite we're working on.


biruwon (approved):

JGAntunes (approved):
Yes, I approve this test deletion!

PR #1726, opened 2024-04-22
feat: ship instrumentation file in functions bundle

lukasholzer:

#### Summary

Fixes https://linear.app/netlify/issue/COM-455/build-console-span-exporter-for-the-new-runtime

Once https://github.com/netlify/serverless-functions-api/pull/221 is released we need to update this PR with the version and then we can ship the newly added `___netlify-telemetry.mjs` file in each function bundle.

---

      passes our tests.

### DIFF
--- a/tests/symlinked_included_files.test.ts
+++ b/tests/symlinked_included_files.test.ts
@@ -7,7 +7,6 @@ import { dir as getTmpDir } from 'tmp-promise'
 import { expect, test } from 'vitest'
 
 import { ARCHIVE_FORMAT, zipFunction } from '../src/main.js'
-import { BOOTSTRAP_FILE_NAME, TELEMETRY_FILE_NAME } from '../src/runtimes/node/utils/entry_file.js'
 
 import { FIXTURES_ESM_DIR } from './helpers/main.js'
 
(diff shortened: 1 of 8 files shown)


Skn0tt (commented):
could you write a test for this?

Skn0tt:
I don't want to add another file to every single bundle a customer of ours creates, without an easy to way to roll this back. Let's please add a flag, just so that we can roll this back for individual customers where this might break things. Customers have the most weird setups ...

tests/symlinked_included_files.test.ts
```diff
@@ -60,8 +61,9 @@ test.skipIf(platform() === 'win32')('Symlinked directories from `includedFiles`
 
   // expect that the symlink for `node_modules/crazy-dep` is preserved
   expect(await readDirWithType(unzippedPath)).toEqual({
-    '___netlify-bootstrap.mjs': false,
+    [BOOTSTRAP_FILE_NAME]: false,
```
Skn0tt, line 64: let's keep this as a string. we want a change in the constant to fail some tests, because it's part of the contract we have with e.g. functions origin

Skn0tt (approved):

PR #1732, opened 2024-05-06
fix: add `name` and `generator` support to ISC

Skn0tt:
Resolves https://linear.app/netlify/issue/COM-428/support-generator-and-name-in-isc-for-functions


src/runtimes/node/index.ts
```diff
@@ -151,9 +151,9 @@ const zipFunction: ZipFunction = async function ({
     bundler: bundlerName,
     bundlerWarnings,
     config,
-    displayName: config?.name,
+    displayName: config?.name ?? staticAnalysisName,
```
eduardoboucas, line 154: Should the ISC value not take precedence?

Skn0tt: Both is fine. I gave precedence to the existing value, because that way it's not a breaking change for functions that have both defined right now. Happy to flip though.

Skn0tt: yes, looks like we're giving ISC precedence for edge functions: https://github.com/netlify/edge-bundler/blob/0af1086fb92fa4eaa3e63709baac40df3f9ce06c/packages/edge-bundler/node/declaration.test.ts#L32

Skn0tt: flipped precedence in https://github.com/netlify/zip-it-and-ship-it/pull/1732/commits/14ac467e6b4ae3cff7f805cd01e432d19ea3d4d6

src/runtimes/node/index.ts
```diff
@@ -151,9 +151,9 @@ const zipFunction: ZipFunction = async function ({
     bundler: bundlerName,
     bundlerWarnings,
     config,
-    displayName: config?.name,
+    displayName: staticAnalysisName ?? config?.name,
     entryFilename: zipPath.entryFilename,
-    generator: config?.generator || getInternalValue(isInternal),
+    generator: staticAnalysisGenerator ?? config?.generator ?? getInternalValue(isInternal),
```
eduardoboucas, line 156: By changing `||` to `??`, you're saying that an empty string might take precedence over a non-empty value. I don't think this should be the case.

```suggestion
    generator: staticAnalysisGenerator || config?.generator || getInternalValue(isInternal),
```

src/runtimes/node/index.ts
```diff
@@ -151,9 +151,9 @@ const zipFunction: ZipFunction = async function ({
     bundler: bundlerName,
     bundlerWarnings,
     config,
-    displayName: config?.name,
+    displayName: staticAnalysisName ?? config?.name,
```
eduardoboucas, line 154: Same as below.

```suggestion
    displayName: staticAnalysisName || config?.name,
```

eduardoboucas (approved):
🚀

